PR #6, opened 2015-05-19
Initialise the default locale

skitt:
In non-English default locales, the tests fail because the target
strings expect filters to use English.


jaredstehler:
thanks!

PR #14, opened 2015-07-16
Escape string representation of objects too

markus-s24:
String representations of objects are not escaped, if requested. This patch computes the string representation of all objects like ExpressionNode does and escapes it.

I think, if someone explicitly applies this filter, it is OK to convert any object to a string.


jaredstehler:
deployed in ```2.0.8-SNAPSHOT```

PR #16, opened 2015-08-10
Change templates to extension resources

cubeton:
@jaredstehler 

We're moving to using the term extension_resources instead of templates. This PR can be merged whenever


cubeton:
not going to use extension_resource anymore

PR #28, opened 2016-01-28
add support for multi-variable assignment using 'set'

amannm:
I had to add this code to support parsing Jinja written like this:
```
{% set myvar1, myvar2, myvar3, myvar4 =
  'foo',
  'bar',
  '(hello, world, asdf, ooooo, fffff)',
  'yoooooo'
%}
```
which needs to result in assigning all 4 variables to all corresponding values


jaredstehler:
Thanks! I added a couple tests to cover edge cases.

PR #29, opened 2016-01-28
Add ErrorItem type

cubeton:
@jaredstehler

Add an `ErrorItem` enum to `TemplateError` so we can specify what type of object is causing a given error. In a perfect world this would be called `ErrorType` and `ErrorType` would be renamed `ErrorSeverity`. Sadly this is not a perfect world.

Let me know if I missed any possible 'types' or have too many.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/TemplateError.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/TemplateError.java
@@ -18,7 +18,6 @@ public class TemplateError {
     BAD_URL,
     EXCEPTION,
     MISSING,
-    MISSING_TEMPLATE,
     OTHER
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/hubspot/jinjava/interpret/TemplateError.java
```diff
@@ -21,8 +21,17 @@
     OTHER
   }
 
+  public enum ErrorItem {
+    TEMPLATE,
+    TOKEN,
+    TAG,
+    FUNCTION,
+    OTHER
```
jaredstehler: let's add one for `PROPERTY`, and update the missing errors above for `MISSING -> PROPERTY`

PR #32, opened 2016-02-04
add a pyishdate constructor that takes an Instant

dkulla:
I know I can get around this with something like `new PyishDate(instant.toEpochMilli())`, but having a constructor that takes an `Instant` directly seemed appropriate.

Let me know what you think.


jaredstehler:
deployed to maven central as `2.1.4-SNAPSHOT`

PR #40, opened 2016-02-24
Add Bool filter

### LINKED ISSUE #38: Add bool filter
I'm using this library to template files from an api where all values are strings and I wonder if you would be open to add a `bool` filter that would convert a true/false string to a boolean. I'm happy to submit a PR if you think it's suiting. This is what ansible does in their [playbook filters](http://docs.ansible.com/ansible/playbooks_filters.html#other-useful-filters).

jhaals:
The bool filter will convert a string/int value to a boolean

```
"true"|bool = true
"false"|bool = false
0|bool = false
1|bool = true
true|bool = true
false|bool = false
```

fixes #38


jaredstehler:
thanks!

jaredstehler:
deployed as `2.1.6-SNAPSHOT`

Issue #43, opened 2016-03-04
Problem with xmlattr filter


zepernick:
There seems to be a problem with the xmlattr and the way it binds itself to the filter method of the XmlAtttrFilter class.  Using this in my template:

`{{ boot_attributes_map|xmlattr }}`

Produces this error:

`com.hubspot.jinjava.interpret.FatalTemplateErrorsException: TemplateError{severity=FATAL, reason=SYNTAX_ERROR, message=TemplateSyntaxException: Method not found: class com.hubspot.jinjava.lib.filter.XmlAttrFilter.filter(com.hubspot.jinjava.objects.collections.PyMap, com.hubspot.jinjava.interpret.JinjavaInterpreter), fieldName=null, lineno=5}`

It seems like it is ignoring the String... on the method when trying to find it.  I can work around the error by sending a parameter to the xmlattr so that it finds the method:

`{{ boot_attributes_map|xmlattr('') }}`

I will see if I can identify where the problem is in the code and submit a pull request.

jaredstehler:
Can you post a reproducible test case? We call xmlattr with no filter args in the unit test: https://github.com/HubSpot/jinjava/blob/master/src/test/java/com/hubspot/jinjava/lib/filter/XmlAttrFilterTest.java#L30

jaredstehler:
And the version of JVM you're running as well?

zepernick:
I will get a test case together.  I did find out that this was happening with any filter that did not have parameters passed and not just the xmlattr.  I am running jdk1.8.0_65

jaredstehler:
Yeah I figured it wouldn't be specific to xmlattr. I'm grasping at straws but there might be some kind of conflict going on with the juel javax.el impl used in jinjava and struts/xwork...

zepernick:
see #39 for the resolution

Issue #39, opened 2016-02-23
Filters don't seem to work


xnaveira:
Hi, trying to use filters I get the following error:

```
com.hubspot.jinjava.interpret.FatalTemplateErrorsException: TemplateError{severity=FATAL, reason=EXCEPTION, message=InterpretException: Error resolving expression [testvar | upper]: IllegalArgumentException: wrong number of arguments, fieldName=null, lineno=1}
```

This is running:

```
Jinjava jinjava = new Jinjava(new JinjavaConfig());
String template = "hello {{ testvar | upper }}";
jinjava.render(template, ImmutableMap.of("testvar", "stranger")));
```

Running with no filters works fine.

I've checked with a colleague that's using jinjava in another project and he seems to be getting the same error.

jaredstehler:
which version are you using? I wasn't able to reproduce on the main branch.

xnaveira:
We're using 2.1.0 fetched via Maven.

jaredstehler:
I am still unable to reproduce..

```
import com.google.common.collect.ImmutableMap;
import com.hubspot.jinjava.Jinjava;
import com.hubspot.jinjava.JinjavaConfig;
import com.hubspot.jinjava.interpret.RenderResult;

public class Issue39 {

  public static void main(String[] args) {
    Jinjava jinjava = new Jinjava(new JinjavaConfig());
    String template = "hello {{ testvar | upper }}";

    RenderResult result = jinjava.renderForResult(template, ImmutableMap.of("testvar", "stranger"));
    System.out.println(result.getErrors());

    System.out.println(jinjava.render(template, ImmutableMap.of("testvar", "stranger")));
  }
}
```

produces following output:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[]
hello STRANGER
```

with pom dep:
```
    <dependency>
      <groupId>com.hubspot.jinjava</groupId>
      <artifactId>jinjava</artifactId>
      <version>2.1.0</version>
    </dependency>
```

jaredstehler:
Can you post the output of `mvn dependency:tree`?

xnaveira:
I just created an standalone project and get the same output as you. 

The problem appears when trying to use jinjava inside a dropwizard project. Let me gather some more data and get back to you.

[...]

PR #44, opened 2016-03-09
Send correct amount of parameters for Filter instance

zepernick:
Filters being called with no parameters were trying to execute the java method with 2 parameters which is not found on the Filter implementation.  There needs to be at least 3 parameters sent.

https://github.com/HubSpot/jinjava/issues/39
https://github.com/HubSpot/jinjava/issues/43

Previously, this expression was failing

`{{ boot_attributes_map|xmlattr }}`

but could be worked around by passing an empty parameter

`{{ boot_attributes_map|xmlattr('') }}`

This change ads the 3rd parameter as a null if none was sent from the filter expression.


zepernick:
This was an issue with Apache Tomcat running el-api 3.0 implementation and JinJava is bundled with juel-api which is an implementation for the 2.2 spec.  I will log more details here https://github.com/HubSpot/jinjava/issues/39

PR #45, opened 2016-03-16
Update Raw Tag

cubeton:
@boulter @mattfurtado 

Adding a `getTag` method for the TagNode class

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/RawTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/RawTag.java
@@ -41,7 +41,7 @@ public class RawTag implements Tag {
     return result.toString();
   }
 
-  private String renderNodeRaw(Node n) {
+  public String renderNodeRaw(Node n) {
     StringBuilder result = new StringBuilder(n.getMaster().getImage());
 
     for (Node child : n.getChildren()) {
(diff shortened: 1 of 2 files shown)


boulter:
:+1:

src/main/java/com/hubspot/jinjava/lib/tag/RawTag.java
```diff
@@ -41,7 +41,7 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
     return result.toString();
   }
 
-  private String renderNodeRaw(Node n) {
+  public String renderNodeRaw(Node n) {
```
jaredstehler: why is this becoming public? what's the use case?

src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -73,4 +73,8 @@ public String getHelpers() {
     return master.getHelpers();
   }
 
+  public Tag getTag() {
```
jaredstehler: :+1:

src/main/java/com/hubspot/jinjava/lib/tag/RawTag.java
cubeton: There's a problem where Custom HTML tags are actually rendering jinja within RawTags`, so I'm updating the render call within `WidgetAttributeTag` to set the value to the raw tag string instead of what the raw tag renders to, that way it won't double render and strip out the jinja.

jaredstehler: ah, cool

PR #48, opened 2016-03-31
list filter for strings

### LINKED ISSUE #47: list filter on strings
According to the list filter file: https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/lib/filter/ListFilter.java 
Applying a list filter to a string variable should return a sequence of characters.

```
{% set my_test = 'Test'%}
{{ my_test|list }}
```

prints something like

` [[C@45a31e18]`

Is this expected behavior for this filter, when used with strings? If so is there any practical application for this or should the filter really only be applied to numeric variables?

aaylward:
fixes #47 

@jaredstehler does this look ok to you?

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/ListFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/ListFilter.java
@@ -4,6 +4,7 @@ import java.util.Collection;
 import java.util.List;
 
 import com.google.common.collect.Lists;
+import com.google.common.primitives.Chars;
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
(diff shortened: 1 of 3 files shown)


boulter:
Would it be more useful just to return the a single element list containing the string? I'm not sure which behavior is more expected or useful.

jaredstehler:
according to jinja docs the `list` filter is supposed to convert a string into a list of chars. I agree that is not useful, but then neither is python so...

```
list(value)
Convert the value into a list. If it was a string the returned list will be a list of characters.```

LGTM

aaylward:
does this need version bump?

aaylward:
oop -- looks like it doesn't

PR #53, opened 2016-04-25
add range function

### LINKED ISSUE #49: Add range function to HubL
There is a Python function called "range", which is also available in Jinja2, that allows you to create an incrementing loop without the need to declare a scope. Can we have this added to our HubL accessible functions? This would make it a lot easier for us to create simple loops for stuff like numbered blog pagination. 

Thanks!
- Colby

boulter:
Implements `range()` ala Jinja2: http://jinja.pocoo.org/docs/dev/templates/

Fixes #49

Examples: 

` System.out.println(jinjava.render("| {% for num in range(2) %} {{ num }}, {% endfor %}|", context));`
`|  0,  1, |`
`System.out.println(jinjava.render("| {% for num in range(2, 8, 2) %} {{ num }}, {% endfor %}|", context));`
`|  2,  4,  6, |` 
`System.out.println(jinjava.render("| {% for num in range(2, -4, -2) %} {{ num }}, {% endfor %}|", context));`
`|  2,  0,  -2, |`

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
@@ -148,8 +148,8 @@ public class Functions {
     return pointer + 1;
   }
 
-  @JinjavaDoc(value = "<p>Return a list containing an arithmetic progression of integers. " +
-      " With one parameter, range will return a list from 0 up to or down to the value. " +
+  @JinjavaDoc(value = "<p>Return a list containing an arithmetic progression of integers.</p>" +
+      "<p>With one parameter, range will return a list from 0 up to or down to the value. " +
       " With two parameters, the range will start at the first value and end at the second value. " +
       " The third parameter specifies the step increment.</p> <p>All values can be negative.</p>" +
       "<p>Ranges can generate a maximum of " + RANGE_LIMIT + " values.</p>",
--- a/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
@@ -149,9 +149,9 @@ public class Functions {
   }
 
   @JinjavaDoc(value = "<p>Return a list containing an arithmetic progression of integers.</p>" +
-      "<p>With one parameter, range will return a list from 0 up to or down to the value. " +
-      " With two parameters, the range will start at the first value and end at the second value. " +
-      " The third parameter specifies the step increment.</p> <p>All values can be negative.</p>" +
+      "<p>With one parameter, range will return a list from 0 up to (but not including) the value. " +
+      " With two parameters, the range will start at the first value and increment by 1 up to (but not including) the second value. " +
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -145,4 +148,59 @@ public static int movePointerToJustBeforeLastWord(int pointer, String s) {
...
+    if (start < end) {
+      for (int i = start; i < end; i += step) {
+        if (result.size() >= RANGE_LIMIT) {
+          break;
+        }
+        result.add(i);
+      }
+    } else {
+
+      // make sure we're decrementing
```
brilee: This can be moved to a top-level `if (start < end && step > 0) { step = step * -1;}` and you won't have to copy-paste the for loop.

That being said, I think that trying to interpret the user's intention by fixing their buggy inputs is generally a bad idea. Better to return an empty list and let them figure out what they did wrong.

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -145,4 +148,59 @@ public static int movePointerToJustBeforeLastWord(int pointer, String s) {
...
+
+    List<Integer> result = new ArrayList<>();
+
+    int start = 0;
+    int end;
+    int step = 1;
+
+    switch (args.length) {
+      case 0:
+        end = Integer.parseInt(arg1.toString());
```
cubeton: Is it worth wrapping this in a try/catch or otherwise doing some check to make sure the arguments are integers (to avoid throwing a `NumberFormatException`)?

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
boulter: I still need two for loops because the end condition is different in each. Any idea what the python equivalent does with bad input?

boulter: python is not as lenient. I'll make it compatible.

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
boulter: python throws a `TypeError`. Not sure if throwing some other error would be helpful.

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
brilee: Ah, I didn't notice the end condition flip, carry on.

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
jaredstehler: I like using commons-lang NumberUtils.toInt(str, def), as it silently uses default for any number parse errors on inputs

tbchj: @cubeton  , answer me #104 
thank you

PR #54, opened 2016-04-26
Prevent Macro Cycles

boulter:
It's currently possible to define a macro that calls itself, either directly or indirectly and that can lead to a cycle.

This prevents cycles (and any recursion) in macros. 

I also refactored the other call stacks used for include and import cycle detection.


cubeton:
This is cool, nothing wrong sticks out to me

PR #55, opened 2016-04-26
Adding a new enum and error message information to TemplateError

cubeton:
@boulter @mattfurtado 

Adding more information to `TemplateError` which will allow customer-facing error messages to be constructed from an enum (and corresponding `categoryError` information) than just the English string returned.


boulter:
Looks like #54 's added exception will have to be updated as well.

cubeton:
Will make sure to add that @boulter if your PR gets merged first (which I believe it will)

Issue #52, opened 2016-04-25
String splitting does not work


lycis:
When I try to get a part of a string like this (currently using jinjava 2.1.7):

```jinja
{{ repository[0:4] }}
```

I get the following Exception thrown:
```
com.hubspot.jinjava.interpret.TemplateSyntaxException: Property repository is not a sequence.
	at com.hubspot.jinjava.el.ExpressionResolver.resolveExpression(ExpressionResolver.java:74)
	at com.hubspot.jinjava.interpret.JinjavaInterpreter.resolveELExpression(JinjavaInterpreter.java:323)
	at com.hubspot.jinjava.tree.ExpressionNode.render(ExpressionNode.java:41)
	at com.hubspot.jinjava.lib.tag.ForTag.interpret(ForTag.java:137)
	at com.hubspot.jinjava.lib.tag.Tag.interpretOutput(Tag.java:29)
	at com.hubspot.jinjava.tree.TagNode.render(TagNode.java:50)
	at com.hubspot.jinjava.lib.fn.MacroFunction.doEvaluate(MacroFunction.java:69)
	at com.hubspot.jinjava.el.ext.AbstractCallableMethod.evaluate(AbstractCallableMethod.java:72)
	at com.hubspot.jinjava.el.JinjavaInterpreterResolver.invoke(JinjavaInterpreterResolver.java:82)
	at de.odysseus.el.tree.impl.ast.AstMethod.eval(AstMethod.java:91)
	at de.odysseus.el.tree.impl.ast.AstMethod.eval(AstMethod.java:100)
	at de.odysseus.el.tree.impl.ast.AstEval.eval(AstEval.java:51)
	at de.odysseus.el.tree.impl.ast.AstNode.getValue(AstNode.java:31)
	at de.odysseus.el.TreeValueExpression.getValue(TreeValueExpression.java:122)
	at com.hubspot.jinjava.el.ExpressionResolver.resolveExpression(ExpressionResolver.java:62)
	at com.hubspot.jinjava.interpret.JinjavaInterpreter.resolveELExpression(JinjavaInterpreter.java:323)
	at com.hubspot.jinjava.tree.ExpressionNode.render(ExpressionNode.java:41)
	at com.hubspot.jinjava.interpret.JinjavaInterpreter.render(JinjavaInterpreter.java:181)
	at com.hubspot.jinjava.interpret.JinjavaInterpreter.render(JinjavaInterpreter.java:153)
[...]

PR #61, opened 2016-05-12
String range implemented according to #4.

anev:
String range implemented according to #52.
Now `{{ theString[1:4] }}` will render as a substring.

More examples:
```
{{ theString[1:] }}
{{ theString[:3] }}
{{ theString[-7:-4] }}
```


src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
```diff
@@ -1,21 +1,19 @@
...
-import javax.el.PropertyNotFoundException;
-
 import com.hubspot.jinjava.objects.collections.PyList;
-
 import de.odysseus.el.misc.LocalMessages;
 import de.odysseus.el.tree.Bindings;
 import de.odysseus.el.tree.impl.ast.AstBracket;
 import de.odysseus.el.tree.impl.ast.AstNode;
 
+import javax.el.ELContext;
```
boulter: could you preserve the import order here?

src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
```diff
@@ -81,6 +86,19 @@ public Object eval(Bindings bindings, ELContext context) {
     return result;
   }
 
+  private String evalString(String base, int startNum, int endNum) {
+    if (startNum > endNum) {
+      return "";
+    }
+    if (startNum <= 0) {
```
boulter: could replace these with `Math.max` and `Math.min`

src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java
```diff
@@ -1,17 +1,16 @@
...
-
-import org.junit.Before;
-import org.junit.Test;
-
 import com.google.common.base.Throwables;
 import com.google.common.io.Resources;
 import com.hubspot.jinjava.Jinjava;
 import com.hubspot.jinjava.interpret.Context;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import org.junit.Before;
```
boulter: please preserve import order

src/test/java/com/hubspot/jinjava/el/ext/RangeStringTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertThat(jinjava.render("{{ theString[3:8] }}", context)).isEqualTo("ti\nli");
+    }
+
+    @Test
+    public void testStringRangeCyrillic() {
+        Map<String, Object> context = new HashMap<>();
+        context.put("theString", "Строка с non ascii символами");
+
+        assertThat(jinjava.render("{{ theString[1:4] }}", context)).isEqualTo("тро");
+    }
```
boulter: Could you also support negative and unspecified indexes such as `[-7:-3]` and `[3:]` for compatibility with python and Jinja? Some good examples here: http://pythoncentral.io/cutting-and-slicing-strings-in-python/

boulter:
Thanks for your contribution!

src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
anev: @boulter 
Not sure which is right. Is the order correct?
1. java.*
2. javax.*
3. com.*
4. de.*

anev:
@boulter 
empty ranges like `[:5]` are possible now, negative like `[-7:-2]` too.

Hope this pull request is useful for the project.

PR #65, opened 2016-06-01
Adds escapejs filter

Specialkbyte:
@boulter


src/main/java/com/hubspot/jinjava/lib/filter/EscapeJsFilter.java
```diff
@@ -0,0 +1,108 @@
...
+    },
+    snippets = {
+        @JinjavaSnippet(
+            code = "{% set escape_string = \"This string can safely be inserted into JavaScript\" %}\n" +
+                "{{ escape_string|escapejs }}")
+    })
+public class EscapeJsFilter implements Filter {
+
+  @Override
+  public Object filter(Object o, JinjavaInterpreter jinjavaInterpreter, String... strings) {
```
boulter: single character parameter names are meh

boulter:
looks good

PR #66, opened 2016-06-14
Update macro tag to allow for included macros

cubeton:
@mattfurtado @boulter 

Currently, if a file contains a macro (let's call it File "Macro") and that file is included via some tag in another file (let's call it File "Included Macro"), then File "Macro" will save fine, but File "Included Macro" will have errors because the macro is saved to a lower/different context level.

This updates saving macros to save to the top-level interpreter. Looking for feedback on the approach.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
@@ -142,6 +142,7 @@ public class MacroTagTest {
   @Test
   public void itCanAccessMacrosAcrossFiles() throws IOException {
     String template = Resources.toString(Resources.getResource("tags/macrotag/include-another-macro.jinja"), StandardCharsets.UTF_8);
+    interpreter.render(template);
     assertThat(interpreter.getErrors()).isEmpty();
   }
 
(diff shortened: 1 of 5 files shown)


src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
```diff
@@ -139,6 +139,13 @@ public void itAllowsMacrosCallingMacrosUsingCall() throws IOException {
     assertThat(out).contains("Hello World Two");
   }
 
+  @Test
+  public void itCanAccessMacrosAcrossFiles() throws IOException {
+    String template = Resources.toString(Resources.getResource("tags/macrotag/include-another-macro.jinja"), StandardCharsets.UTF_8);
+    interpreter.render(template);
+    assertThat(interpreter.getErrors()).isEmpty();
```
boulter: can you verify that the macro works too?

boulter:
The only problem I can see is that you could have macro conflicts if you've defined the same macro in different files, but that's more of a code organization problem.

cubeton:
@boulter agreed, that could be a possible issue - although it's technically an issue now anyways, in the case of a given file having multiple macros of the same name. In the current approach we 'overwrite' a map of macros based on name, so the macro that's last defined is the one whose definition we'd use (unless macros and macro-usages are interleaved).

I also updated the test to check output as well.

cubeton:
@boulter I'm going back over this PR because of the test failure in the `ImportTagTest` - in doing so I noticed that we do sort of have this functionality already. If a user wants to 'combine' a file using the `import` tag over `include` tag, everything's fine. I'm thinking of maybe closing this PR and directing users to use `import` over `include` in these cases. Thoughts?

boulter:
sounds good!

PR #74, opened 2016-09-09
Patch from Artur Pogoda de la Vega

markus-s24:


boulter:
Thanks for the contribution. It looks like this adds support for truncated division and power operators. It looks like the build failed with this change. Would you mind fixing and adding some tests. Thank you!

markus-s24:
Sorry, the patch was not complete. I will fix that next week.

PR #75, opened 2016-09-09
Patch from Artur Pogoda de la Vega

markus-s24:


boulter:
Thanks for the PR. Would you mind adding a test?

PR #76, opened 2016-09-12
Power and truncated division operators

markus-s24:

Content:
1) Add operators ** and //
2) Fix of "for" tag with ranges containing spaces


boulter:
ok, I cleaned up the formatting a bit. In general it's best if the operators and range checks are separate PRs, but I'm merging this one. 

Thanks for your contribution.

PR #83, opened 2016-10-28
Restrict tags, expression tests and filters

boulter:
Certain tags, expression tests and filters may be incompatible with a tag or perhaps you want to disable some of these for security. This allows you to specific tags, expression tests and filters that cannot be called in child nodes.


bgoodies:
👍

PR #89, opened 2016-11-16
respect '-' in tags to eliminate whitespace

### LINKED ISSUE #86: Manual whitespace control doesn't seem to work
The Jinja specification allows for manual whitespace control by adding a minus at the start or end of tags (http://jinja.pocoo.org/docs/dev/templates/#whitespace-control) like:

```
{% for foo in [1,2,3,4,5,6,7,8,9] -%}
{{ foo }}
{%- endfor %}
```

This code should yield `123456789`. However it yields

```
1

2

3

4

5

6

7

8

9
```

boulter:
```
{% for foo in [1,2,3] -%}
  {{ foo }}
{%- endfor %}
```

now produces `1,2,3`, eliminating whitespace chars after or before the tag.

Fixes #86 

@pfarrel @lcmartinez @bgoodies @derkork


pfarrel:
👍

PR #92, opened 2016-11-22
Enable use of arbitrary variable names when looping over dictionary entries

glindste:
Previously you could only use the variable names `key` and `value` when
looping over a dictionary. This conflicted with the python implementation
of jinja where the loop variables can have arbitrary names, and would
have made nested looping over dictionary entries hard due to name
collisions.

Worth noting is that the existing `forLoopMultipleLoopVars()` still passed if the template was changed to use different names than `key` and `value` for the variables. The loop still worked but the variables didn't contain any content, i.e. they were empty strings. The test still passed since it only checked the number of lines.


boulter:
looks good, thanks for the contribution!

boulter:
I deployed a new version of 2.1.15-SNAPSHOT with this change.

glindste:
Thanks!

PR #97, opened 2017-01-17
Shade-relocate JUEL dependencies

joncfoo:
Related to #39 - this PR shade-relocates the JUEL dependency which solves the problem of conflicting EL implementations available at run-time.

I'm not sure how to exclude dependencies if building for a separate classifier - any help on that front would be much appreciated!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 .DS_Store
 
 # Maven
+dependency-reduced-pom.xml
 log/
 target/
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,30 @@ Get it:
   </dependency>
 ```
 
+or if you have a conflicting EL implementation on your classpath:
+```xml
+<dependency>
+  <groupId>com.hubspot.jinjava</groupId>
+  <artifactId>jinjava</artifactId>
+  <version>2.1.14</version>
+  <classifier>with-shaded-juel</classifier>
+  <exclusions>
+    <exclusion>
+      <artifactId>juel-api</artifactId>
+      <groupId>de.odysseus.juel</groupId>
+    </exclusion>
+    <exclusion>
+      <artifactId>juel-impl</artifactId>
+      <groupId>de.odysseus.juel</groupId>
+    </exclusion>
+    <exclusion>
+      <artifactId>juel-spi</artifactId>
+      <groupId>de.odysseus.juel</groupId>
+    </exclusion>
+  </exclusions>
+</dependency>
+```
+
 or if you're stuck on java 7:
 ```xml
   <dependency>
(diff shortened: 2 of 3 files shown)


boulter:
Can you think of any better way to do this, @jhaber?

jhaber:
AFAIK there's no way to exclude dependencies for the shaded artifact because it shares a POM with the main artifact (ie, only 1 POM will ultimately get published to Maven central). To accomplish that you could publish the shaded artifact under a separate version (for example `2.1.15-with-shaded-juel`) rather than using a classifier. Having a separate version would let you publish a separate POM, and would also be more friendly to managing with `dependencyManagement`. For example if you have a large dependency tree with Jinava being pulled in from multiple places, if some of those are using the classifier and some are using the main artifact you will end up with both artifacts (which could be different versions entirely) in your dependency tree and on your classpath (because the classifier is part of the Maven coordinates). To fix, you would need to exclude one of the artifacts from every dependency that pulls it in. Whereas if you use a version to differentiate the shaded artifact, then the Maven coordinates are the same between the shaded and non-shaded artifact so Maven guarantees that you only end up with one in your dependency tree and you can easily control which one via `dependencyManagement`. If you wanted the regular artifact you would do:

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>com.hubspot.jinjava</groupId>
      <artifactId>jinjava</artifactId>
      <version>2.1.15</version>
    </dependency>
  </dependencies>
</dependencyManagement>
```

And if you wanted the shaded artifact you would do:

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>com.hubspot.jinjava</groupId>
      <artifactId>jinjava</artifactId>
      <version>2.1.15-with-shaded-juel</version>
    </dependency>
  </dependencies>
</dependencyManagement>
```

That being said, it seems like `juel-api` and `juel-impl` haven't gotten a new release in 3 years so my recommendation would be to skip the classifier and separate version and just publish a single artifact with these dependencies bundled.

joncfoo:
@jhaber I've updated the PR to remove the classifier & the SPI (no longer needed if shade-relocating all of JUEL).

However I still see juel-* as dependencies in a downstream project.  I thought that the shade plugin would remove the dependencies from pom.xml when deployed.  I'm sure I'm missing something.

jhaber:
Does it work if you add:

```xml
<shadedArtifactAttached>false</shadedArtifactAttached>
```

to the maven-shade-plugin configuration?

joncfoo:
That did the trick, thanks!  It works as expected.  Would it be possible to cut a release soon-ish?

boulter:
2.1.15 has been released. Thanks!

joncfoo:
Thank you!

PR #101, opened 2017-01-27
persist dependencies found in lower scopes

olyazavr:
Dependencies in lower scopes would not be bubbled up to the parent; this copies all lower scope dependencies to the parent context before leaving the scope

@boulter


boulter:
Thanks! Any tests you can add to verify these bubble all the way up?

olyazavr:
Added a test!

bgoodies:
👍

Issue #11, opened 2015-06-16
add support for named parameters to filters


jaredstehler:
currently filters only support parameters positionally (even though we allow naming syntactically, we actually ignore the names when invoking the filter).

boulter:
Hi Jose, I don't think we have a clear idea of how we'd do named parameters. How about starting a PR and we'll work though it there?

Thanks!

igalarzab:
Sorry for the delay, here is the PR: #105 

Please, give some feedback when you have some free time :)

PR #105, opened 2017-03-20
Adding named parameters to filters

igalarzab:
Implementing named arguments in the filters (Issue #11).

I hope the code is self-explanatory but here are some comments to clarify design decisions:

* The AST is not changed in any way (to be a representation of what is read in the template)
* There is a new interface called `AdvancedFilter` (a better name is welcome 😄 ) that has one abstract method with the signature `filter(var, interpreter, args, kwargs)`. This method should be implemented by new Filters.
* I take the chance to make `AdvancedFilter` receive `Object` instead of `String` (as `Filter` does) so they can receive more variety of arguments (numbers, floats, bools, ...).
* The old interface `Filter` still exists to maintain backward-compatibility and to allow clients to create simpler filters (without named parameters compatibility and receiving only strings).
* The `AdvancedFilter` implements the `Filter` interface to maintain backwards compatibility (so they can be registered as normal filters).
* The Interpreter Resolver changes the signature when it detects that a filter should be run. The signature separate positional arguments and named arguments. Because at this point we don't know if the filter implements the `Filter` or the `AdvancedFilter` interface, the `Filter` interface knows how to transform the `(args, kwargs)` arguments to an old `(args)` one.
* I changed the `SumFilter` as an example.

At this point all tests pass. I would like to create more tests but I want to receive some feedback from you before doing it.

Cheers

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
@@ -7,12 +7,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 
 import javax.el.ArrayELResolver;
 import javax.el.CompositeELResolver;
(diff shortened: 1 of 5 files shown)


boulter:
Thanks, Jose! I hope to get some time to look at this later in the week.

src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
```diff
@@ -7,12 +7,7 @@
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
boulter: our style is not to collapse imports

src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
```diff
@@ -22,14 +17,11 @@
 import javax.el.PropertyNotFoundException;
 import javax.el.ResourceBundleELResolver;
 
+import com.hubspot.jinjava.el.ext.*;
```
boulter: same comment as above

igalarzab:
Hey Jeff,

Thanks for the feedback. I unfolded the imports given your comments and I'll upload some tests later today.

And again, thank you for the library! We are using it at TransferWise and we are really happy with it 😀 

Cheers

igalarzab:
Test uploaded

PR #107, opened 2017-03-29
correct function binding syntax in custom function example

nojo:
Example in README.md contains incorrect syntax for calling a custom function - correcting this so that people don't have to read through the code to find out how to call a custom function.


boulter:
thanks!

PR #108, opened 2017-03-30
Limit output size

boulter:
Adds a configuration option to limit the maximum size of an output node.

@jmagnarelli-hs

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
+++ b/src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
@@ -3,12 +3,25 @@ package com.hubspot.jinjava.tree.output;
 import java.util.LinkedList;
 import java.util.List;
 
+import com.hubspot.jinjava.interpret.OutputTooBigException;
+
 public class OutputList {
 
   private final List<OutputNode> nodes = new LinkedList<>();
   private final List<BlockPlaceholderOutputNode> blocks = new LinkedList<>();
+  private long maxOutputSize;
+
+  public OutputList(long maxOutputSize) {
+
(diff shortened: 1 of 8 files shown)


src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
```diff
@@ -3,12 +3,25 @@
...
+import com.hubspot.jinjava.interpret.OutputTooBigException;
+
 public class OutputList {
 
   private final List<OutputNode> nodes = new LinkedList<>();
   private final List<BlockPlaceholderOutputNode> blocks = new LinkedList<>();
+  private long maxOutputSize;
+
+  public OutputList(long maxOutputSize) {
+
```
jmagnarelli-hs: Nit: why the extra line here?

src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
```diff
@@ -3,12 +3,25 @@
...
+    this.maxOutputSize = maxOutputSize;
+  }
 
   public void addNode(OutputNode node) {
+
+    if (maxOutputSize > 0 && node.getSize() > maxOutputSize) {
```
jmagnarelli-hs: Does this only check the size of this node against the max? What if we're adding a large number of very small nodes? Just pushed a test that shows what I mean.

If this is only intended to limit the size of individual nodes, I think we should change the name of `maxOutputSize` to something else.

boulter: should be fixed now

jmagnarelli-hs:
🍹

PR #112, opened 2017-04-10
Null check within filter

bgoodies:
This prevents null pointer exception in cases where a default value is used and is not supplied during validation.

@jboulter @olyazavr

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/Filter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/Filter.java
@@ -15,13 +15,13 @@ limitations under the License.
  **********************************************************************/
 package com.hubspot.jinjava.lib.filter;
 
-import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/Filter.java
```diff
@@ -49,7 +49,7 @@
   default Object filter(Object var, JinjavaInterpreter interpreter, Object[] args, Map<String, Object> kwargs) {
     // We append the named arguments at the end of the positional ones
     Object[] allArgs = ArrayUtils.addAll(args, kwargs.values().toArray());
-    String[] stringArgs = Arrays.stream(allArgs).map(Object::toString).toArray(String[]::new);
+    String[] stringArgs = Arrays.stream(allArgs).map(arg -> arg != null ? arg.toString() : null).toArray(String[]::new);
```
boulter: I think you can use `Object.toString(arg, null)` here

PR #114, opened 2017-04-12
Sort named parameters

boulter:
When converting named parameters to positional parameters for backwards compatibility, make sure we maintain the order.

@bgoodies @olyazavr


bgoodies:
👍

PR #118, opened 2017-05-12
remove FileResourceLocator as a default

boulter:
ClasspathResourceLocator is the only default now to avoid accidentally allowing loading files from arbitrary locations.


kevinbluett:
LGTM 🚀

Issue #121, opened 2017-05-17
Why is expression result reinterpreted?


michal-klimek:
Please consider the following code:
```
final String tpl = "The question is: {{ obj.text }}!";
final Map<String, Object> ctx = new HashMap<>();
final Map<String, Object> obj = new HashMap<>();
obj.put("text", "{{ obj.val }}");
obj.put("val", "is it a bug?");
ctx.put("obj", obj);
Jinjava jj = new Jinjava();
System.out.println(jj.render(tpl, ctx));
```
The expected output is: `The question is {{ obj.val }}!`, but `The question is: is it a bug?!` appears on my console.

I have created similar code in Jinja2 (python):
```
from jinja2 import Template
t = Template('The question is {{ obj.text }}!')
ctx = {"obj": {"text": "{{ obj.val }}", "val": "is it a bug?"}}
print t.render(ctx)
```
The render result is as expected: `The question is {{ obj.val }}!`

I would like to know if it is an intended behavior, because in my application I have some templates where expressions insert google-encoded polylines which may occasionally contain `{{  }}` characters, which in turn are interpreted as another expression with unresolvable name.

hs-lsong:
This seems working as designed for jinjava. There is in fact a test to verify that the expression is interpreted. 
https://github.com/HubSpot/jinjava/blob/master/src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java#L28

It is definitely a divergence from the python version. It is unclear if anyone is depending this behavior.

straight-shoota:
This is a potential security risk, because it allows user-generated text to be evaluated as template.

mattcoley:
I believe this should be turned off by default. In addition to the security issue, this behavior also breaks `raw` tags. It took me a while to figure out this setting and I am guessing most developers building applications that are using jinjava will not be aware of this issue. https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/JinjavaConfig.java#L55

boulter:
I agree it would be safer to turn it off, but for existing templates, it could definitely break things.

PR #127, opened 2017-07-17
Disable interpreting variables in expression.

hs-lsong:
Fixes https://github.com/HubSpot/jinjava/issues/121

This PR removes the code that interprets strings containing '{{' that is already an expression.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -43,7 +43,7 @@ public class JinjavaConfig {
 
   private Map<Context.Library, Set<String>> disabled;
   private final boolean failOnUnknownTokens;
-  private final boolean enableNestedInterpretation;
+  private final boolean nestedInterpretationEnabled;
 
   public static Builder newBuilder() {
     return new Builder();
@@ -69,7 +69,7 @@ public class JinjavaConfig {
                         boolean enableRecursiveMacroCalls,
                         boolean failOnUnknownTokens,
                         long maxOutputSize,
-                        boolean enableNestedInterpretation) {
+                        boolean nestedInterpretationEnabled) {
     this.charset = charset;
     this.locale = locale;
     this.timeZone = timeZone;
@@ -81,7 +81,7 @@ public class JinjavaConfig {
     this.enableRecursiveMacroCalls = enableRecursiveMacroCalls;
     this.failOnUnknownTokens = failOnUnknownTokens;
     this.maxOutputSize = maxOutputSize;
-    this.enableNestedInterpretation = enableNestedInterpretation;
+    this.nestedInterpretationEnabled = nestedInterpretationEnabled;
   }
 
   public Charset getCharset() {
@@ -128,8 +128,8 @@ public class JinjavaConfig {
     return failOnUnknownTokens;
   }
 
-  public boolean isEnableNestedInterpretation() {
-    return enableNestedInterpretation;
+  public boolean isNestedInterpretationEnabled() {
+    return nestedInterpretationEnabled;
   }
 
   public static class Builder {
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -43,17 +43,18 @@
 
   private Map<Context.Library, Set<String>> disabled;
   private final boolean failOnUnknownTokens;
+  private final boolean noNestedInterpretation;
```
boulter: in general, I think it's easier to grok booleans as positives rather than negatives. So instead of `noNestedIntepretation`, maybe `enableNestedExpressions`

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -197,8 +205,13 @@ public Builder withMaxOutputSize(long maxOutputSize) {
       return this;
     }
 
+    public Builder withNoNestedInpterpretation(boolean noNestedInterpretation) {
```
boulter: typo in method name

src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java
```diff
@@ -30,7 +31,19 @@ public void itRendersResultAsTemplateWhenContainingVarBlocks() throws Exception
     context.put("place", "world");
 
     ExpressionNode node = fixture("simplevar");
-    assertThat(node.render(interpreter).toString()).isEqualTo("hello {{ place }}");
+    assertThat(node.render(interpreter).toString()).isEqualTo("hello world");
+  }
+
+  @Test
+  public void itRendersResultWithoutInterpretateExpression() throws Exception {
```
boulter: typo in method name

hs-lsong: Done.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
hs-lsong: Done. I tried to make the default value to be false.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
hs-lsong: Done.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -125,6 +128,10 @@ public boolean isFailOnUnknownTokens() {
     return failOnUnknownTokens;
   }
 
+  public boolean isEnableNestedInterpretation() {
```
boulter: if the boolean was called `nestedInterpretationEnabled`, then the getter would read more cleanly as `isNestedInterpretationEnabled`

hs-lsong: There is also one enableRecursiveMacroCalls, do you think we should rename them together?

boulter: good point. That was mine. :) Maybe rename in a separate PR so as not to clutter up this one.

hs-lsong: OK, done. Leave that name as is for now.

src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java
boulter: I think there's still a typo in there. Should be `itRendersResultWithoutInterpreterExpression`

hs-lsong: Done.

PR #129, opened 2017-07-21
Default nestedInterpretationEnabled to be true in JinjavaConfig builder.

hs-lsong:


src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java
```diff
@@ -47,6 +47,18 @@ public void itRendersResultWithoutNestedExpressionInterpretation() throws Except
   }
 
   @Test
+  public void itRendersResultWithDefaultBuilder() throws Exception {
```
boulter: maybe `itRendersWithNestedExpressionInterpretationByDefault`?

hs-lsong: Done.

PR #132, opened 2017-08-01
Fixing filters with uppercase letters

igalarzab:
Hi!

Filters that are using uppercase characters are broken as the `lowercase()` method is being executed to the name before registering them into the `SimpleLibrary`.

They are broken because in the `fetch` method the filter name is just lowercased before being compared with the existing ones, but **not** before adding them to the library, so what's happening is that:

* The filters are registered with the camelcase name inside the library, ex: `filterName`.
* The filters cannot be accessed, as the fetch method tries to find a filter named `filtername`.

There were two solutions:

1) Lowercasing the filter before doing the `put` into the HashMap.
2) Allowing filter names to have not just lowercase characters.

I chose the second option as I didn't see any reason for the restriction.

I added one test and the patch is backwards compatible, as filters with uppercase characters were not usable before.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/SimpleLibrary.java
+++ b/src/main/java/com/hubspot/jinjava/lib/SimpleLibrary.java
@@ -57,7 +57,7 @@ public abstract class SimpleLibrary<T extends Importable> {
       throw new DisabledException(item);
     }
 
-    return lib.get(StringUtils.lowerCase(item));
+    return lib.get(item);
   }
 
   @SafeVarargs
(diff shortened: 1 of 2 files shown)


igalarzab:
Hi @boulter,

Thanks for merging it :)

Any possibility of having a new release in mvn?

jmp3833:
Hi @igalarzab. Going to create a new release 2.2.4 in maven that includes these changes in addition to #133. I'll let you know when the new version is available 👍

igalarzab:
great, thanks @jmp3833 !

jmp3833:
No problem! Should be available in the central repository within the next couple of hours

PR #133, opened 2017-08-01
Add function to apply resolved strings from another Context object

jmp3833:
Useful for passing resolved expressions, functions, and values to the parent Context object. This way when evaluating nested content, these resolved values can be exposed up the context chain. 

@jboulter @jmagnarelli-hs

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -221,6 +221,18 @@ public class Context extends ScopeMap<String, Object> {
     this.superBlock = null;
   }
 
+  /**
+   * Take all resolved strings from a context object and apply them to this context.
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -221,6 +221,18 @@ public void removeSuperBlock() {
     this.superBlock = null;
   }
 
+  /**
+   * Take all resolved strings from a context object and apply them to this context.
+   * Useful for passing resolved values up a tag hierarchy.
+   *
+   * @param context - context object to apply resolved values from.
+   */
+  public void applyResolvedFrom(Context context) {
```
boulter: why 'apply' instead of just 'add'?

boulter:
@hs-lsong

hs-lsong:
LGTM.

PR #135, opened 2017-08-02
Add configurable random number generator

boulter:
Useful for testing when you want to compare against known values.

@hs-lsong @pfarrel @jmagnarelli-hs

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/RandomFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/RandomFilter.java
@@ -20,7 +20,6 @@ import java.math.BigDecimal;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
-import java.util.concurrent.ThreadLocalRandom;
 
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
@@ -52,7 +51,7 @@ public class RandomFilter implements Filter {
       if (size == 0) {
         return null;
       }
-      int index = ThreadLocalRandom.current().nextInt(size);
+      int index = interpreter.getRandom().nextInt(size);
       while (index-- > 0) {
         it.next();
       }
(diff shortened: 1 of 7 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/RandomFilter.java
```diff
@@ -75,20 +74,20 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
...
+      int index = interpreter.getRandom().nextInt(size);
       while (index-- > 0) {
         it.next();
       }
       return it.next();
     }
     // number
     if (object instanceof Number) {
-      return ThreadLocalRandom.current().nextLong(((Number) object).longValue());
+      return interpreter.getRandom().nextInt(((Number) object).intValue());
```
hs-lsong: Why changing long to int here?

src/main/java/com/hubspot/jinjava/lib/filter/RandomFilter.java
```diff
@@ -75,20 +74,20 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
...
     }
     // string
     if (object instanceof String) {
       try {
-        return ThreadLocalRandom.current().nextLong(new BigDecimal((String) object).longValue());
+        return interpreter.getRandom().nextInt(new BigDecimal((String) object).intValue());
```
hs-lsong: Same here.

src/test/java/com/hubspot/jinjava/lib/filter/ShuffleFilterTest.java
```diff
@@ -2,27 +2,31 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
+import static org.mockito.Mockito.*;
```
hs-lsong: import static org.mockito.Mockito.when should be enough.

src/main/java/com/hubspot/jinjava/lib/filter/RandomFilter.java
boulter: There's no `nextLong()` that takes a bound in `java.util.Random`. I figured random ints are probably big enough.

src/test/java/com/hubspot/jinjava/lib/filter/ShuffleFilterTest.java
boulter: Our style is to always collapse imports for Mockito and a few other deps.

PR #137, opened 2017-08-03
Limit length of strings while building

boulter:
In https://github.com/HubSpot/jinjava/pull/108, we limited the size of output as we concatenated nodes. This did not prevent the case of a single node (tag) producing a large string. This replaces the standard `StringBuilder` with a builder that refuses to grow over the specified size in every place where we could possibly generate a very large string.

@hs-lsong @jmagnarelli-hs


jmagnarelli-hs:
A little late, but LGTM as well.

PR #140, opened 2017-08-12
Delegate PyMap::toString to underlying Map

pfarrel:
PyMap is a slightly leaky abstraction here, ForwardingMap does not forward toString to the underlying Map. This is a problem if using an underlying Map with custom / important toString behaviour (such as the maps used for contact properties in the COS).

@boulter @jmagnarelli-hs


boulter:
Congrats First-time contributor! :)

LGTM

PR #142, opened 2017-08-14
Prevent recursion in Jinjava.

hs-lsong:
Although we limit the rendering depth at 10, but that can still be large if the recursive expression has multiple self references. With this change, we still render the recursion once, then stop render is any further.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -396,8 +396,8 @@ public class Context extends ScopeMap<String, Object> {
     return renderStack.pop();
   }
 
-  public Stack getRenderStack() {
-    return renderStack;
+  public boolean doesRenderStackContain(String template) {
+    return renderStack.contains(template);
   }
 
   public void addDependency(String type, String identification) {
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Stack;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.HashMultimap;
@@ -73,6 +74,8 @@ public class Context extends ScopeMap<String, Object> {
   private int renderDepth = -1;
   private Boolean autoEscape;
   private List<? extends Node> superBlock;
+
+  private final Stack<String> renderStack = new Stack<>();
 
   public Context() {
     this(null, null, null);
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -385,6 +388,18 @@ public void setRenderDepth(int renderDepth) {
...
 
+  public void pushRenderStack(String template) {
+    renderStack.push(template);
+  }
+
+  public String popRenderStack() {
+    return renderStack.pop();
+  }
+
+  public Stack getRenderStack() {
```
boulter: If we really want to protect the underlying stack from being modified outside the Context, then I think you just need to implement `doesStackContain()`. Otherwise this allows total access to the stack contents.

src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
```diff
@@ -158,7 +158,27 @@ public void itAllowsMacroRecursionWhenEnabledInConfiguration() throws IOExceptio
...
+        "    'html_1' : {" +
+        "        'tag'           : 'html {{ selector }}'," +
+        "        'span'          : 12" +
+        "    }" +
+        "} %}" +
+        "{% set selector='{{spans}}' %}" +
+        "{{ allSpans(spans, '') }}" +
+        "";
+    Node node = new TreeParser(interpreter, jinja).buildTree();
+    assertThat(JinjavaInterpreter.getCurrent() == interpreter).isTrue();
```
boulter: what is this line testing for?

src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
```diff
@@ -158,7 +158,27 @@ public void itAllowsMacroRecursionWhenEnabledInConfiguration() throws IOExceptio
     }
   }
 
-
+  @Test
+  public void itPreventsRecursionForMacroWithVar() {
+    String jinja = "{%- macro allSpans(spans) %}" +
```
boulter: Can this example be simplified? It reflects the original issue, but I think the the test would be easier to understand if it the content of spans was smaller and it wasn't so specific to the original issue with spans, since it really could be a problem with any type of map.

src/test/java/com/hubspot/jinjava/tree/ExpressionNodeTest.java
```diff
@@ -77,7 +77,7 @@ public void itAvoidsInfiniteRecursionOfItself() throws Exception {
   }
 
   @Test
-  public void itNoRecursionHere() throws Exception {
+  public void itDoesNotRescursivelyEvaluateExpressions() throws Exception {
```
hs-lsong: In fact, there is not recursion here.

src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
```diff
@@ -158,7 +158,25 @@ public void itAllowsMacroRecursionWhenEnabledInConfiguration() throws IOExceptio
     }
   }
 
-
+  @Test
+  public void itPreventsRecursionForMacroWithVar() {
```
hs-lsong: Maybe this test is not necessary, since the recursion cases are tested in the ExpressionNodeTest. This is to confirm that the recursion happens even when macro is called.

PR #144, opened 2017-08-15
Fix import tag tests for cycle detection.

hs-lsong:
We have redundant cycle detection now. This PR fixes the tests that expect the detection inside the importTag, in fact it is detected and stopped before the importTag.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
@@ -33,11 +33,16 @@ import java.util.concurrent.ThreadLocalRandom;
 import org.apache.commons.lang3.StringUtils;
 
 import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -208,11 +213,15 @@ public String render(Node root, boolean processExtendRoots) {
 
     for (Node node : root.getChildren()) {
       lineNumber = node.getLineNumber();
-      if (context.doesRenderStackContain(node.getMaster().getImage())) {
+      String renderStr = node.getMaster().getImage();
+      if (context.doesRenderStackContain(renderStr)) {
         // This is a circular rendering. Stop rendering it here.
-        output.addNode(new RenderedOutputNode(node.getMaster().getImage()));
+        addError(new TemplateError(ErrorType.WARNING, ErrorReason.EXCEPTION, ErrorItem.TAG,
+            "Circular rendering detected: '" + renderStr + "'", null, getLineNumber(),
```
boulter: perhaps "Rendering cycle"?

PR #147, opened 2017-08-28
Add resolved expressions recursively up the context chain

jmp3833:
Rather than using `addResolvedFrom` to combine resolved expressions from multiple `Context` objects, it would be more widely applicable to recursively add resolved expressions up the parent context chain. This way information isn't lost when rendering a nested structure, with multiple nested contexts and resolved expressions. 

@boulter do you think this would be worth applying to `Context`'s resolved functions and values as well? 

cc @hs-lsong

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -3,6 +3,7 @@ package com.hubspot.jinjava.el;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
 
 import java.util.List;
+import java.util.Optional;
 
 import javax.el.ELException;
 import javax.el.ExpressionFactory;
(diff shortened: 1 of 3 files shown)


hs-lsong:
Does jinjava has the concept of scope? If so, a child's expression may not be the same as the parent's. I do not have the context of this issue. Why do we need this?

boulter:
I agree that we should do the same for function and variables too.

jmp3833:
@hs-lsong currently the `Context` objects aren't scoped, but a nested set of expressions will have a Context object applied to each expression. We would like the ability to traverse the entire collection of resolved expressions from any `Context` object from the root `Context`. 

This way we can perform selective logic if we detected a certain resolved expression, function, or value after processing a large chunk of text with JinJava. We had been using the [addResolvedFrom](https://github.com/HubSpot/jinjava/pull/133) function to apply values up the context chain but rather than doing this on a case by case basis, it would be nice to have all of these values applied to the root `Context` all the time.

PR #168, opened 2017-12-11
Add new EscapeJinjavaFilter

nbelisle11:
@jboulter 
I had issues rendering jinjava syntax text within a jinjava interpreted page and found this function missing from our library. Figured it was worth adding.


src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @Override
+  public Object filter(Object object, JinjavaInterpreter interpreter, String... arg) {
+    return escapeJinjavaEntities(Objects.toString(object, ""));
+  }
+
+  @Override
+  public String getName() {
+    return "escapeJinjava";
```
boulter: can you make this snake case?

PR #174, opened 2018-02-01
Don't put stack trace in the exception message

zbstof:
Exception stacktrace for each error could be easily extracted by doing
`((FatalTemplateErrorsException) e).getErrors().get(n).getException()`,
if needed


zbstof:
Problem for us was that we tried extract root cause's message ourselves, and this works ok except with this library, which caused buffer overflow down the line.

boulter:
Thanks, @zbstof!

PR #175, opened 2018-02-08
Add more sequence expression tests.

mattcoley:
Right now it is difficult to use `selectattr` or `if val is` for set functions. This adds three new expression tests: `containing`, `containingall`, and `within`.

- `containing`: if the sequence contains an element. `if [1, 2, 3] is containing 2)`.
- `containingall`, if the sequence contains all elements of another sequence. `if [1, 2, 3] is containingall [1, 2]`.
- `within`: if a value is in a sequence of elements. `if 1  is within [1, 2, 3])`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsContainingAllExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsContainingAllExpTest.java
@@ -6,27 +6,36 @@ import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.util.ForLoop;
 import com.hubspot.jinjava.util.ObjectIterator;
 
-public class IsInExpTest implements ExpTest {
+public class IsContainingAllExpTest implements ExpTest {
 
   @Override
   public boolean evaluate(Object var, JinjavaInterpreter interpreter, Object... args) {
 
-    if (args == null || args.length == 0) {
+    if (null == var || args.length == 0 || args[0] == null) {
       return false;
     }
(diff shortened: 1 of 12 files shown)


mattcoley:
We have no unit tests right now for these expression tests. Maybe I should just bite the bullet and write all of them? @boulter ?

src/main/java/com/hubspot/jinjava/lib/exptest/IsContainingAllExpTest.java
```diff
@@ -0,0 +1,41 @@
...
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  @Override
+  public String getName() {
+    return "containingall";
```
boulter: "containsall"?

mattcoley:
Ok @boulter tests are added for these new expression tests.

src/main/java/com/hubspot/jinjava/lib/exptest/IsContainingAllExpTest.java
mattcoley: Whereas `is containing`, `is containingall`, and `is within` are the best verbs I have thought of.

mattcoley: Or we could reserve the keyword `does` and do `if val1 does contain val2` :)

mattcoley: But that seems like overkill.

PR #178, opened 2018-02-26
Make int/float parsing locale aware

pfarrel:
Currently, the int and float filters always parse with the system default locale. This doesn't make much sense, when jinjava is otherwise locale aware. Instead this PR use a NumberFormat based on `JinjavaConfig::getLocale`.

This could definitely cause a change in behaviour for users who have been using a non-default locale, if that locale has different number formatting rules. 

What's protocol here for a version number bump @jboulter ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 
   <groupId>com.hubspot.jinjava</groupId>
   <artifactId>jinjava</artifactId>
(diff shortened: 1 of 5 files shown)


boulter:
Looks good. I think it makes sense to do a minor version bump.

anthmatic:
RE: #184 , is this breaking change expected?

PR #181, opened 2018-03-01
Enable blazar

stevegutz:
This will allow us to publish SNAPSHOTs internally rather than on sonatype and should help us avoid what appear to be occasional consistency issues with sonatype snapshots.

@boulter @pfarrel @mattcoley @jmp3833 @hs-lsong @jhaber


mattcoley:
🥇

Issue #180, opened 2018-03-01
Error parsing 'is not defined'


piandrus-n:
The following expression yields error:
{% if variable is not defined %}

com.hubspot.jinjava.interpret.TemplateSyntaxException: Error parsing ' variable is not defined ': syntax error at position 16, encountered 'is', expected '}'

The following is ok:
{% if not variable is defined %}

Both should be ok.

PR #187, opened 2018-04-17
Negated expressions

boulter:
Adds support for "not" as a prefix for expression tests such as

`{% if variable is not defined %}`

Fixes https://github.com/HubSpot/jinjava/issues/180

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/exptest/NegatedExpTestTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/exptest/NegatedExpTestTest.java
@@ -11,7 +11,7 @@ import com.hubspot.jinjava.Jinjava;
 
 public class NegatedExpTestTest {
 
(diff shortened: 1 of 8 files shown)


src/test/java/com/hubspot/jinjava/lib/exptest/NegatedExpTestTest.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  @Test
+  public void itNegatesDefined() {
+    assertThat(jinjava.render(String.format(TEMPLATE, "blah", "defined"), new HashMap<>())).isEqualTo("pass");
+  }
+
+  @Test
+  public void itNegatesContaining() {
+    assertThat(jinjava.render(String.format(CONTAINING_TEMPLATE, "[1, 2, 3]", "4"), new HashMap<>())).isEqualTo("pass");
```
mattcoley: Should we have a test that evaluates to `fail` to make sure that this statement doesn't always evaluate to true.

PR #188, opened 2018-04-17
Correctly report column positions of syntax errors

boulter:
Follows up on https://github.com/HubSpot/jinjava/pull/152. 

The syntax exceptions report only the position of the error within the expression, not within the entire rendered content. This adds the interpreter column position and the exception position together to report accurate column positions.


corymartin:
🔢 Thank you!

Issue #184, opened 2018-04-05
Float type coercion has changed


anthmatic:
Previously the `float` filter would return `0.0` when trying to coerce a non-numeric string. It looks like after #178 (change to `numberFormat.parse()`) strings that start with numbers can be converted to floating point numbers.

Before:
`"30%"|float == 0`

After #178 :
`"30%"|float == 30`

boulter:
I believe this was fixed by #191

PR #191, opened 2018-04-18
Make int and float parsing more strict

pfarrel:
Fix for https://github.com/HubSpot/jinjava/issues/184 reported by @anthmatic.
https://github.com/HubSpot/jinjava/pull/178 unintentionally made int and float parsing a lot more lax. This change restores some of the original strictness to match behaviour with `jinja` , while keeping support for our locale specific changes. It does so by:
- Checking that all of the input string was consumed by the parse operation, to ensure that trailing symbols or characters cause a parse to fail.
- Parsing integers first as a float, then converting to an integer, so that the above change doesn't prevent decimal expressions from parsing as integers.


mattcoley:
Can we add a test to make sure parsing values such as `1.100000` still works as expected? Should work now but just want to cover all bases for future changes.

PR #202, opened 2018-06-08
Wrap items in for loop with "PyIsh" equivalents

jessbrandi:
The `wrap` code in `JinjavaInterpreterResolver`, wraps objects in their `Pyish*` counterparts. However, if the value being wrapped was a list, the children of that value were not wrapped. This resulted in Date values being treated differently if they appeared as a single value or in a list.

My initial attempt was modify the list to wrap the children in the `wrap` method, but that resulted in issues modifying the list later on using `set` and `append` functions.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
@@ -17,6 +17,7 @@ package com.hubspot.jinjava.lib.tag;
 
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -129,6 +130,7 @@ public class ForTag implements Tag {
       LengthLimitingStringBuilder buff = new LengthLimitingStringBuilder(interpreter.getConfig().getMaxOutputSize());
       while (loop.hasNext()) {
         Object val = loop.next();
+        val = interpreter.resolveProperty(val, Collections.emptyList());
 
         // set item variables
         if (loopVars.size() == 1) {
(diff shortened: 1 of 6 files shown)


src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
```diff
@@ -193,6 +195,18 @@ public void testForLoopRangeWithStringsWithSpaces() {
       assertEquals("a b", rendered);
   }
 
+  @Test
+  public void testForLoopWithDates() {
+    Map<String, Object> context = Maps.newHashMap();
+    Date testDate = new Date();
+    context.put("the_list", Lists.newArrayList(new Date()));
```
nbelisle11: I think you want to use `testDate` when initializing `the_list` right?

jessbrandi: Yep, good catch. I guess this test worked anyway, because PyishDates aren't that precise? (`toString` only prints out to seconds, not millis).

src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
```diff
@@ -129,6 +130,7 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
       LengthLimitingStringBuilder buff = new LengthLimitingStringBuilder(interpreter.getConfig().getMaxOutputSize());
       while (loop.hasNext()) {
         Object val = loop.next();
+        val = interpreter.resolveProperty(val, Collections.emptyList());
```
boulter: I see what you're trying to do here, but it's weird to call `resolveProperty` with an empty list of properties to resolve. Perhaps it would make more sense just to add a delegate method in `JinjavaInterpreter` for `wrap`.

jessbrandi: 👍 updated

PR #211, opened 2018-07-17
Deepen equalto expression test comparison.

mattcoley:
Let's say you have a string `TEST` and an enum TEST. Under the usual `==` operation the expression evaluater will convert the enum to a string for comparison using `BooleanOperations`. This mimics this behavior in `equalto` expression test so there is no distinction between them.


mattcoley:
Done for general type coercions. However can't test the exact case I wanted to fix since I can't define enums in Jinjava right now.

PR #220, opened 2018-08-14
Selectattrfirst

hs-lsong:
Add a filter `selectattrfirst`. This is for porformance purpose. Sometimes, a larger number of items are using the `objects|selectattr|first`, which is not efficient because it go through all the items and select the attribute, and then returns the first one. This new filter will bail out at the first matching item that matches.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/FilterLibrary.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/FilterLibrary.java
@@ -50,6 +50,7 @@ public class FilterLibrary extends SimpleLibrary<Filter> {
         RejectFilter.class,
         SelectFilter.class,
         SelectAttrFilter.class,
+        SelectAttrFirstFilter.class,
         SliceFilter.class,
         ShuffleFilter.class,
         SortFilter.class,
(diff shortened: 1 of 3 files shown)


mattcoley:
What if we lazily compute some filters instead? When we apply a function to a sequence, we wrap the sequence with a function and don't apply the filter until an item property is accessed? That way we don't have to create a new function for every optimization.

hs-lsong:
Can you elaborate? How would a lazy computation work here?

hs-lsong:
It is a very good suggestion. I have played a bit. It seems the hurdle is how to do hasNext(). Some collections function will just check hasNext(), then happily accept any value from next().

hs-lsong:
Played more with a peekable iterator. It is mostly working. However, you cannot loop multiple times with the iterator, while the list can. Let's get this in for the specific optimization, while continue to find solutions for generic cases.

hs-lsong:
@boulter What do you think?

boulter:
I agree that using an iterator would be much better than introducing another proprietary filter.

hs-lsong:
I agree too, and I tried to implement that. But an iterator does not support multiple loop iterations, it breaks some of the current usage, not only loop multiple times, and also has some random access.

boulter:
Can I see a PR where the iterator breaks things?

hs-lsong:
https://github.com/HubSpot/jinjava/pull/226
The specific file is FilteringForLoop.

boulter:
@hs-lsong can we close this?

PR #221, opened 2018-08-16
Detect fromTag cycle.

hs-lsong:
Recursive fromTag will cause stack overflow.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -392,6 +392,7 @@ public class Context extends ScopeMap<String, Object> {
   public void pushFromStack(String path, int lineNumber, int startPosition) {
     fromStack.push(path, lineNumber, startPosition);
   }
+
   public void popFromStack() {
     fromStack.pop();
   }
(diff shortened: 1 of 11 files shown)


src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -377,10 +381,21 @@ public CallStack getIncludePathStack() {
...
+  }
+
   public CallStack getMacroStack() {
     return macroStack;
   }
 
+  public void pushFromStack(String path, int lineNumber, int startPosition) {
+    fromStack.push(path, lineNumber, startPosition);
+  }
+  public void popFromStack() {
```
mattcoley: nit: missing newline.

src/test/java/com/hubspot/jinjava/lib/tag/FromTagTest.java
```diff
@@ -54,6 +56,13 @@ public void importedContextExposesVars() {
         .contains("wrap-padding: padding-left:42px;padding-right:42px");
   }
 
+  @Test
+  public void importedCycleDected() {
```
mattcoley: Is it possible for A to import from B which imports from C which imports from A? If so can we have a test to make sure it catches this cycle as well?

PR #222, opened 2018-08-16
Limit the number errors.

hs-lsong:
We could get a very larger number of error store in the memory. Limited number of errors should be enough for debugging purpose.


hs-lsong:
I don't see why that is helpful.

mattcoley:
A code editor could show everything below that line as a yellow warning.

hs-lsong:
It is easy to add if the code editor wants that.

PR #224, opened 2018-08-17
Support arguments for macro block callbacks

leromarinvit:


leromarinvit:
I'm a bit confused by the findbugs message:

> [INFO] Redundant nullcheck of argName which is known to be null in com.hubspot.jinjava.lib.tag.CallTag.interpret(TagNode, JinjavaInterpreter) [com.hubspot.jinjava.lib.tag.CallTag] Redundant null check at CallTag.java:[line 118] RCN_REDUNDANT_NULLCHECK_OF_NULL_VALUE

`CallTag.java:118` is a closing tag of a try-with-resources, and `interpret` doesn't seem to contain any explicit null checks at all. Running `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` locally doesn't yield any errors either...

mattcoley:
I can try pulling your branch later today and see what I find.

leromarinvit:
Found it: https://sourceforge.net/p/findbugs/bugs/1169/
Supposedly it's fixed, but at least the workaround listed there works.

PR #229, opened 2018-08-24
Add lazy iterator for selectattr filter.

mattcoley:
WIP to lazily evaluate list filters.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/LazyList.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/LazyList.java
@@ -1,0 +1,315 @@
+package com.hubspot.jinjava.objects.collections;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Set;
+
+import com.google.common.collect.Iterators;
+
+public class LazyList<T> implements List<T> {
+
+  private ArrayWrappedIterator<T> iterator;
+  private ArrayList<T> list;
+
+  public LazyList(Iterator<T> iterator) {
+    this.list = new ArrayList<>();
+    this.iterator = new ArrayWrappedIterator<>(iterator, this.list);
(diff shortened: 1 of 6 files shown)


mattcoley:
@hs-lsong @boulter I still have to fix some code so the list test suite passes but thoughts on general implementation?

src/main/java/com/hubspot/jinjava/objects/collections/LazyList.java
```diff
@@ -0,0 +1,361 @@
...
+
+  @Override
+  public boolean isEmpty() {
+    return !iterator.hasNext() && list.isEmpty();
+  }
+
+  @Override
+  public boolean contains(Object o) {
+
+    for (int i = 0; i < list.size(); i++) {
```
boulter: hopefully this doesn't get called much 😢

src/main/java/com/hubspot/jinjava/objects/collections/LazyList.java
```diff
@@ -0,0 +1,361 @@
...
+    iterator.unpack();
+    return list.hashCode();
+  }
+
+  @Override
+  public boolean removeIf(Predicate<? super T> filter) {
+    iterator.unpack();
+    return list.removeIf(filter);
+  }
+}
```
boulter: where did you pull this code from?

src/test/java/com/hubspot/jinjava/objects/collections/LazyListTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.common.collect.testing.features.ListFeature;
+
+@RunWith(AllTests.class)
+public class LazyListTest {
+
+  public static junit.framework.TestSuite suite() {
+    return ListTestSuiteBuilder.using(new TestStringListGenerator() {
+      @Override
+      protected List<String> create(String[] elements) {
+        // Fill here your collection with the given elements
```
boulter: definitely would be good to test that it doesn't evaluate more than it needs to here.

hs-lsong:
Cool, looks good.

PR #232, opened 2018-08-31
disallow returning Class objects

boulter:
Class objects have security issues.


src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -20,7 +20,7 @@
   private static final Set<String> RESTRICTED_METHODS = ImmutableSet.<String> builder()
       .add("clone")
       .add("hashCode")
-      .add("getClass")
```
kenbreeman: let's leave this in, layers are good

boulter: ok, I took it out mostly so I could test the new method of blocking, but I'll put it back.

Issue #169, opened 2017-12-11
jinjava // either buggy or unimplemented


prior:
from [jinja math docs](http://jinja.pocoo.org/docs/2.10/templates/#math):
> //
Divide two numbers and return the truncated integer result. {{ 20 // 7 }} is 2.

However attempting this with jinjava throws an exception.  For example, I would expect this template:
```
20 // 7 = {{ 20 // 7 }}
```
to return `20 // 7 = 2`, but instead it throws this:
```
com.hubspot.jinjava.interpret.InterpretException: Error resolving expression [20 // 7]: NullPointerException: 
	at com.hubspot.jinjava.el.ExpressionResolver.resolveExpression(ExpressionResolver.java:89)
	at com.hubspot.jinjava.interpret.JinjavaInterpreter.resolveELExpression(JinjavaInterpreter.java:397)
```
```
Caused by: java.lang.NullPointerException
	at jinjava.de.odysseus.el.tree.impl.Parser.mul(Parser.java:547)
	at com.hubspot.jinjava.el.ext.ExtendedParser.add(ExtendedParser.java:149)
	at jinjava.de.odysseus.el.tree.impl.Parser.cmp(Parser.java:462)
```

boulter:
Looks like this works for variables as params, just not literals. I'm looking at fixing.

boulter:
Also cc @markus-s24

bkrainer:
fixed in #234.

PR #234, opened 2018-09-06
update ** and // to support int/float literals

bkrainer:
These changes address issue #169.

The current logic for // and ** specifies that the operands need to be `IDENTIFIER` symbols. This works when jinjava variables are used, but fails for actual numerical values (as these are `INTEGER` symbols, not `IDENTIFIER`).

The fix for this was to remove the special // and ** logic in `ExtendedParser.value` and instead rely on the existing extension framework via the pre-existing calls to `Parser.getExtensionHandler`


bkrainer:
@boulter could you add me as a contributor to this repo? I don't have permissions to merge the PR

mattcoley:
Done

Issue #236, opened 2018-09-19
Filter ipaddr always return false -> returns nothing


Devl11:
Hi,
I am trying to use the ipaddr filter but it does not return any value.

``` java
Jinjava jinjava = new Jinjava();
String template = "hello {{ testvar | ipaddr }}";

HashMap values = new HashMap();
values.put("testvar", "127.0.0.1");
jinjava.render(template, values);  
```

It only returns "hello ". 
Am I missing something?

mattcoley:
`ipaddr` has not been implemented yet in jinjava.

Devl11:
Well thats explains it!
What I really wanted was to use ``ìpaddr('prefix')```, does jinjava have any similar filter?

mattcoley:
Shouldn't take too long to add the `ipaddr` filter.

Devl11:
Is it possible to add the ```ipaddr('prefix')``` etc to get the prefix from an ip? 
Like ```{{ '192.0.2.0/24' | ipaddr('prefix') }} ``` which return 24

MikaelMollberg:
> Is it possible to add the `ipaddr('prefix')` etc to get the prefix from an ip?
> Like `{{ '192.0.2.0/24' | ipaddr('prefix') }}` which return 24

That would be great!

PR #237, opened 2018-09-19
Adds `ipaddr` filter to test if string is valid IP Address.

mattcoley:
Support for https://github.com/HubSpot/jinjava/issues/236

Adds `ipaddr` filter to test if string is valid IPv4 or IPv6 filter. Checking IPv6 is tricky due to the compression, however the regex seems to be working against the tests.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
@@ -34,7 +34,7 @@ public class IpAddrFilter implements Filter {
     }
 
     if (object instanceof String) {
-      String address = (String) object;
+      String address = ((String) object).trim();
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public Object filter(Object object, JinjavaInterpreter interpreter, String... arg) {
+
+    if (object == null) {
+      return false;
+    }
+
+    if (object instanceof String) {
+      String address = (String) object;
```
boulter: Should this do a `trim()` first?

mattcoley: Yea that would be smart.

PR #238, opened 2018-10-08
Implement nested properties for selectattr and rejectattr filters.

mattcoley:
Continuation of https://github.com/HubSpot/jinjava/pull/183 to make these filters expressive. Previously if you had a nested object there would be no way to `selectattr` on the nested property, thereby requiring the need to use a for loop or make another API call. This changes `selectattr` to use the chain property resolver and modifies `rejectattr` to share code with `selectattr`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.base.Splitter;
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
```diff
@@ -71,9 +76,10 @@ public Object filter(Object var, JinjavaInterpreter interpreter, Object[] args,
     ForLoop loop = ObjectIterator.getLoop(var);
     while (loop.hasNext()) {
       Object val = loop.next();
-      Object attrVal = interpreter.resolveProperty(val, attr);
 
-      if (expTest.evaluate(attrVal, interpreter, expArgs)) {
+      Object attrVal = new Variable(interpreter, String.format("%s.%s", "placeholder", attr)).resolve(val);
+      boolean pass = expTest.evaluate(attrVal, interpreter, expArgs);
+      if ((acceptObjects && pass) || (!acceptObjects && !pass)) {
```
boulter: could just do `acceptObjects == pass`

mattcoley:
Anthony mentioned he would update the public documentation. When using `selectattr` on a list of objects, you could only apply an expression test to a top level property. If an object is composed of other objects, you would not be able to filter on those objects' properties.

boulter:
Ok, I get it now. This allows you to filter by `option.id` instead of just `option` for example.

mattcoley:
Yep 👍 On the HubSpot side, it will encourage people to use HubL instead of multiple API calls to do filtering on HubDB rows.

PR #242, opened 2018-10-12
remove findbugs annotations because there's no findbugs analysis happening

dbyron0:
Seems findbugs analysis was on in the past (https://github.com/HubSpot/jinjava/commit/43c7e72d225842747fe2821de096b41c00c746f7), but has since been disabled (maybe https://github.com/HubSpot/jinjava/commit/4916b55ce1ef353d4f3e7d882222b00e927ba022#diff-600376dffeb79835ede4a0b285078036, maybe earlier).  In any case, it doesn't appear that findbugs is running at the moment (or at least, removing the suppression doesn't cause any failures for me).

It'd be wonderful to get com.google.code.findbugs:annotations off the classpath as it causes conflicts for me.  With the [migration to spotbugs](https://spotbugs.readthedocs.io/en/stable/migration.html) it seems like conflicts are all too easy at the moment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,10 +65,6 @@
       <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>annotations</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


dbyron0:
And....apparently travis is telling me that I was wrong.  Sorry for the noise.

PR #243, opened 2018-10-14
Add 'prefix' function to ipaddr filter.

mattcoley:
Final request for ipaddr filter https://github.com/HubSpot/jinjava/issues/236.

Returns the CIDR prefix of a valid IP address. If the IP address is invalid or the prefix is not a number it will return null. Example usage:
`'1.0.0.1/24'|ipaddr('prefix')` would return `24`.


boulter:
The only problem with this is that it creates more reasons to build uncacheable pages. :(

PR #245, opened 2018-10-18
support negative list indeces

dbyron0:
so e.g.

{{ (split_me|split('-'))[-2] }}

works instead of

{% set parts = split_me|split('-') %}{{ parts[parts|length - 2] }}"

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
@@ -34,10 +34,9 @@ public class JinjavaListELResolver extends ListELResolver {
       // If we're dealing with a negative index, convert it to a positive one.
       if (isResolvable(base)) {
         int index = toIndex(property);
-        List<?> list = (List<?>) base;
         if (index < 0) {
           // Leave the range checking to the superclass.
-          property = index + list.size();
+          property = index + ((List<?>) base).size();
         }
       }
       return super.getValue(context, base, property);
--- a/src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
@@ -31,73 +31,25 @@ public class JinjavaListELResolver extends ListELResolver {
   @Override
   public Object getValue(ELContext context, Object base, Object property) {
     try {
-      final Object superValue = super.getValue(context, base, property);
-      if (superValue != null) {
-        return superValue;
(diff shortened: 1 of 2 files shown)


mattcoley:
It seems easier to first check if the value is resolvable to a list. If so, call `toIndex(property)`. If that value is negative, then `property += value.size();`

dbyron0:
I was trying to tread as lightly as possible -- to leave as much to the superclass as possible.  If you're OK with totally taking over, which is what I think you're suggesting, I'll give it a shot.

I could use a hand coming up with test cases that cover the non-string parts of toIndex.

dbyron0:
@mattcoley like this? https://github.com/HubSpot/jinjava/pull/245/commits/7d7949265b8976b083553be80506801a807ced91

src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
```diff
@@ -30,12 +31,61 @@ public boolean isReadOnly(ELContext context, Object base, Object property) {
...
     try {
+      // If we're dealing with a negative index, convert it to a positive one.
+      if (isResolvable(base)) {
+        int index = toIndex(property);
+        List<?> list = (List<?>) base;
+        if (index < 0) {
+          // Leave the range checking to the superclass.
+          index += list.size();
+        }
+        return super.getValue(context, base, index);
```
mattcoley: we can just set `property = index + list.size()` and just have one `super.getValue()` outside the if

mattcoley:
Yea that approach is a bit cleaner.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaListELResolver.java
```diff
@@ -30,12 +31,60 @@ public boolean isReadOnly(ELContext context, Object base, Object property) {
   @Override
   public Object getValue(ELContext context, Object base, Object property) {
     try {
+      // If we're dealing with a negative index, convert it to a positive one.
+      if (isResolvable(base)) {
+        int index = toIndex(property);
+        List<?> list = (List<?>) base;
+        if (index < 0) {
+          // Leave the range checking to the superclass.
+          property = index + list.size();
```
boulter: `list` can be inlined.

dbyron0: https://github.com/HubSpot/jinjava/pull/245/commits/6cda4efc04ad96459c8ad4287ec0c88a454cd931

dbyron0:
Thanks much.  Would love to see this merged and a new version released.

dbyron0:
Thanks for merging!

PR #249, opened 2018-11-02
Support expressions in selectattr/rejectattr

mattcoley:
Previous PR allowed nested properties in `selectattr` and `rejectattr` https://github.com/HubSpot/jinjava/pull/238.

This is great however is still limiting because you cannot execute filters. For example you may have a list of objects you would like to map to a value before executing the filter: `|selectattr("optionList|map('id')", "containing", 1)` or perhaps you would like to trim the values `|selectattr("name|trim|lower", "equalto", "matt")`. 

This resolves the first argument as an expression and then applies the test case. The solution is hacky as it stores a temporary variable to resolve the expression against. I have not thought of a way to get around this.

Motivation: This makes it easier to filter many HubL/HubSpot objects that have properties that are lists of other objects. Previously this required a for loop.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
@@ -79,18 +79,19 @@ public class SelectAttrFilter implements AdvancedFilter {
 
       // push temporary variable to be resolved
       String tempValue = generateTempVariable();
-      while (interpreter.getContext().containsKey(tempValue)) {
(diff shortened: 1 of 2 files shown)


TheWebTech:
Understand the performance issues. Perhaps there's a way to allow some of the functionality without allowing everything.
Change:
`|selectattr("name|trim|lower", "equalto", "matt")`
 to
`|selectattr("name", "equalto", "matt",["trim","lower"])`

The 4th parameter being permitted filters to apply to the value of the attribute.  Tou could then restrict which filters were applied.

PR #252, opened 2018-11-09
Adds a new regex_replace filter

### LINKED ISSUE #251: Support regex in replace filter OR implement regex_replace filter
It's currently not possible to use a regular expression and remove unwanted characters from a string. 

**Example**:  This would be helpful for taking a user generated string and converting it into an alphanumeric URL. Transform this `"How I saved $300k by switching to this obscure insurance company!"` into this: `"how-i-saved-300k-by-switching-to-this-obscure-insurance-company"`

There is no way to strip out special characters without chaining a bunch of `replace()` filters.

anthmatic:
Adds a new filter called `regex_replace()` that allows the template creator to match a regular expression and replace it with a new string.

Resolves #251

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,10 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.re2j</groupId>
+      <artifactId>re2j</artifactId>
+    </dependency>
 
     <dependency>
       <groupId>ch.qos.logback</groupId>
@@ -120,6 +124,11 @@
         <groupId>de.odysseus.juel</groupId>
         <artifactId>juel-impl</artifactId>
         <version>2.2.7</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.re2j</groupId>
+        <artifactId>re2j</artifactId>
+        <version>1.2</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,10 @@
       <artifactId>juel-impl</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.re2j</groupId>
+      <artifactId>re2j</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
@@ -75,10 +79,6 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.google.re2j</groupId>
-      <artifactId>re2j</artifactId>
     </dependency>
 
     <dependency>
--- a/src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
@@ -1,16 +1,19 @@
 package com.hubspot.jinjava.lib.filter;
 
+import com.google.re2j.Matcher;
+import com.google.re2j.Pattern;
+import com.google.re2j.PatternSyntaxException;
+
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.InterpretException;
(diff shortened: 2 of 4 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        if (args.length < 2) {
+            throw new InterpretException("filter " + getName() + " requires two string args");
+        }
+
+        String s = (String) var;
+        String toReplace = args[0];
+        String replaceWith = args[1];
+
+        return s.replaceAll(toReplace, replaceWith);
```
mattcoley: This may throw a `PatternSyntaxException` we should catch and add to the template errors.

src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
```diff
@@ -0,0 +1,50 @@
...
+                         String... args) {
+
+        if (var == null) {
+            return null;
+        }
+        if (args.length < 2) {
+            throw new InterpretException("filter " + getName() + " requires two string args");
+        }
+
+        String s = (String) var;
```
mattcoley: What if the variable isn't a string?

src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
```diff
@@ -0,0 +1,50 @@
...
+        String toReplace = args[0];
+        String replaceWith = args[1];
+
+        return s.replaceAll(toReplace, replaceWith);
+    }
+
```
mattcoley: nit: extra newline

src/test/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilterTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public void noopOnNullExpr() {
+        assertThat(filter.filter(null, interpreter, "foo", "bar")).isNull();
+    }
+
+    @Test
+    public void replaceString() {
+        assertThat(filter.filter("It costs $300", interpreter, "[^a-zA-Z]", "")).isEqualTo("Itcosts");
+    }
+
```
mattcoley: I would add a test for invalid regex

src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
boulter: I'd like to use https://github.com/google/re2j so we don't have to worry about regexes consuming an infinite amount of resources here.

src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.InterpretException;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+
+@JinjavaDoc(
+        value = "Return a copy of the value with all occurrences of a matched regular expression replaced with a new one. " +
+                "The first argument is the regular expression to be matched, the second is the replacement string",
+        params = {
```
boulter: We should specify which flavor of regex we're supporting here since they differ between python, Java, etc.

mattcoley:
Looks good, only thing I am worried about is bad regex blowing up the interpreter. I wonder if there are ways to limit regex user input.

pom.xml
```diff
@@ -121,6 +125,11 @@
         <artifactId>juel-impl</artifactId>
         <version>2.2.7</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.re2j</groupId>
+        <artifactId>re2j</artifactId>
+        <version>1.2</version>
```
mattcoley: duplicated dependency?

boulter: no, this is specifying the version.

src/main/java/com/hubspot/jinjava/lib/filter/RegexReplaceFilter.java
boulter: Add a link to the supported syntax? https://github.com/google/re2/wiki/Syntax

pom.xml
anthmatic: So, is this the correct way to include a dep and specify a version? If there is a better way, I can adjust

boulter: You can just put the version in with the dependency. I'm not sure why some are separated out in this pom.

Issue #56, opened 2016-04-27
Add support for converting string variables to timestamps


davidjosephhunt:
While you can use the datetimeformat filter to format timestamps, there is currently no way to take a convert variables into datetime objects. For example, if you wanted to calculate 30 days before local_dt, there isn't any way to take a string or numeric variable and convert it to a datetime object.

It would be great if a filter or function could be added to support string/int to datetime object conversion.

PR #258, opened 2018-11-27
Adds filters for datetime functions.

mattcoley:
Adds `plus_time`, `minus_time`, and `between_times` filters to operate on datetime objects/timestamp. Examples: `date|plus_time('1', 'days')` will give you a date 1 day in the future; `begin|between_times(end, 'minutes')` will give you the number of minutes between those two dates.

I also added a `strtotime` to convert a string and datetime format to a datetime object. For example, `"2018-07-14T14:31:30+0530"|strtotime("yyyy-MM-dd'T'HH:mm:ssZ")|unixtimestamp` will output `1531558890000`

Fixes https://github.com/HubSpot/jinjava/issues/56

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/BaseDateFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/BaseDateFilter.java
@@ -40,5 +40,4 @@ public abstract class BaseDateFilter implements Filter {
     }
     return unitMap.get(lowercase);
   }
-
 }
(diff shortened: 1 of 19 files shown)


boulter:
Do these have any precedent in Jinja? I would prefer to keep it as compatible as possible. If not, perhaps these could filters could be named with a little more descriptively like `plus_time`, `minus_time` and `between_times`.

mattcoley:
The recommended way in jinja is with python commands. We could encourage developers to use Java methods like `plusDays()` but those seem like the only options.

mattcoley:
I changed this to support all [ChronoUnit](https://docs.oracle.com/javase/8/docs/api/java/time/temporal/ChronoUnit.html) enum values and added a link at the top of each class.

PR #260, opened 2018-12-11
Add conversion to Java datetime format for strtotime.

mattcoley:
We need to convert the pattern for `strtotime` and `datetimeFormat` to be unified.


mattcoley:
Also as part of https://issues.hubspotcentral.com/browse/CG-11772 I noticed that we are converting `Z` to `ZZZ` which does not make sense. We should be converting `Z` to `z` to fit with Java formatting.

Issue #261, opened 2018-12-18
No filter to intersect two lists


jfudman:
Currently there is no way to get the intersection of 2 lists (example below).
```jinja2
{% set list1 = [ 'a', 'b', 'c' ] %}
{% set list2 = [ 'c', 'd' ] %}
{{ list1 | intersect(list2) }} {# would output ['c'] #}
```

See the[ set theory filters for ansible](https://ansible-docs.readthedocs.io/zh/stable-2.0/rst/playbooks_filters.html#set-theory-filters) as a good example.

### A more elaborate example below
```{% set section_active = false %}
{% set tags = content.tagList|map('slug') %}
{% set list2 =  [ 'foo','poo', 'doo', 'boo' ] %}
{% set joined =  tags|intersect(list2) %}
{% unless joined == [] %}
  {% set section_active = true %}
{% endunless %}
```

mattcoley:
Shouldn't be too hard to support those set functions.

PR #262, opened 2018-12-18
Add common set theory filters.

mattcoley:
Fixes https://github.com/HubSpot/jinjava/issues/261
Adds `union`, `intersect`, `difference`, and `symmetric_difference` similar to Ansible https://ansible-docs.readthedocs.io/zh/stable-2.0/rst/playbooks_filters.html#set-theory-filters.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DifferenceFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DifferenceFilter.java
@@ -1,20 +1,16 @@
 package com.hubspot.jinjava.lib.filter;
 
 import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.Map;
-import java.util.Set;
 
+import com.google.common.collect.Sets;
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
-import com.hubspot.jinjava.util.ForLoop;
-import com.hubspot.jinjava.util.ObjectIterator;
 
 @JinjavaDoc(
-    value = "Returns a list containing elements present in the first list and not the second list",
+    value = "Returns a list containing elements present in the first list but not the second list",
     params = {
         @JinjavaParam(value = "value", type = "sequence", desc = "The first list"),
         @JinjavaParam(value = "list", type = "sequence", desc = "The second list")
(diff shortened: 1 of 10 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/DifferenceFilter.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.util.ForLoop;
+import com.hubspot.jinjava.util.ObjectIterator;
+
+@JinjavaDoc(
+    value = "Returns a list containing elements present in the first list and not the second list",
```
boulter: _but_ not

src/main/java/com/hubspot/jinjava/lib/filter/UnionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+    snippets = {
+        @JinjavaSnippet(
+            code = "{{ [1, 2, 3]|union([2, 3, 4]) }}")
+    })
+public class UnionFilter implements AdvancedFilter {
+
+  @Override
+  public Object filter(Object var, JinjavaInterpreter interpreter, Object[] args, Map<String, Object> kwargs) {
+
+    LinkedHashSet<Object> result = new LinkedHashSet<>();
```
boulter: How about using `Sets.union`?

src/main/java/com/hubspot/jinjava/lib/filter/SymmetricDifferenceFilter.java
```diff
@@ -0,0 +1,68 @@
...
+    snippets = {
+        @JinjavaSnippet(
+            code = "{{ [1, 2, 3]|symmetric_difference([2, 3, 4]) }}")
+    })
+public class SymmetricDifferenceFilter implements AdvancedFilter {
+
+  @Override
+  public Object filter(Object var, JinjavaInterpreter interpreter, Object[] args, Map<String, Object> kwargs) {
+
+    Set<Object> keysFirst = new HashSet<>();
```
boulter: `Sets.symmetricDifference`?

src/main/java/com/hubspot/jinjava/lib/filter/IntersectFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    snippets = {
+        @JinjavaSnippet(
+            code = "{{ [1, 2, 3]|intersect([2, 3, 4]) }}")
+    })
+public class IntersectFilter implements AdvancedFilter {
+
+  @Override
+  public Object filter(Object var, JinjavaInterpreter interpreter, Object[] args, Map<String, Object> kwargs) {
+
+    Set<Object> keys = new HashSet<>();
```
boulter: `Sets.intersection`?

src/main/java/com/hubspot/jinjava/lib/filter/DifferenceFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    })
+public class DifferenceFilter implements AdvancedFilter {
+
+  @Override
+  public Object filter(Object var, JinjavaInterpreter interpreter, Object[] args, Map<String, Object> kwargs) {
+
+    Set<Object> keys = new HashSet<>();
+    Set<Object> result = new LinkedHashSet<>();
+
+    Object attr = null;
```
boulter: `Sets.difference`?

src/main/java/com/hubspot/jinjava/lib/filter/UnionFilter.java
mattcoley: The reason I did not use the set function helpers is that they would be slower/require more space. I would need to loop over the two iterables to form the sets, then apply those functions. This wouldn't be noticeable for small sets but might be for large sets. The functions are pretty simple anyways, and I'm not using the set helper functions is worth it but it's up to you.

mattcoley:
@boulter Refactored to use the `Sets` functions to simplify the code.

PR #264, opened 2019-01-02
Add validation mode

boulter:
Currently jinjava only parses expressions inside active code paths. This is efficient, but also does not look at expressions that might have syntax errors. This adds a validation mode to the interpreter that parses inactive code blocks for syntax, but does not execute any macros, functions or filters. 

This allows full syntax checking of templates by calling `setValidationMode(true)` when building the context.


mattcoley:
Only issue I can think of is if a later expression becomes invalid due to not evaluating a macro or set tag.

boulter:
> Only issue I can think of is if a later expression becomes invalid due to not evaluating a macro or set tag.

Any example you can think of? This only checks the syntax, not the expression values in validation mode.

mattcoley:
Yea just realized this was for syntax only which should be safe.

src/main/java/com/hubspot/jinjava/lib/tag/IfTag.java
```diff
@@ -58,46 +58,51 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
...
+    } finally {
+      interpreter.getContext().setValidationMode(parentValidationMode);
     }
 
-    return null;
+    return sb.toString();
   }
 
-  protected boolean evaluateIfElseTagNode(TagNode tagNode, JinjavaInterpreter interpreter) {
-    if (tagNode.getName().equals(ElseTag.ELSE)) {
```
hs-lsong: Is this checking hurting?

boulter: I don't think it's necessary anymore. `isPositiveIfElseNode` is only called once at the top of the `if` statement and again when an `elseif` is encoutered.

src/main/java/com/hubspot/jinjava/lib/tag/PrintTag.java
```diff
@@ -26,9 +26,9 @@ public String getName() {
   }
 
   @Override
-  public String interpret(TagNode tagNode,
-      JinjavaInterpreter interpreter) {
-    return Objects.toString(interpreter.resolveELExpression(tagNode.getHelpers(), tagNode.getLineNumber()), "");
+  public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
+    String result = Objects.toString(interpreter.resolveELExpression(tagNode.getHelpers(), tagNode.getLineNumber()), "");
+    return interpreter.getContext().isValidationMode() ? result : "";
```
hs-lsong: Is the `PrintTag` for `ValidationMode` only?

hs-lsong:
Will `ForTag` be treated differently?

src/main/java/com/hubspot/jinjava/lib/tag/PrintTag.java
boulter: Nope, it's a little-known tag. 

`{% print "hello" %}`

boulter:
> Will ForTag be treated differently?

You're right. I need to do this for `for`, `while`, and `do`.

Issue #263, opened 2018-12-30
Range function uses end=0 instead of end=RANGE_LIMIT when end exceeds max int


MosheElisha:
Hi,

When using a number (e.g., 9999999999) that exceeds Integer.MAX_VALUE (2147483647) - the range function uses an "end" of 0.

Unlike invalid number inputs (e.g. "abc") I believe that in this case, the "end" should be set to RANGE_LIMIT as the value is a valid number (although not a valid int).

To reproduce:

`{% for counter in range(1, 9999999999) %}`

The issue is caused because the code uses NumberUtils.toInt(...) which return 0 if the int conversion failed.

https://github.com/HubSpot/jinjava/blob/fa2cb672afe9a59254b96ecb61c2deca89258bdc/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java#L266

mattcoley:
Maybe we could parse these into `BigInteger` so we get arbitrary precision. This would mimic the behavior of Jinja with Python's arbitrary precision numbers. @boulter

boulter:
Thanks for the issue. I've PRed a fix in https://github.com/HubSpot/jinjava/pull/265

PR #265, opened 2019-01-03
handle large values in range function

boulter:
Fixes https://github.com/HubSpot/jinjava/issues/263 by defaulting numeric, out of range values in the range function to `start + RANGE_LIMIT`


boulter:
@MosheElisha is this what you expected?

MosheElisha:
@boulter Yes. That's great! Thanks!

PR #275, opened 2019-01-17
Add max string length to Jinjava config.

mattcoley:
Adds a max string length to the config. As a test, I have used this setting to limit the output of the `Join` filter, however it can be used in other filters as well if the approach looks fine.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
@@ -75,7 +75,7 @@ public class JoinFilter implements Filter {
         interpreter.addError(new TemplateError(ErrorType.WARNING,
             ErrorReason.OTHER,
             ErrorItem.FILTER,
-            String.format("Result of %s filter has been truncuated to the max String length of %d", getName(), interpreter.getConfig().getMaxStringLength()),
+            String.format("Result of %s filter has been truncated to the max String length of %d", getName(), interpreter.getConfig().getMaxStringLength()),
             null,
             interpreter.getLineNumber(),
             interpreter.getPosition(),
--- a/src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
@@ -7,6 +7,10 @@ import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.OutputTooBigException;
+import com.hubspot.jinjava.interpret.TemplateError;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorItem;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorType;
 import com.hubspot.jinjava.util.ForLoop;
 import com.hubspot.jinjava.util.LengthLimitingStringBuilder;
 import com.hubspot.jinjava.util.ObjectIterator;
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
```diff
@@ -52,17 +56,36 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
+        if (!first) {
+          stringBuilder.append(separator);
+        } else {
+          first = false;
+        }
+        stringBuilder.append(Objects.toString(val, ""));
+      } catch (OutputTooBigException ex) {
+        interpreter.addError(new TemplateError(ErrorType.WARNING,
+            ErrorReason.OTHER,
+            ErrorItem.OTHER,
```
hs-lsong: ErrorItem.FILTER

src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
```diff
@@ -52,17 +56,36 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
-      vals.add(Objects.toString(val, ""));
+      try {
+        if (!first) {
+          stringBuilder.append(separator);
+        } else {
+          first = false;
+        }
+        stringBuilder.append(Objects.toString(val, ""));
+      } catch (OutputTooBigException ex) {
+        interpreter.addError(new TemplateError(ErrorType.WARNING,
```
hs-lsong: This seems FATAL.

src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
```diff
@@ -52,17 +56,36 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
+    boolean first = true;
     while (loop.hasNext()) {
       Object val = loop.next();
 
       if (attr != null) {
         val = interpreter.resolveProperty(val, attr);
       }
 
-      vals.add(Objects.toString(val, ""));
+      try {
```
hs-lsong: You can also simple sum the length of each string and compare to the limit. It may look simpler. Up to you.

src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
mattcoley: Not necessarily FATAL IMO since it will still output a string, just shorter than expected.

src/main/java/com/hubspot/jinjava/lib/filter/JoinFilter.java
mattcoley: That's how the modified string builder works and throws the exception once the length > limit. I didn't see the need to duplicate the code we already have in there.

src/test/java/com/hubspot/jinjava/lib/filter/JoinFilterTest.java
```diff
@@ -32,6 +34,17 @@ public void testJoinAttrs() {
         .isEqualTo("foo, bar");
   }
 
+  @Test
+  public void itLimitsString() {
```
boulter: nit: itLimitsStringLength

mattcoley:
Yes I am planning adding this elsewhere once the approach is approved.

src/test/java/com/hubspot/jinjava/lib/filter/JoinFilterTest.java
```diff
@@ -32,6 +34,17 @@ public void testJoinAttrs() {
...
+    jinjava = new Jinjava(JinjavaConfig.newBuilder()
+        .withMaxStringLength(5)
+        .build());
+
+    RenderResult result = jinjava.renderForResult("{{ [1, 2, 3, 4, 5]|join('|') }}", new HashMap<String, Object>());
+    assertThat(result.getOutput()).isEqualTo("1|2|3");
+    assertThat(result.getErrors().size()).isEqualTo(1);
```
boulter: it would be good to verify that error is what you expect.

PR #276, opened 2019-01-22
support negative indeces in list slices

dbyron0:
so e.g.

{{ (split_me|split('-'))[0:-2] }}

works

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
@@ -9,8 +9,8 @@ import javax.el.ELContext;
 import javax.el.ELException;
 import javax.el.PropertyNotFoundException;
 
+import com.google.common.collect.Iterables;
 import com.hubspot.jinjava.objects.collections.PyList;
-
 import de.odysseus.el.misc.LocalMessages;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
```diff
@@ -72,6 +72,17 @@ public Object eval(Bindings bindings, ELContext context) {
     PyList result = new PyList(new ArrayList<>());
     int index = 0;
 
+    // Handle negative indeces.
+    if ((startNum < 0) || (endNum < 0)) {
+       int size = Iterables.size(baseItr);
```
dbyron0: If it's cheaper enough to use array length here I can special case that..

mattcoley: `Iterables.size()` has some smart logic to determine the fastest way to calculate the size depending on the underlying iterable implementation IIRC.

dbyron0: Sorry, ignore me.  Arrays turn into lists which are cheap for calculating size...

dbyron0:
Sure would love to get this in.  Any more I can do?

PR #277, opened 2019-01-28
support negative indeces with arrays

dbyron0:
so e.g.

{{ (split_me.split('-'))[-2] }}

works

It's probably dicey of me to use the native java split method, as opposed to the split filter implemented in jinjava.  I haven't yet figured out how much better the filter is, but it's pretty tempting to have at least some templates work in both jinjava and python without needing to add a split filter to python.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/el/ExtendedSyntaxBuilderTest.java
+++ b/src/test/java/com/hubspot/jinjava/el/ExtendedSyntaxBuilderTest.java
@@ -274,6 +274,8 @@ public class ExtendedSyntaxBuilderTest {
     assertThat(val("stringToSplit.split('-')")).isEqualTo(new String[]{ "one", "two", "three", "four", "five" });
 
     assertThat(val("stringToSplit.split('-')[-1]")).isEqualTo("five");
+    assertThat(val("stringToSplit.split('-')[1.5]")).isEqualTo("");
+    assertThat(val("stringToSplit.split('-')[-1.5]")).isEqualTo("");
 
     // out of range returns null, as -6 + the length of the array is still
     // negative, and java doesn't support negative array indices.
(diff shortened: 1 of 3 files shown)


src/test/java/com/hubspot/jinjava/el/ExtendedSyntaxBuilderTest.java
```diff
@@ -264,6 +264,25 @@ public void listRangeSyntaxNegativeIndeces() {
...
+  @Test
+  public void arrayWithNegativeIndeces() {
+    String stringToSplit = "one-two-three-four-five";
+    context.put("stringToSplit", stringToSplit);
+
+    // Negative index handling on lists happens elsewhere, so make sure we're
+    // dealing with an array of Strings.
+    assertThat(val("stringToSplit.split('-')")).isEqualTo(new String[]{ "one", "two", "three", "four", "five" });
+
+    assertThat(val("stringToSplit.split('-')[-1]")).isEqualTo("five");
```
dbyron0: This is the only test that failed before this change.

src/test/java/com/hubspot/jinjava/el/ExtendedSyntaxBuilderTest.java
```diff
@@ -264,6 +264,25 @@ public void listRangeSyntaxNegativeIndeces() {
     assertThat(val("mylist[-2:-3]")).isEqualTo(Lists.newArrayList());
   }
 
+  @Test
+  public void arrayWithNegativeIndeces() {
```
boulter: spelling nit: Indices

dbyron0: https://github.com/HubSpot/jinjava/pull/277/commits/cc94435229a21652d51a96b7ede2e0ef92e0be7b

dbyron0:
I made a tweak to keep from doing anything for float indices: https://github.com/HubSpot/jinjava/pull/277/commits/f23721c6cffb62cddf652742beaec2c4f3c70a3d.  Torn about whether to call intValue vs. longValue but it seems like intValue is enough elsewhere.

dbyron0:
Seems like there's a bunch of activity going on.  Is this ready to go?

dbyron0:
Thanks much.

PR #282, opened 2019-02-18
Allow partial evaluation of templates

pfarrel:
This PR implements partial evaluation of templates, or, deferring the evaluation of parts of a template, however you want to think about it.

The aim here is to support a use case for HubSpot's email infrastructure. We want to evaluate as much as possible of an email's template, except for the parts which are specific to an individual contact. This will allow us to minimise the work which needs to be repeated per-contact for large email sends to 100,000s of contacts. In this case the CPU time for evaluating the template isn't really a concern, but many Tags perform I/O, and this is the part we really don't want to repeat more than necessary.

How it works:
- I've defined a new "marker" object - `DeferredValue`.
- The user can insert these objects into their template context wherever they want. For example, in email's use case we would do `context.put("contact, DeferredValue.instance())`, because we want to defer the per-contact properties which will appear in the email.
- When the `ExpressionResolver` encounters one of these objects, for example when encountering the expression `{{contact.firstname}}` while rendering a template, it will throw a `DeferredValueException`. 
- `DeferredValueException` is caught in a few specific places. When caught the engine attempts to just echo the Node / Tag / Expression which caused the exception, leaving it untouched.
- A later run over the output can then fill in the parts which were not evaluated.

Concerns:
- `{% raw %}` wrapped content will be lost in the deferred evaluation scenario, because they will be evaluated twice. I could add a new mode to the tag which rewrapped its output in another raw wrapper?
- I'm pretty unsure about what I've done in `TagNode`. It seems to work, but will definitely need some more work.
- Yes this is using exceptions for flow control, but it works out really elegantly. The other alternative would be constantly checking the `resolvedExpressions` collection and throwing if we see something being added which we don't like. Maybe this would be better? I think the exception approach is more flexible, because we can do things like making functions throw `DeferredValueException` under certain circumstances, like if the function returned the current time and we are in deferred evaluation mode.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -17,7 +17,7 @@ package com.hubspot.jinjava.tree;
 
 import com.hubspot.jinjava.interpret.InterpretException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
-import com.hubspot.jinjava.interpret.DeferredValueEncounteredException;
+import com.hubspot.jinjava.interpret.DeferredValueException;
 import com.hubspot.jinjava.lib.tag.Tag;
 import com.hubspot.jinjava.tree.output.OutputNode;
 import com.hubspot.jinjava.tree.output.RenderedOutputNode;
@@ -55,7 +55,7 @@ public class TagNode extends Node {
 
     try {
       return tag.interpretOutput(this, interpreter);
-    } catch (DeferredValueEncounteredException e) {
+    } catch (DeferredValueException e) {
       return new RenderedOutputNode(reconstructImage());
     } catch (InterpretException e) {
       throw e;
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -15,12 +15,9 @@
  **********************************************************************/
 package com.hubspot.jinjava.tree;
 
-import java.util.stream.Collectors;
-
-import com.google.common.base.Joiner;
 import com.hubspot.jinjava.interpret.InterpretException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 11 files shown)


src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -84,4 +86,17 @@ public Tag getTag() {
...
 
+
+  private String reconstructImage() {
+    StringBuilder builder = new StringBuilder().append(master.getImage());
+
+    for (Node n : getChildren()) {
+      builder.append(n.getMaster().getImage());
+    }
+
+    builder.append("{% ").append(getEndName()). append(" %}");
```
boulter: can you use/create constants for `{%` and `%}`? People have requested to make these configurable and I'd like to consolidate them.

src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -84,4 +86,17 @@ public Tag getTag() {
     return tag;
   }
 
+
+  private String reconstructImage() {
+    StringBuilder builder = new StringBuilder().append(master.getImage());
```
boulter: I wonder if we could store a reference to the original template string and an index value of the start and end n the node object. I think this would solve the raw case and preserve all the whitespace.

src/main/java/com/hubspot/jinjava/tree/TagNode.java
pfarrel: Sure

src/main/java/com/hubspot/jinjava/tree/TagNode.java
pfarrel: I'll try to get that working in a separate PR. If I can get it to work, I'll loop back and update this PR to use it.

pfarrel: Actually, it looks like it might require a larger version bump because I will need to mess around with Tag / Node etc constructors. I'll go ahead with this for now and come back to that as I refine this.

PR #284, opened 2019-02-21
Add namespace (Global Functions) support for jinjava

pastorcmentarny:
Add namespace (Global Functions) support for jinjava.

Feel free to contact me if you have any questions.

I agree to your license for contribution under the terms of the APLv2.

Have an awesome day!


mattcoley:
I am a bit confused by this approach. From my understanding namespaces are not equivalent to global variables. They are essentially dicts whose values can be updated within for loops. In that case `namespace` should be a function that returns a `Namespace` object wrapper around a map.

pastorcmentarny:
Hello Matt!

Thank you for your quick feedback.
Oh.. I apologize, I misunderstood things a little bit.

I will change the code and commit changes ( I just need to figure out the right way to get Namespace through FunctionLibrary and get the correct way to get data from the namespace.

One more time thank you for your feedback.

PR #285, opened 2019-02-21
Use constants for token start/end symbols

pfarrel:
I forgot to include this change in my previous PR: https://github.com/HubSpot/jinjava/pull/282

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -94,9 +94,15 @@ public class TagNode extends Node {
       builder.append(n.getMaster().getImage());
     }
 
-    builder.append(TokenScannerSymbols.TOKEN_EXPR_START_CHAR).append(TokenScannerSymbols.TOKEN_TAG_CHAR).append(" ")
-        .append(getEndName())
-        .append(" ").append(TokenScannerSymbols.TOKEN_TAG_CHAR).append(TokenScannerSymbols.TOKEN_EXPR_END_CHAR);
+    String endTag = String.format(
+        "%s%s %s %s%s",
+        TokenScannerSymbols.TOKEN_EXPR_START_CHAR,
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -86,15 +87,16 @@ public Tag getTag() {
...
-
   private String reconstructImage() {
     StringBuilder builder = new StringBuilder().append(master.getImage());
 
     for (Node n : getChildren()) {
       builder.append(n.getMaster().getImage());
     }
 
-    builder.append("{% ").append(getEndName()). append(" %}");
+    builder.append(TokenScannerSymbols.TOKEN_EXPR_START).append(TokenScannerSymbols.TOKEN_TAG).append(" ")
```
mattcoley: Can we make this just one append with a `String.format` for readability? The StringBuilder makes sense for the above code however makes this bit hard to read.

pfarrel: Sure, done. I think it's a bit better now.

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
```diff
@@ -17,13 +17,22 @@
...
 
-  int TOKEN_PREFIX = '{';
-  int TOKEN_POSTFIX = '}';
-  int TOKEN_FIXED = 0;
-  int TOKEN_NOTE = '#';
-  int TOKEN_TAG = '%';
-  int TOKEN_EXPR_START = '{';
-  int TOKEN_EXPR_END = '}';
-  int TOKEN_NEWLINE = '\n';
+  char TOKEN_PREFIX_CHAR = '{';
```
jboulter: for convenience it might be nice to have constants for `{%`, `%}`, `{{` and `}}`

PR #289, opened 2019-02-27
WIP Error and documentation overhaul

mattcoley:
This is a work in progress attempt to make error messaging consistent and improve documentation. On the documentation side I have split the `input` into a filter/expression test from the arguments for that filter/test. When the input is invalid (for example passing a string into a numeric function), we throw a `InvalidInputException`. When an argument is invalid we throw an `InvalidArgumentException`. If the number of arguments do not match the required number of arguments for the function, we throw an `TemplateSyntaxException`. 

For invalid inputs and arguments, we specify one of the `InvalidReason` enum values and allow the exceptions to automatically handle building a consistent error message. For example, getting the absolute value of a string `'test'|abs` would result in `throw new InvalidInputException(interpreter, this, InvalidReason.NUMBER_FORMAT, object.toString());`. The exception will automatically craft a nice error message: 

`Invalid input in 'abs': input with value 'test' must be a number`. 

For an invalid argument like (`10|add('test')`), we also pass in the argument number: `throw new InvalidArgumentException(interpreter, this, InvalidReason.NUMBER_FORMAT, 0, arg[0]);` which will automatically craft a message:

`Invalid argument in 'add': 1st argument with value 'test' must be a number.`

### DIFF
--- a/src/main/java/com/hubspot/jinjava/doc/annotations/JinjavaDoc.java
+++ b/src/main/java/com/hubspot/jinjava/doc/annotations/JinjavaDoc.java
@@ -11,6 +11,8 @@ public @interface JinjavaDoc {
 
   String value() default "";
 
+  JinjavaParam input();
+
   JinjavaParam[] params() default {};
 
   JinjavaSnippet[] snippets() default {};
--- a/src/main/java/com/hubspot/jinjava/doc/annotations/JinjavaDoc.java
+++ b/src/main/java/com/hubspot/jinjava/doc/annotations/JinjavaDoc.java
@@ -11,7 +11,7 @@ public @interface JinjavaDoc {
 
   String value() default "";
 
-  JinjavaParam input();
+  JinjavaParam[] input() default {};
 
   JinjavaParam[] params() default {};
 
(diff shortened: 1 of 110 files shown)


bgoodies:
🆒 👍 

after this PR, can we deprecate constructors for some of the other exceptions and template error that do not use line number and start position?

src/main/java/com/hubspot/jinjava/interpret/InvalidArgumentException.java
```diff
@@ -0,0 +1,63 @@
...
+  public int getStartPosition() {
+    return startPosition;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  private static String formatArgumentNumber(int argumentNumber) {
+    switch (argumentNumber){
```
jboulter: this is not very i18nable, but in any case, you can do this for all numbers by just looking at the last digit of the int and using "st" if 1, "nd" if 2, "rd" if 3 and "th" for all other values.

src/main/java/com/hubspot/jinjava/doc/annotations/JinjavaDoc.java
```diff
@@ -11,6 +11,8 @@
 
   String value() default "";
 
+  JinjavaParam input();
```
jboulter: while you're in here, can you add a `required` attribute to `JinjavaParam`? @williamspiro wanted one.

mattcoley: Labeled all filter/exp tests with this new boolean value.

PR #296, opened 2019-03-05
Update expression resolver to return null instead of blank string.

mattcoley:
It doesn't make sense to return an empty string from the expression resolver if the expression cannot be evaluated due to an exception. Some built-in tags will treat `""` differently from null. For example `for` will treat `""` as `[""]` whereas it should be treated as `[]`.


mattcoley:
I could not find a reason why we return empty string here. We used to return `false` and then that was updated to empty string. The tags that use the expression resolver handle the null value better than empty string.

mattcoley:
Here: https://github.com/HubSpot/jinjava/commit/1d694cd8a5faaf904c0791a17e9b011e0247f121. The expression resolver is only used in tags: `do, for, if, macro, etc.` and in the `ExpressionNode` https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java#L42. The expression node handles the null value fine by defaulting to empty string.

Issue #292, opened 2019-03-04
Need More Math (Example: square root)


jfudman:
It'd be huge to have a math function where we could directly use [python math functions here](https://docs.python.org/3/library/math.html).
**EDIT** In this case I guess it'd be [java math class](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html)

## Solution 1:
[Here is a link](https://www.programiz.com/python-programming/modules/math) that includes a table summary of those functions above.

In hubl, it would possibly look like this:
```
{% set num = 100 %}
{{ math.sqrt( num ) }}  {# outputs 10 #}
```
## Solution 2:
You could just use filters for more common math functions like square root `sqrt`
```
{% set num = 100 %}
{{ num|sqrt }}  {# outputs 10 #}
```

PR #299, opened 2019-03-08
Add `root` and `log` filters.

mattcoley:
Adds a `root` and a `log` filter as requested here https://github.com/HubSpot/jinjava/issues/292 similar to https://ansible-docs.readthedocs.io/zh/stable-2.0/rst/playbooks_filters.html#math.

Complications: supporting BigDecimal and exact values requires adding a dependency on https://github.com/eobermuhlner/big-math

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
@@ -17,8 +17,8 @@ import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import ch.obermuhlner.math.big.BigDecimalMath;
 
 @JinjavaDoc(
-    value = "Return the log of the input.",
-    input = @JinjavaParam(value = "number", type = "number", desc = "The number that you want to get the log of", required = true),
+    value = "Return the natural log of the input.",
+    input = @JinjavaParam(value = "number", type = "number", desc = "The number to get the log of", required = true),
     params = @JinjavaParam(value = "base", type = "number", defaultValue = "e (natural logarithm)", desc = "The base to use for the log calculation"),
     snippets = {
         @JinjavaSnippet(
--- a/src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
@@ -3,92 +3,103 @@ package com.hubspot.jinjava.lib.filter;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.math.MathContext;
+import java.math.RoundingMode;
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
```diff
@@ -0,0 +1,120 @@
...
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.InvalidArgumentException;
+import com.hubspot.jinjava.interpret.InvalidInputException;
+import com.hubspot.jinjava.interpret.InvalidReason;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+
+import ch.obermuhlner.math.big.BigDecimalMath;
+
+@JinjavaDoc(
+    value = "Return the log of the input.",
```
jboulter: "return the natural log value of the input"

src/test/java/com/hubspot/jinjava/lib/filter/RootFilterTest.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Test
+  public void itCalculatesNthRoot() {
+    assertThat(filter.filter(9765625d, interpreter, "5.0")).isEqualTo(25d);
+  }
+
+  @Test
+  public void itCalculatesNthRootOfBigDecimal() {
+    BigDecimal result = ((BigDecimal) filter.filter(new BigDecimal(9765625d), interpreter, "5.0"));
```
jboulter: can this be computationally expensive with large values?

src/main/java/com/hubspot/jinjava/lib/filter/LogFilter.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    if (base != null && base <= 0) {
+      throw new InvalidArgumentException(interpreter, this, InvalidReason.POSITIVE_NUMBER, 0, base);
+    }
+  }
+
+  @Override
+  public String getName() {
+    return "log";
```
jboulter: I'm not thrilled about the very generic name of this, but it makes sense to be compatible with ansible.

src/test/java/com/hubspot/jinjava/lib/filter/RootFilterTest.java
mattcoley: Performance is largely a function of precision in this case set to 50 digits. With such a low precision operations are in the nanoseconds range. The goal of this library we are using is for performance remain stable as the input grows so we should be safe http://obermuhlner.ch/wordpress/2016/06/02/bigdecimalmath/

PR #306, opened 2019-03-12
Add namespace functionality

pastorcmentarny:


pastorcmentarny:
I am glad to hear that it will be added :D.

Issue #91, opened 2016-11-21
Macro recursion broken in versions 2.1.8 and later


derkork:
Consider the following code:

```
{% macro example(prefix, items) %}
{{ prefix }} {{ items | first }}
{% if (items | count > 1) %}{{ example( prefix ~'-', items[1:items|count]) }}{% endif %}
{% endmacro %}

{{  example('-',['lorem', 'ipsum', 'dolor']) }}
```

With jinjava 2.1.7 or earlier this correctly outputs:

```
- lorem

-- ipsum

--- dolor
```

Starting with version 2.1.8 this no longer works and instead just outputs:

```
- lorem
```

boulter:
@derkork, this was an intentional change in 2.1.8. See the release notes in https://github.com/HubSpot/jinjava/blob/master/CHANGES.md

derkork:
@boulter, I understand. Could you give me some insights on why this has been removed and what could be used as a replacement functionality? Also is it possible to call macros from other macros or has this functionality also been removed?

boulter:
@derkork you can call other macros from a macro, but you cannot have any sort of cycle, direct or indirect. 

We added this as a safety because we noticed our customers would accidentally create cycles.

A possible patch you could implement could make this behavior configurable per Jinjava instance.

derkork:
Yeah something like a "--i-know-what-i-am-doing" flag would be nice here, e.g.:

```
JinjavaConfig.newBuilder().withMacroRecursion(true)..
```

PR #308, opened 2019-03-12
Allow ability to set a max recursion depth in config.

mattcoley:
This is a middle ground option for https://github.com/HubSpot/jinjava/issues/91. You can now set `enableRecursiveMacroCalls` to `true` and `maxMacroRecursionDepth` to a reasonable value to prevent a template from killing your thread due to stack overflows.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -41,6 +41,7 @@ public class JinjavaConfig {
 
   private final boolean readOnlyResolver;
   private final boolean enableRecursiveMacroCalls;
+  private final int maxMacroRecursionDepth;
 
   private Map<Context.Library, Set<String>> disabled;
   private final boolean failOnUnknownTokens;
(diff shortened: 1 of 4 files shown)


mattcoley:
The need for this is to be able to build nesting structures like site menus without resolving to copying and pasting a macro n times to support a menu with depth n.

src/main/java/com/hubspot/jinjava/el/ext/AstMacroFunction.java
```diff
@@ -34,15 +34,27 @@ public Object eval(Bindings bindings, ELContext context) {
...
+            } else {
+              macroStack.pushWithoutCycleCheck(getName());
+            }
           } else {
             macroStack.push(getName(), -1, -1);
           }
         } catch (MacroTagCycleException e) {
+
+          int maxDepth = interpreter.getConfig().getMaxMacroRecursionDepth();
+          String message;
```
boulter: nit: tertiary statement

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -54,11 +55,11 @@ public static Builder newBuilder() {
   }
 
   public JinjavaConfig() {
-    this(StandardCharsets.UTF_8, Locale.ENGLISH, ZoneOffset.UTC, 10, new HashMap<>(), false, false, true, false, false, 0, true, RandomNumberGeneratorStrategy.THREAD_LOCAL, false, 0);
+    this(StandardCharsets.UTF_8, Locale.ENGLISH, ZoneOffset.UTC, 10, new HashMap<>(), false, false, true, false, 0, false, 0, true, RandomNumberGeneratorStrategy.THREAD_LOCAL, false, 0);
   }
 
   public JinjavaConfig(Charset charset, Locale locale, ZoneId timeZone, int maxRenderDepth) {
-    this(charset, locale, timeZone, maxRenderDepth, new HashMap<>(), false, false, true, false, false, 0, true, RandomNumberGeneratorStrategy.THREAD_LOCAL, false, 0);
+    this(charset, locale, timeZone, maxRenderDepth, new HashMap<>(), false, false, true, false, 0, false, 0, true, RandomNumberGeneratorStrategy.THREAD_LOCAL, false, 0);
```
boulter: might be nigh time for a builder here.

mattcoley:
Java 8 stack frame size in the worst case is going to be about 1kb so I was thinking 50 as a limit.

boulter:
I think 10 or 20 would be more reasonable since you could possibly interleave other functions recursively.

mattcoley:
I think 20 would be good. This is total macro stack size but I think it is unlikely that someone currently has a template that calls 20 nested macros.

TheWebTech:
> I think 20 would be good. This is total macro stack size but I think it is unlikely that someone currently has a template that calls 20 nested macros.

especially on HubSpot, where nesting macros previously didn't actually work.

PR #315, opened 2019-03-25
Return long value from int filter if over max int length.

mattcoley:
Int is unbounded in python so trying to come up with a solution to bring Jinjava `|int` closer to parity with Jinja. Here we are returning a long value instead of a int if the value is above the maximum allowed for ints. This should fix most issues where people are trying to convert long IDs into numbers using the `|int` filter. Some thought is needed to figure out a better long term solution for higher values than what long supports.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/IntFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/IntFilter.java
@@ -37,33 +37,40 @@ public class IntFilter implements Filter {
   public Object filter(Object var, JinjavaInterpreter interpreter,
       String... args) {
 
-    int defaultVal = 0;
+    Long defaultVal = 0L;
     if (args.length > 0) {
-      defaultVal = NumberUtils.toInt(args[0], 0);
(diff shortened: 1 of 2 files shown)


boulter:
would there be harm in always returning a `long` from the `int` filter if in python `|int` means any whole number?

mattcoley:
Not sure what the implications are for always returning `long` so I am trying to be safe by only returning long if the value overflows.

mattcoley:
Filters/functions that don't do type conversions (i.e. advanced filters) may be affected by always returning a long.

boulter:
> Filters/functions that don't do type conversions (i.e. advanced filters) may be affected by always returning a long.

I'd say it would be more likely to create problems if the returned type varies by input. Users won't expect they'll sometimes get longs.

mattcoley:
I am going to fix type conversion in a few other places then make this always return long.

mattcoley:
The only issue I can think of is if you explicitly check for the type: `type(num) == 'int'`

PR #316, opened 2019-03-25
Add support for resolving relative paths in separate files

liamrharwood:
This PR is to support resolving relative paths located in a separate file for cases such as HubSpot custom modules.
- It adds a `peek()` method to `CallStack` to allow looking at the top of the include/extend stack in order to determine what file a tag is located in.
- It adds `importResourcePath` to the local context of each macro in `ImportTag` so that the macro will know which file it originated from.

Let me know if additional information is needed.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
@@ -91,27 +91,31 @@ public class ImportTag implements Tag {
       String template = interpreter.getResource(templateFile);
       Node node = interpreter.parse(template);
 
+      JinjavaInterpreter child = new JinjavaInterpreter(interpreter);
+      child.getContext().put("import_resource_path", templateFile);
+      JinjavaInterpreter.pushCurrent(child);
+
+      try {
+        child.render(node);
+      } finally {
+        JinjavaInterpreter.popCurrent();
+      }
+
+      interpreter.addAllErrors(child.getErrorsCopy());
+
+      Map<String, Object> childBindings = child.getContext().getSessionBindings();
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
```diff
@@ -95,6 +95,7 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
         interpreter.render(node);
       } else {
         JinjavaInterpreter child = new JinjavaInterpreter(interpreter);
+        child.getContext().put("importResourcePath", templateFile);
```
mattcoley: Since jinjava was directly ported from python we generally use snakecase for context variables.

liamrharwood:
- This now uses snake case for `import_resource_path`
- I added the `import_resource_path` to context in `FromTag` as well
- I changed `ImportTag` to use a child interpreter for the `import` case instead of just `import...as`. This is necessary so that the macros imported without `as` can still use `import_resource_path` to know what file they came from.

liamrharwood:
Alright, I changed this around a bit to better support all types of includes, imports, and extends.
- I added a `currentPathStack` to `Context` which keeps track of the current file no matter what kind of tag is used.
- I still pass the `import_resource_path` into imported macros but now I push it onto the `currentPathStack` when evaluating the macro. I do this without checking for cycles in order to allow for macros referencing other macros in the same file.
- I added a test to ensure that `import_resource_path` is included in the local context scope for macros.
- I push the `extends` path from the top of the `extendsPathStack` onto the `currentPathStack` during the `processExtendParents` step of `JinjavaInterpreter::render`.

PR #318, opened 2019-03-28
Refactor tags to use static strings for tag names.

mattcoley:
This makes it a lot easier to build tag exclusion lists for interpreters.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/AutoEscapeTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/AutoEscapeTag.java
@@ -23,18 +23,12 @@ import com.hubspot.jinjava.util.LengthLimitingStringBuilder;
 public class AutoEscapeTag implements Tag {
 
   public static final String TAG_NAME = "autoescape";
-  public static final String END_TAG_NAME = "endautoescape";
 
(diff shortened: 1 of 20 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/AutoEscapeTag.java
```diff
@@ -21,16 +21,20 @@
                 "{% endautoescape %}")
     })
 public class AutoEscapeTag implements Tag {
+
+  public static final String TAG_NAME = "autoescape";
+  public static final String END_TAG_NAME = "endautoescape";
```
boulter: `END_TAG_NAME` can always be "end" + `TAG_NAME`;

mattcoley: I was trying to come up with the best way to do this. Should we just default to the endtag being "end" + TAG_NAME and hope that people override the method for the null case?

mattcoley: Ok now we have `default String getEndTagName()` on `Tag.class` since I only ready need to use the start tag for building exclusion lists.

PR #327, opened 2019-04-30
defer variables containing deferred vals & track deferred nodes in context

jkollmann:
Following https://github.com/HubSpot/jinjava/pull/282

1. **Make deferred infectious for variables**
It's possible that variables contain deferred values, in that case the assigned variable(s) should be deferred too, since expressions depending on variables containing deferred values can't and shouldn't be evaluated & rendered.

Example:

Input: 

```
<html>
{% set some_var = "email:" + contact.email %} {% if contact.email  %}{{ contact.email }}{% endif %}
<head></head>
<body></body>
</html>
```

Render before the change:

```
<html>
{% set some_var = "email:" + contact.email %}
<head></head>
<body></body>
</html>
```

After the change:

```
<html>
{% set some_var = "email:" + contact.email %}
{% if contact.email  %}
{{ contact.email }}
{% endif %}
<head></head>
<body></body>
</html>
```

2. **Add list to the context to track deferred nodes**
We need some way of tracking which nodes were deferred so that we can escape them before post-processing the rendered template with Jsoup. If Jsoup encounters these kinds of tags in a place where no characters are allowed it changes the template and even ignores following tags inside of that tag.

Example: 

Input:

```
<html>
{% set some_var = "email:" + contact.email %}
{% if contact.email  %}
{{ contact.email }}
{% endif %}
<head>
<meta name="description" content="some content">
</head>
<body></body>
</html>
```

Jsoup output:
```
<html>
 <head></head>
 <body>
{% set some_var = "email:" + contact.email %}
{% if contact.email %}
{{ contact.email }}
{% endif %}
  <meta name="description" content="some content">
 </body>
</html>
```

In order to escape these tags properly we need to know which tags were deferred. 

We also want to handle the case where a deferred tag contains not-deferred variables. We need to be able to take these out of the context later for further processing.

Example:

Input:

```
<html>
{% set defined_var = "defined" %}
{% set some_var = "email:" + contact.email + defined_var  %} 
{% if contact.email  %}
{{ contact.email }}
{% endif %}
<head>
<meta name="description" content="some content">
</head>
<body></body>
</html>
```

Output:

```
<html>
{% set some_var = "email:" + contact.email + defined_var  %} 
{% if contact.email  %}
{{ contact.email }}
{% endif %}
<head>
<meta name="description" content="some content">
</head>
<body></body>
</html>
```

`defined_var` disappears but `some_var` will not be rendered since it contains a deferred value. If we know which tags have been deferred after the render we can take the vars out of the context and pass them along for a second render.

**Considerations:** I've tried a couple of different approaches, including partial evaluation and extracting variables of deferred tags. Both are a lot more complex than I initially thought and too specific to our use-case, especially given the fact that we don't own `ExpressionFactoryImpl` and all the code that this class depends on. IMO this is the most flexible solution, since users can decide what to do with the deferred nodes after rendering is done.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -144,6 +144,7 @@ public class Context extends ScopeMap<String, Object> {
     resolvedValues.clear();
     resolvedFunctions.clear();
     dependencies = HashMultimap.create();
+    deferredNodes.clear();
   }
 
   @Override
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -28,7 +28,6 @@ import java.util.Stack;
 import java.util.stream.Collectors;
 
 import com.google.common.collect.HashMultimap;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.SetMultimap;
 import com.hubspot.jinjava.lib.Importable;
@@ -78,7 +77,7 @@ public class Context extends ScopeMap<String, Object> {
   private final Set<String> resolvedValues = new HashSet<>();
   private final Set<String> resolvedFunctions = new HashSet<>();
 
-  private List<Node> deferredNodes = new ArrayList<>();
+  private Set<Node> deferredNodes = new HashSet<>();
 
   private final ExpTestLibrary expTestLibrary;
   private final FilterLibrary filterLibrary;
@@ -144,7 +143,7 @@ public class Context extends ScopeMap<String, Object> {
     resolvedValues.clear();
     resolvedFunctions.clear();
     dependencies = HashMultimap.create();
-    deferredNodes = new ArrayList<>();
+    deferredNodes = new HashSet<>();
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -238,6 +241,17 @@ public void addResolvedFunction(String function) {
     }
   }
 
+  public void addDeferredNode(Node node) {
+    deferredNodes.add(node);
```
boulter: make sure you also clear these in `clear`

src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
```diff
@@ -81,23 +84,28 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
...
 
-    } else {
-      // handle single variable assignment
-      interpreter.getContext().put(var, interpreter.resolveELExpression(expr, tagNode.getLineNumber()));
+      } else {
+        // handle single variable assignment
+        interpreter.getContext().put(var, interpreter.resolveELExpression(expr, tagNode.getLineNumber()));
+      }
+    } catch (DeferredValueException e) {
+      Stream.of(varTokens).forEach(varToken -> interpreter.getContext().put(varToken.trim(), DeferredValue.instance()));
```
pfarrel: This is the key bit that makes "deferred-ness" infectious, for any other readers trying to see what actually changed through the whitespace diff.

PR #328, opened 2019-05-01
Create a new one instead of clear.

hs-lsong:
Since the JinJava is shared globally, so are the global context and its dependencies. Let's create a new one instead clear it.

```
ConcurrentModificationException
 
java.util.HashMap$HashIterator in nextNode at line 1445
java.util.HashMap$ValueIterator in next at line 1474
com.google.common.collect.AbstractMapBasedMultimap in clear at line 273
com.google.common.collect.HashMultimap in clear at line 46
com.hubspot.jinjava.interpret.Context in reset at line 143
```


boulter:
@jkollmann I think you may need to do the same in https://github.com/HubSpot/jinjava/commit/84be45ee28638989349ac03e02ca9b44e1a6557d

PR #330, opened 2019-05-15
Support nominative date formats.

mattcoley:
Some languages have two version of months (nominative and genitive). By default, JDK8 uses genitive format which can cause issues. This allows you to use `%OB` instead of `%B` to get the nominative version of the month name.


anev:
LGTM

PR #331, opened 2019-05-16
Support empty bracket implicit index syntax.

mattcoley:
It was not possible previously do something like `list[3:]` to get the sublist of all elements from index 3 to the end, as well as `list[:3]`, the sublist of all elements from index 0 up to index 3.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
@@ -41,14 +41,9 @@ public class AstRangeBracket extends AstBracket {
       return evalString((String) base, bindings, context);
     }
 
-    Iterable<?> baseItr;
-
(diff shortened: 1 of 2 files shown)


bgoodies:
👍

src/main/java/com/hubspot/jinjava/el/ext/AstRangeBracket.java
```diff
@@ -41,15 +41,24 @@ public Object eval(Bindings bindings, ELContext context) {
       return evalString((String) base, bindings, context);
     }
 
-    Object start = property.eval(bindings, context);
+    Iterable<?> baseItr;
```
boulter: you could use a ternary here

mattcoley: I need to have a better eye out for refactoring code that I come across.

PR #334, opened 2019-05-30
Upgrade to newer basepom

jhaber:
This upgrades to new basepom and also builds on Java 11

@jboulter @boulter


src/test/java/com/hubspot/jinjava/objects/date/StrftimeFormatterTest.java
```diff
@@ -94,6 +94,6 @@ public void itConvertsNominativeFormats() {
     ZonedDateTime zonedDateTime = ZonedDateTime.parse("2019-06-06T14:22:00.000+00:00");
 
     assertThat(StrftimeFormatter.formatter("%OB").withLocale(Locale.forLanguageTag("ru")).format(zonedDateTime))
-        .isEqualTo("Июнь");
+        .isIn("Июнь", "июнь");
```
jhaber: Seems like there was a minor change with Java 11; I made the test accept either format so it still works on Java 8

mattcoley: `июнь` is correct, it was a bug with Java 8.

boulter:
Will it still target Java 8 bytecode?

jhaber:
Yup

Issue #291, opened 2019-03-03
ipaddr - Support more filters


MosheElisha:
Hi,

Currently, ipaddr supports only the "prefix" filter.
Requesting to extend support of other useful filters such as "netmask", "address" and "broadcast".

Thanks.

MikaelMollberg:
I agree. Additional filters would be great to have.

PR #338, opened 2019-06-06
Support more ipaddr filters

mattcoley:
For https://github.com/HubSpot/jinjava/issues/291
Adds support for `broadcast`, `netmask`, and `address`.


mattcoley:
Unfortunately I have to add apache `commons-net` dependency for the calculations.

PR #343, opened 2019-06-20
Add initial support for resolving relative paths

liamrharwood:
This PR adds support for resolving resource locations, such when using relative paths in a file system. 
- This adds a new `LocationResolver` interface, which resolves a string to the absolute location of a resource (i.e. resolves a relative path to an absolute path).
- Each `ResourceLocator` can have an optional `LocationResolver`. I decided to have this live in the `ResourceLocator` since the behavior should be fairly linked.
- This also adds the `resolveResourceLocation()` method to `JinjavaInterpreter` which uses the `LocationResolver` of the current `ResourceLocator` to resolve the location of a resource.
- I modified the `include`, `import`, `extends`, and `from [..] import` tags to resolve the location before retrieving a resource _and_ before adding it to the `currentPathStack` (this is to make sure all paths in the stack are absolute).
- I added a basic `RelativePathResolver` which is a `LocationResolver` that resolves relative paths beginning with `./` based on the `currentPathStack` and the `current_path` context value. I also added tests for each of the tags mentioned above to test this behavior using the `RelativePathResolver`. (I wasn't sure if this class should live in Jinjava or not, but felt it was a good basic example of a `LocationResolver`.)


src/main/java/com/hubspot/jinjava/loader/RelativePathResolver.java
```diff
@@ -0,0 +1,26 @@
...
+
+  @Override
+  public String resolve(String path, JinjavaInterpreter interpreter) {
+    if (path.startsWith("./")) {
+      String parentPath = interpreter.getContext().getCurrentPathStack().peek()
+          .orElseGet(() -> (String) interpreter.getContext().getOrDefault(CURRENT_PATH_CONTEXT_KEY, ""));
+
+      Path templatePath = Paths.get(parentPath);
+      Path folderPath = templatePath.getParent() != null ? templatePath.getParent() : Paths.get("");
+      if (folderPath != null) {
```
liamrharwood: I don't think that `folderPath` would ever be null, but the build was failing because `Travis CI` was saying there was a possible null dereference. 🤷‍♂

src/main/java/com/hubspot/jinjava/loader/RelativePathResolver.java
```diff
@@ -0,0 +1,26 @@
...
+        return folderPath.resolve(path).normalize().toString();
+      }
+    }
+    return path;
+  }
+
```
mattcoley: nit: extra line

src/main/java/com/hubspot/jinjava/loader/ResourceLocator.java
```diff
@@ -17,11 +17,16 @@
...
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 
 public interface ResourceLocator {
 
   String getString(String fullName, Charset encoding, JinjavaInterpreter interpreter) throws IOException;
 
+  default Optional<LocationResolver> getLocationResolver() {
+    return Optional.empty();
+  }
+
```
mattcoley: Same here 🙃

liamrharwood:
That should only be possible using `FileLocator`, which incidentally already had [its own relative path resolution logic](https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/loader/FileLocator.java#L44-L52). We could update the `FileLocator` to use a `LocationResolver` which explicitly blocks such an exploit.

PR #345, opened 2019-06-25
Resolves Failure on Unknown Incompatible with default filter

### LINKED ISSUE #256: Fail on Unknown Incompatible with default filter
Not sure what the desired behaviour should be here but feels wrong to me.

{{ foo | default('bar') }}

When Fail on unknown tokens is set to true and foo doesnt exist in the context this fails with a syntax exception instead of resolving the default filter.

gabru-md:
Fixes #256 
Resolves failure on unknown incompatible with default filter.

Checks if the produced result is null and FailOnUnknownTokens in set to true, then only throw the UnknownToken Exception.

had to change the method signature though 😅

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -72,7 +72,7 @@ public class ExpressionResolver {
       String elExpression = EXPRESSION_START_TOKEN + expression.trim() + EXPRESSION_END_TOKEN;
       ValueExpression valueExp = expressionFactory.createValueExpression(elContext, elExpression, Object.class);
       Object result = valueExp.getValue(elContext);
-      if(result == null && interpreter.getConfig().isFailOnUnknownTokens()) {
+      if (result == null && interpreter.getConfig().isFailOnUnknownTokens()) {
         throw new UnknownTokenException(expression);
       }
       validateResult(result);
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -1,23 +1,12 @@
 package com.hubspot.jinjava.el;
 
-import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
-
-import java.util.List;
-
-import javax.el.ELException;
-import javax.el.ExpressionFactory;
-import javax.el.PropertyNotFoundException;
-import javax.el.ValueExpression;
-
-import org.apache.commons.lang3.StringUtils;
-
 import com.google.common.collect.ImmutableMap;
 import com.hubspot.jinjava.el.ext.NamedParameter;
 import com.hubspot.jinjava.interpret.DeferredValueException;
+import com.hubspot.jinjava.interpret.InvalidInputException;
+import com.hubspot.jinjava.interpret.InvalidArgumentException;
 import com.hubspot.jinjava.interpret.DisabledException;
 import com.hubspot.jinjava.interpret.InterpretException;
-import com.hubspot.jinjava.interpret.InvalidArgumentException;
-import com.hubspot.jinjava.interpret.InvalidInputException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.TemplateError;
 import com.hubspot.jinjava.interpret.TemplateError.ErrorItem;
@@ -27,8 +16,16 @@ import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.interpret.UnknownTokenException;
 import com.hubspot.jinjava.interpret.errorcategory.BasicTemplateErrorCategory;
 import com.hubspot.jinjava.lib.fn.ELFunctionDefinition;
-
 import de.odysseus.el.tree.TreeBuilderException;
+import org.apache.commons.lang3.StringUtils;
+
+import javax.el.ELException;
+import javax.el.ExpressionFactory;
+import javax.el.PropertyNotFoundException;
+import javax.el.ValueExpression;
+import java.util.List;
+
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
 
 /**
  * Resolves Jinja expressions.
@@ -72,7 +69,9 @@ public class ExpressionResolver {
       String elExpression = EXPRESSION_START_TOKEN + expression.trim() + EXPRESSION_END_TOKEN;
       ValueExpression valueExp = expressionFactory.createValueExpression(elContext, elExpression, Object.class);
       Object result = valueExp.getValue(elContext);
-
+      if(result == null && interpreter.getConfig().isFailOnUnknownTokens()) {
+        throw new UnknownTokenException(expression);
+      }
       validateResult(result);
 
       return result;
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -1,12 +1,23 @@
 package com.hubspot.jinjava.el;
 
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCauseMessage;
+
+import java.util.List;
+
+import javax.el.ELException;
+import javax.el.ExpressionFactory;
+import javax.el.PropertyNotFoundException;
+import javax.el.ValueExpression;
+
+import org.apache.commons.lang3.StringUtils;
+
 import com.google.common.collect.ImmutableMap;
 import com.hubspot.jinjava.el.ext.NamedParameter;
 import com.hubspot.jinjava.interpret.DeferredValueException;
-import com.hubspot.jinjava.interpret.InvalidInputException;
-import com.hubspot.jinjava.interpret.InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
```diff
@@ -27,8 +16,16 @@
 import com.hubspot.jinjava.interpret.UnknownTokenException;
 import com.hubspot.jinjava.interpret.errorcategory.BasicTemplateErrorCategory;
 import com.hubspot.jinjava.lib.fn.ELFunctionDefinition;
-
 import de.odysseus.el.tree.TreeBuilderException;
+import org.apache.commons.lang3.StringUtils;
```
boulter: Please revert the import order changes.

src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
```diff
@@ -72,7 +69,9 @@ public Object resolveExpression(String expression) {
       String elExpression = EXPRESSION_START_TOKEN + expression.trim() + EXPRESSION_END_TOKEN;
       ValueExpression valueExp = expressionFactory.createValueExpression(elContext, elExpression, Object.class);
       Object result = valueExp.getValue(elContext);
-
+      if(result == null && interpreter.getConfig().isFailOnUnknownTokens()) {
```
boulter: space after `if` please

src/main/java/com/hubspot/jinjava/interpret/UnknownTokenException.java
```diff
@@ -5,8 +5,8 @@
   private static final long serialVersionUID = -388757722051666198L;
   private final String token;
 
-  public UnknownTokenException(String token, int lineNumber, int startPosition) {
-    super("Unknown token found: " + token, lineNumber, startPosition);
+  public UnknownTokenException(String token) {
```
boulter: we definitely want to keep these line numbers and positions in the error message.

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
```diff
@@ -40,4 +41,21 @@ public void itReplaceTokensWithoutException() {
...
     assertThat(renderedTemplate).isEqualTo("hello test and test1");
   }
+
+  @Test
+  public void itReplacesTokensInContextButThrowsExceptionForOthers() {
+    Map<String, String> context = new HashMap<>();
+    context.put("animal", "lamb");
+    context.put("fruit", "apple");
+    String template = "{{ name }} has a {{ animal }}";
+    try{
```
boulter: add space after `try`

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
```diff
@@ -40,4 +41,21 @@ public void itReplaceTokensWithoutException() {
...
+      jinjava.render(template, context);
+    } catch (Exception ex) {
+      assertThat(ex.getMessage()).containsIgnoringCase("unknown token found");
```
boulter: this doesn't guarantee that the exception is thrown. You can use `assertThatThrownBy` for that.

src/main/java/com/hubspot/jinjava/interpret/UnknownTokenException.java
gabru-md: since the exception is being thrown at a completely new place now, there seems no way to know the line number and the starting position of the token, hence I changes the method signature.

Let me see if I can make some changes to alter this back to exactly how it was :+1:

gabru-md: Screenshot 2019-06-26 at 11 50 02 AM

currently, it doesn't show any lineNumber or startingPosition as well. So I guess removing it is okay?
The above screenshot was taken after I created a test and ran it on the current code in the master branch and it seems not to show the lineNumber and the startingPosition of the variable that is not found.

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
gabru-md: okay, will make the changes

src/main/java/com/hubspot/jinjava/interpret/UnknownTokenException.java
boulter: It's not in the `toString()` method, but we use it to build errors when we catch these exceptions.

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
```diff
@@ -40,4 +35,25 @@ public void itReplaceTokensWithoutException() {
     String renderedTemplate = jinjava.render(template, context);
     assertThat(renderedTemplate).isEqualTo("hello test and test1");
   }
+
+  @Test
+  public void itReplacesTokensInContextButThrowsExceptionForOthers() {
```
boulter: could you break this up into two separate tests?

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
```diff
@@ -40,4 +35,25 @@ public void itReplaceTokensWithoutException() {
...
+    context.put("fruit", "apple");
+    String template = "{{ name }} has a {{ animal }}";
+
+    Node node = new TreeParser(jinjavaInterpreter, template).buildTree();
+
+    assertThatThrownBy(() -> jinjavaInterpreter.render(node))
+            .isInstanceOf(UnknownTokenException.class)
+            .hasMessageContaining("Unknown token found: name");
+
+
```
boulter: nit: extra blank lines here

src/test/java/com/hubspot/jinjava/tree/FailOnUnknownTokensTest.java
gabru-md: on it :+1:

src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
gabru-md: the problem I'm facing is that when here I do not get the lineNumber and startPosition when I'm about to throw an error here. I only have the expression string with me.

gabru-md:
```java
String elExpression = EXPRESSION_START_TOKEN + expression.trim() + EXPRESSION_END_TOKEN;
      ValueExpression valueExp = expressionFactory.createValueExpression(elContext, elExpression, Object.class);
      Object result = valueExp.getValue(elContext);
      if (result == null && interpreter.getConfig().isFailOnUnknownTokens()) {
        throw new UnknownTokenException(expression);
      }
```

gabru-md: should I, in this case, pass -1 and -1 for the same?

boulter: you have access to the interpreter, so you can use `interpreter.getLineNumber(), interpreter.getPosition()`

gabru-md:
done @boulter

src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
gabru-md: done, thanks for the help 😄

src/main/java/com/hubspot/jinjava/interpret/UnknownTokenException.java
```diff
@@ -5,8 +5,8 @@
   private static final long serialVersionUID = -388757722051666198L;
   private final String token;
 
-  public UnknownTokenException(String token, int lineNumber, int startPosition) {
-    super("Unknown token found: " + token, lineNumber, startPosition);
+  public UnknownTokenException(String token, int lineNumber, int startingPosition) {
```
boulter: can you remove the changes here?

gabru-md: to the method signature?
okay

src/main/java/com/hubspot/jinjava/interpret/UnknownTokenException.java
```diff
@@ -6,7 +6,7 @@
   private final String token;
 
   public UnknownTokenException(String token, int lineNumber, int startPosition) {
-    super("Unknown token found: " + token, lineNumber, startPosition);
+    super("Unknown token found: " + token.trim(), lineNumber, startPosition);
```
boulter: do we still need the `trim()`

gabru-md: yes. since there was a test that ended up failing because of incorrect padding.
the only way was to trim the token to get rid of the extra space and make the tests pass

gabru-md: [this test](https://github.com/HubSpot/jinjava/pull/345/files#diff-b386f56a1fcb39f707f026174037eab7R152) was failing since the value was coming out to be
`"Unknown token found:  bad " `, where as the expected value is `"Unknown token found: bad"`

therefore I trimmed the token

boulter: You can use `hasMessageContaining` instead.

PR #347, opened 2019-06-26
Change initialization of JinjavaInterpreter to instantiation

Roobun:
Now it's possible to override instantiate method
It'll be helpful if you use extended interpreter

### DIFF
--- a/src/main/java/com/hubspot/jinjava/Jinjava.java
+++ b/src/main/java/com/hubspot/jinjava/Jinjava.java
@@ -194,7 +194,7 @@ public class Jinjava {
       renderConfig = parentInterpreter.getConfig();
     }
 
-    JinjavaInterpreter interpreter = new JinjavaInterpreter(this, context, renderConfig);
+    JinjavaInterpreter interpreter = globalConfig.getInterpreterFactory().newInstance(this, context, renderConfig);
     JinjavaInterpreter.pushCurrent(interpreter);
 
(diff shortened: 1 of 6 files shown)


mattcoley:
I believe you will also have to look at https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/Jinjava.java which also creates interpreters.

mattcoley:
Maybe we can use a factory pattern for the `JinjavaInterpreter` and set the factory on the `JinjavaConfig`. So there would be some interface `InterpreterFactory` with a default `JinjavaInterpreterFactory` that is set as a field on `JinjavaConfig`. Everywhere we create a new interpreter would instead use `config.getInterpreterFactory().newInstance(...)`. You could then create your own factory to serve your own interpreters and set that value in the config.

Roobun:
> Maybe we can use a factory pattern for the `JinjavaInterpreter` and set the factory on the `JinjavaConfig`. So there would be some interface `InterpreterFactory` with a default `JinjavaInterpreterFactory` that is set as a field on `JinjavaConfig`. Everywhere we create a new interpreter would instead use `config.getInterpreterFactory().newInstance(...)`. You could then create your own factory to serve your own interpreters and set that value in the config.

Sure. Please take a look at new changes I added.

mattcoley:
I'll merge and do a release

PR #344, opened 2019-06-25
Fixes #159 : Adds dict support for DefaultFilter

gabru-md:
Fixes #159 

Adds a method for processing the dictionary type in DefaultFilter.

Problem : Arguments were being converted to String [] before being processed by the DefaultFilter

Solution : Created a new method for processing the Object [], incase any ```arg``` is of type ```PyWrapper```. Implemented the function and created tests for the same.

😄

PR #353, opened 2019-07-22
Add codeStyleChecker for jinjava

raomuyang:
refer to #344 

I think it's inefficient to check the code style manually, then I create a `checkstyle` configuration to cover some scenarios as much as possible.

>  I wish to check the code style automatically, however, there too many old codes couldn't apply some rules


boulter:
nice! thank you.

PR #357, opened 2019-08-06
Create new class that lazily resolves

mattcoley:
This could be controversial.
Suppliers are nice to use when a variable that is set in the context that is expensive to calculate may not end up being used. Instead of handling each case individually, we can allow suppliers to be used and automatically call the `get()` function when we resolve an expression that returns a supplier.
As an example use case, if you are currently building a context map (`Map<String, Object>`) and want to lazily generate some of the values, you can replace them with suppliers, and due to this PR, the suppliers will auto-resolve.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/LazyExpression.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/LazyExpression.java
@@ -1,0 +1,21 @@
+package com.hubspot.jinjava.interpret;
+
+import java.util.function.Supplier;
+
+public class LazyExpression implements Supplier {
+
+  private final Supplier supplier;
+
+  private LazyExpression(Supplier supplier) {
+    this.supplier = supplier;
+  }
+
+  public static LazyExpression of(Supplier supplier) {
+    return new LazyExpression(supplier);
+  }
+
+  @Override
+  public Object get() {
+    return supplier.get();
+  }
+}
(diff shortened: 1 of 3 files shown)


mattcoley:
I added an explicit class `LazyExpression` which takes in a supplier and is auto-resolved instead of  auto-resolving all suppleirs.

src/main/java/com/hubspot/jinjava/interpret/LazyExpression.java
```diff
@@ -0,0 +1,21 @@
+package com.hubspot.jinjava.interpret;
+
+import java.util.function.Supplier;
+
+public class LazyExpression implements Supplier {
+
+  private final Supplier supplier;
```
boulter: Should this be `Supplier<?>`?

boulter: Or perhaps `Supplier<Object>`. Maybe that's redundant.

mattcoley: `Supplier` and `Supplier<Object>` will compile down to the same code due to how generics work in Java (IIRC) so it won't make a difference.

mattcoley: This is also only used to return a generic object from the interpreter resolve method so I think this is fine.

PR #359, opened 2019-08-13
Keep track of parent path for each block in order to resolve relative paths

liamrharwood:
The way that `block` tags are resolved is somewhat separate from the rest of the rendering process. This means that the `currentPathStack` does not correctly reflect the origin file of a block tag at the time it is resolved.

In order to support resolving relative paths within block tags, we need to keep track of the parent path for each block. This PR adds a new `BlockInfo` object which holds both the list of nodes that are already used to resolve block stubs, as well as an optional `parentPath`. When each block is rendered, we push its `parentPath` onto the `currentPathStack` which can then be used in inner tags to resolve relative paths.


src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -282,17 +283,22 @@ private void resolveBlockStubs(OutputList output, Stack<String> blockNames) {
     for (BlockPlaceholderOutputNode blockPlaceholder : output.getBlocks()) {
 
       if (!blockNames.contains(blockPlaceholder.getBlockName())) {
-        Collection<List<? extends Node>> blockChain = blocks.get(blockPlaceholder.getBlockName());
-        List<? extends Node> block = Iterables.getFirst(blockChain, null);
+        Collection<BlockInfo> blockChain = blocks.get(blockPlaceholder.getBlockName());
```
boulter: woo hoo, we're on the blockchain! Let's mine some bitcoin!

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -282,17 +283,22 @@ private void resolveBlockStubs(OutputList output, Stack<String> blockNames) {
...
 
           OutputList blockValueBuilder = new OutputList(config.getMaxOutputSize());
 
-          for (Node child : block) {
+          for (Node child : block.getNodes()) {
+            block.getParentPath().ifPresent(path -> getContext().getCurrentPathStack().push(path, lineNumber, position));
+
             blockValueBuilder.addNode(child.render(this));
+
+            block.getParentPath().ifPresent(path -> getContext().getCurrentPathStack().pop());
```
boulter: nit: path is unused, can be replaced with `()`

liamrharwood: I'm getting a compile error if do that: `Cannot infer functional interface type`
I think it's because `ifPresent` takes a `Consumer`.

PR #362, opened 2019-08-21
Add python list operations to PyList

mattcoley:
Extended `PyList` to support most of the python list operations (https://docs.python.org/3/tutorial/datastructures.html).

One issue is with the `remove()` function which I cannot overload.


boulter:
This seems like something that's not supported in Jinja. In python, you prepend with `.insert`. That would be preferable to something completely proprietary.

mattcoley:
👍 Makes sense, let me change this.

mattcoley:
Ok @boulter I added functions to `PyList` to make it similar to a regular python list.

PR #363, opened 2019-08-21
Avoid tag cycles when keeping track of parent paths for blocks

liamrharwood:
Fix for #359 because it was causing `TagCycleExceptions` when pushing a parent path onto the `currentPathStack`.

This changes the solution from that PR so that it will only push the parent path onto the stack if the stack does not already contain the parent path. For instance, this handles a situation where a template using `extends`/`block` is `include`d in a different template, so the path is already in the stack.

This also will only pop the parent path off the stack if it was pushed onto the stack to begin with.


src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -282,17 +283,28 @@ private void resolveBlockStubs(OutputList output, Stack<String> blockNames) {
...
+          List<? extends Node> superBlock = Optional.ofNullable(Iterables.get(blockChain, 1, null))
+              .map(BlockInfo::getNodes).orElse(null);
           context.setSuperBlock(superBlock);
 
           OutputList blockValueBuilder = new OutputList(config.getMaxOutputSize());
 
-          for (Node child : block) {
+          for (Node child : block.getNodes()) {
+            boolean pushedParentPathOntoStack = false;
+            if (block.getParentPath().isPresent() && !getContext().getCurrentPathStack().contains(block.getParentPath().get())) {
```
liamrharwood: > it will only push the parent path onto the stack if the stack does not already contain the parent path

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -282,17 +283,28 @@ private void resolveBlockStubs(OutputList output, Stack<String> blockNames) {
...
+              getContext().getCurrentPathStack().push(block.getParentPath().get(), lineNumber, position);
+              pushedParentPathOntoStack = true;
+            }
+
             blockValueBuilder.addNode(child.render(this));
+
+            if (pushedParentPathOntoStack) {
```
liamrharwood: > This also will only pop the parent path off the stack if it was pushed onto the stack to begin with

PR #364, opened 2019-08-22
Handle `OutputTooBigException` in the interpreter.

mattcoley:
If one of these exceptions is thrown, we lose all rendered output instead of cutting off at the max size.


mattcoley:
Yea I am going to close this now and think of some other solution we could work towards.

PR #368, opened 2019-08-28
Fix datetime zone output.

mattcoley:
Comparing http://strftime.org/ and https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html, `%z` should map to `Z` which outputs the offset `+HHMM` or `-HHMM`. `%Z` should map to `z` which outputs the shorthand name fo the timezone (for example `PDT`).


boulter:
Doesn't this have the potential to break a lot of existing pages that use %z AND %Z?

mattcoley:
I'll have to look at usage for `%Z` since we have always converted this value incorrectly and received a JIRA about it being incorrect a year ago. I am guessing it is not widely used if this bug has only been reported twice.

PR #377, opened 2019-10-16
Ipaddr filter extension 

Roobun:
* Refactor getFunctionValue() method
* Add alias for Ipv4 and Ipv6 filter
* Add ability to filter lists to take elements satisfying requirements
* Add network, private, public, gateway to possible parameters
* Implement netmask, network and broadcast for ipv6

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
@@ -9,10 +9,14 @@ import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.InvalidArgumentException;
 import com.hubspot.jinjava.interpret.InvalidReason;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.util.ForLoop;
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/IpAddrFilter.java
```diff
@@ -52,26 +71,66 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
       return false;
     }
 
-    if (args.length > 0) {
-      String function = args[0].trim();
-      return getFunctionValue(interpreter, function.toLowerCase(), object);
+    String parameter = getParameter(args);
+    if (object instanceof List) {
```
mattcoley: Should we expand this to other iterables supported by Jinjava? Generally filters will use https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/util/ObjectIterator.java#L36 to cover some edge cases. Can we add a static method there to check if this is a list/array/iterable etc. and then use `getLoop` to get a more general `Iterator`?

Roobun: I changed it from List to Iterables and added support for filtering Map objects.
But I don't think that it's worth to add static method for checking if object is Iterable or not because it's just one line with 'instanceof' operator.

PR #380, opened 2019-12-12
Fix template error line numbers

mattcoley:
Use the line number of when a path was pushed to the pathStack as the line number for the error message rather than the line number of the error within another file.

TODO: much more testing if this approach is approved


src/test/java/com/hubspot/jinjava/lib/tag/IncludeTagTest.java
```diff
@@ -101,4 +101,13 @@ public String getString(String fullName, Charset encoding,
...
   }
 
+  @Test
+  public void itSetsErrorLineNumbersCorrectly() throws IOException {
+
+    RenderResult result = jinjava.renderForResult(Resources.toString(Resources.getResource("tags/includetag/errors/base.html"), StandardCharsets.UTF_8),
+        new HashMap<>());
+
+    assertThat(result.getErrors()).hasSize(1);
+    assertThat(result.getErrors().get(0).getLineno()).isEqualTo(7);
```
mattcoley: Previously this would be on line number 4 of `base.html` highlighting `{% set test = 1 + 1 %}` as the error rather than `{{ "test"|add(1) }}` in the included file.

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -498,6 +498,13 @@ public int getPosition() {
   }
 
   public void addError(TemplateError templateError) {
+
+    // fix line numbers not matching up with source template
+    if (!context.getCurrentPathStack().isEmpty()) {
```
liamrharwood: I believe this is needed for `addAllErrors` as well. I'm also wary of any place that might be doing `getErrors().add()` instead of this method.

mattcoley:
To break down how this works:
There are two ways that a file context switch can happen:

- In something like `include` or `extend`, the other file is considered part of the original file so all we do is push to the path call stack. When we add an error, we look at the top of the call stack for the line number of the top template.
- In something like `from` or `import`, the other file is independent. A child interpreter is set up and after the file has been processed, the errors are copied over to the parent interpreter. In this case, we adjust the line numbers of the parent during the copy.

mattcoley:
I am also planning on wrapping the error messages to be clear where the error in the other file occurs.

mattcoley:
Currently the wrapped error messages look like: `Error in file 'tags/includetag/errors/error.html' on line 4: Invalid input for 'add': input with value 'test' must be a number`.
I am open for suggestions on how to improve this messaging.

TheWebTech:
@mattcoley This looks good.  Trying to think of odd edge cases where this is not going to be ideal, but can't think of any.

Minor convenience suggestion. In design manager error output would be cool if the path text linked to the file/open tab in the design manager.

mattcoley:
We can add a path attribute to the error message so that the DM can link to the other file.

PR #381, opened 2019-12-13
Add basic deferred value support for from tag

jkollmann:
This is a follow up to https://github.com/HubSpot/jinjava/pull/333

# Defer from tag

Similar to ImportTag, FromTag imports can depend on deferred values. If this is the case we want to defer the whole import. The logic for imported properties is the same, we'll defer them so that all of the following code that depends on the import will not be rendered.

Example 1: `{% from 'custom/page/web_page_basic/my_properties.html' import header_footer %}` 

➡️  `header_footer` will be deferred

Example 1: `{% from 'custom/page/web_page_basic/my_properties.html' import header_footer as footer %}` 

➡️  `footer` will be deferred

After that a DeferredValueException is thrown from inside of the FromTag in order to avoid rendering it and to reconstruct it properly.

# Defer macros

In case a `FromTag`/`ImportTag` is deferred, we want to defer all calls to imported macros. Since the execution is not a property lookup we need a different way of handling it, we can't just put an instance of a `DeferredValue` into the context. 

I decided to add `setDeferred(boolean deferred)` to `MacroFunction` to allow us to mark them as `deferred`. This way, when `AstMacroFunction` is invoked we can throw the `DeferredValueException` in order to prevent the wrapping tag from being rendered.

Example:

Assuming the property `padding` is deferred, consider this macro definition:

```
{% macro spacer(orientation = "h", size = padding) %}
    {% if orientation == "v" or orientation == "vertical" %}
        <td width="{{ size }}"></td>
    {% elif orientation == "h" or orientation == "horizontal" %}
        <tr><td height="{{ size }}"></td></tr>
    {% endif %}
{% endmacro %}
```

Before this change, the macro definition would have been deferred, however the call to the function would have rendered to `""` since the `MacroFunction` is undefined un the context.

With this PR the `spacer` MacroFunction will still be put into the context but it will have set `isDeferred()` to `true`.

So when `spacer` is called like this `{{ spacer() }}`, the invocation will throw inside of `AstMacroFunction` and `ExpressionNode` will reconstruct the image.

# Additional Notes

1. I added another 

	```
	if (value instanceof DeferredValue) {
	   throw new DeferredValueException(propertyName, interpreter.getLineNumber(), interpreter.getPosition());
	}
	```

	to `JinjavaInterpreterResolver` to make it possible to defer nested properties. Before it was only possible to defer the "root".
2. I updated `DeferredValue` to be able to hold it's original value. This is necessary in case that a deferred value is set to something different, otherwise we'd lose its state.


Joeoh:
@jboulter We have added some changes here to allow us to reconstruct macros. We also fixed a bug with the reconstructEndTags that it does not respect trim tags.

PR #385, opened 2020-01-09
Implement safe filter as SafeString and handle SafeString in filters, functions and expressions

jkollmann:
# Safe filter:

## Documentation

> Marks the value as safe which means that in an environment with automatic escaping enabled the variable will not be escaped.

Jinja docs: http://jinja.palletsprojects.com/en/2.10.x/templates/#working-with-automatic-escaping

## Approach: 

Wrap String with new class `SafeString` when the `|safe` filter is applied to the value. This way we can check the type in `ExpressionNode` and not escape it. I overwrote `toString` to make it work as expected here: https://github.com/HubSpot/jinjava/blob/16679771bf9b81a2435d9140a574ab0d15a05603/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java#L52

## Considerations: 

Since I introduce a new type here we might need to add some implementation for this in `ExpressionResolver` or `JinjavaInterpreterResolver`. The implementation is currently only tested with `|safe` being used at the end of an expression.


src/main/java/com/hubspot/jinjava/lib/filter/SafeFilter.java
```diff
@@ -25,9 +25,15 @@ public String getName() {
...
-  public Object filter(Object var, JinjavaInterpreter interpreter,
-      String... args) {
-    return var;
-  }
+  public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
+    if (var == null) {
+      return null;
+    }
+
+    if (!(var instanceof String)) {
```
jboulter: perhaps the friendlier thing to do here is just to return the passed in object

Joeoh:
@jboulter Would you mind taking a look over my changes here

src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
```diff
@@ -87,6 +88,9 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
         throw new InvalidInputException(interpreter, this, InvalidReason.NUMBER_FORMAT, object.toString());
       }
     }
+    if (object instanceof SafeString) {
```
boulter: Perhaps you could use a Delegate pattern here and implement `SafeStringFilter` which takes care of this rather than manually checking for SafeStrings in every filter?

You could have the default interface implement `filter`, then delegate to a `safeFilter` method in each of these classes, unwrapping and wrapping the `SafeString` around it.

Joeoh: Sounds good, thanks!

src/main/java/com/hubspot/jinjava/lib/filter/TrimFilter.java
```diff
@@ -28,7 +26,9 @@ public String getName() {
 
   @Override
   public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
-    return StringUtils.trim(Objects.toString(var));
+    if (var instanceof String) {
+      return safeFilter(StringUtils.trim(var.toString()), interpreter, args);
```
jkollmann: Should this be `return StringUtils.trim(Objects.toString(var));` instead?

Joeoh:
@boulter This is ready for review again. Sorry about the formatting changes but there was a lot of inconsistencies between the filters. 

My general approach here was to handle `SafeString` as an input to any function where it made sense and to preserve the `SafeString` as the output where it made sense.

For the output, if a filter takes a SafeString and manipulates it eg. replace it should return a safe string.
If it takes a SafeString and does an operation like `md5` or `len` it should not return a safe string.

There is an extra case here is `urlize` which takes a string and converts any URLs in it to <a> tags. If we want this to render properly when autoEscape is enabled then it needs to return a SafeString. This might not be the best solution for the general case, so let me know if you have any thoughts about this.

src/main/java/com/hubspot/jinjava/lib/filter/EscapeJsFilter.java
```diff
@@ -36,11 +36,18 @@
 
   @Override
   public Object filter(Object objectToFilter, JinjavaInterpreter jinjavaInterpreter, String... strings) {
-    String input = Objects.toString(objectToFilter, "");
-    LengthLimitingStringBuilder builder = new LengthLimitingStringBuilder(jinjavaInterpreter.getConfig().getMaxOutputSize());
+    if (objectToFilter instanceof String) {
+      String input = Objects.toString(objectToFilter, "");
```
boulter: nit: unnecessary variable

src/main/java/com/hubspot/jinjava/lib/filter/LowerFilter.java
```diff
@@ -1,17 +1,17 @@
 /**********************************************************************
-Copyright (c) 2014 HubSpot Inc.
+ Copyright (c) 2014 HubSpot Inc.
```
boulter: in general, it's nice to separate formatting changes like this out into a separate PR so make your PRs clean and easy to understand.

src/main/java/com/hubspot/jinjava/objects/SafeString.java
```diff
@@ -0,0 +1,19 @@
+package com.hubspot.jinjava.objects;
+
+public class SafeString {
```
boulter: could you have this extend `String`? That might simplify all your `instanceof` checks.

src/main/java/com/hubspot/jinjava/lib/filter/RootFilter.java
```diff
@@ -64,9 +65,9 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
     if (object instanceof BigInteger) {
       return calculateBigRoot(interpreter, new BigDecimal((BigInteger) object), root);
     }
-    if (object instanceof String) {
+    if (object instanceof String || object  instanceof SafeString) {
```
boulter: nit: extra space after `object`

src/main/java/com/hubspot/jinjava/objects/SafeString.java
```diff
@@ -0,0 +1,19 @@
...
+
+  private final String value;
+
+  public SafeString(String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
```
boulter: is this method needed?

src/main/java/com/hubspot/jinjava/lib/filter/LowerFilter.java
Joeoh: 100% agree

src/main/java/com/hubspot/jinjava/objects/SafeString.java
Joeoh: I wish! Unfortunately String is final so we can't. String with an extra property would have made this a lot more simple. I think if we were designing this from scratch we wouldn't use strings anywhere and instead always pass some sort of StringToken that could be safe/not safe and have some other properties.

src/main/java/com/hubspot/jinjava/objects/SafeString.java
Joeoh: @jkollmann?

src/main/java/com/hubspot/jinjava/lib/filter/TrimFilter.java
Joeoh: Why? If we know its a string, whats the need to wrap it in the call. We know we can cast or do .toString

src/main/java/com/hubspot/jinjava/objects/SafeString.java
Joeoh: Yes its needed to resolve the token to a string when rendering

Joeoh: Never-mind. Its actually not. I got mixed up with tokens.

PR #386, opened 2020-01-15
Fixes #159: Adds dict support for DefaultFilter

### LINKED ISSUE #159: Bug dict default
CODE:
```
{% set a="1" %}
{% set b=b | default("2") %}
{% set c={"key": "value"} %}
{% set d=d | default({"key": "value"}) %}

Type: {{ type(a) }} - Value: {{ a }}
Type: {{ type(b) }} - Value: {{ b }}
Type: {{ type(c) }} - Value: {{ c }}
Type: {{ type(d) }} - Value: {{ d }}
Type: {{ type(c.key) }} - Value: {{ c.key }}
Type: {{ type(d.key) }} - Value: {{ d.key }}
```
OUTPUT:
```
Type: str - Value: 1
Type: str - Value: 2
Type: dict - Value: {key=value}
Type: str - Value: {key=value}
Type: str - Value: value
Type: null - Value: 
```
What the problem with the last output?

gabru-md:
adds a fix for #159 

prev discussion : #344

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
@@ -33,15 +33,15 @@ public class DateTimeFormatFilter implements Filter {
   public Object filter(Object var, JinjavaInterpreter interpreter,
       Object... args) {
 
-	List<String> stringArgs = new ArrayList<>();
-	for(Object arg: args) {
-		stringArgs.add(Objects.toString(arg));
-	}
-	String[] filterArgs = new String[stringArgs.size()];
-	for(int i=0; i < filterArgs.length; i++) {
-		filterArgs[i] = stringArgs.get(i);
-	}
-	  
+    List<String> stringArgs = new ArrayList<>();
+    for(Object arg: args) {
+        stringArgs.add(Objects.toString(arg));
+    }
+    String[] filterArgs = new String[stringArgs.size()];
+    for(int i=0; i < filterArgs.length; i++) {
+        filterArgs[i] = stringArgs.get(i);
+    }
+
     if (args.length > 0) {
       return Functions.dateTimeFormat(var, filterArgs);
     } else {
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
@@ -1,5 +1,9 @@
 package com.hubspot.jinjava.lib.filter;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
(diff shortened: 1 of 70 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
```diff
@@ -27,10 +31,19 @@ public String getName() {
 
   @Override
   public Object filter(Object var, JinjavaInterpreter interpreter,
-      String... args) {
+      Object... args) {
+
+    List<String> stringArgs = new ArrayList<>();
```
jboulter: You could do this all in one loop. In fact, you don't need `stringArgs` at all. I know this was just moved from another file, but might as well clean it up while we're in here.

src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
```diff
@@ -41,11 +42,11 @@
 public class DivideFilter implements Filter {
 
   @Override
-  public Object filter(Object object, JinjavaInterpreter interpreter, String... arg) {
+  public Object filter(Object object, JinjavaInterpreter interpreter, Object... arg) {
     if (arg.length < 1) {
       throw new TemplateSyntaxException(interpreter, getName(), "requires 1 number (divisor) argument");
     }
-    String toMul = arg[0];
+    String toMul = arg[0] == null ? null : Objects.toString(arg[0]);
```
jboulter: `Objects.toString` takes care of nulls for you.

src/main/java/com/hubspot/jinjava/lib/filter/DefaultFilter.java
```diff
@@ -61,7 +64,7 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
       return object;
     }
 
-    return args[0];
+    return (args[0] instanceof PyWrapper) ? args[0] : Objects.toString(args[0]);
```
jboulter: this is the only real significant change, right?

src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
gabru-md: [Docs](https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html) state that it returns `"null"` and not `null`.
Also if I use `Objects.toString(arg[0])` then the [following test](https://github.com/HubSpot/jinjava/blob/master/src/test/java/com/hubspot/jinjava/lib/filter/AdvancedFilterTest.java#L57) breaks. Therefore had to ensure that `arg[0] == null` check was made.

src/main/java/com/hubspot/jinjava/lib/filter/DefaultFilter.java
gabru-md: Yes. basically checking if the argument is an instance of PyWrapper class then we do not need to convert it into String.

src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
gabru-md: done :+1:

src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
boulter: Ah, good to know.

PR #387, opened 2020-01-17
Revert "Fixes #159: Adds dict support for DefaultFilter"

### LINKED ISSUE #159: Bug dict default

boulter:
Reverts HubSpot/jinjava#386 as it's breaking some internal HubSpot builds.


gabru-md:
@boulter why were the changes reverted? :sweat_smile:

boulter:
It broke some of our builds because the change is not backwards compatible. It's also ambiguous which method will be called, especially if there are no varargs. We have a couple of options. 

- Name the method something else like `doFilter`.
- Use the existing `AdvancedFilter` which already supports named arguments.

gabru-md:
@boulter using AdvancedFilter makes sense I tried it on my local machine.
I'll create a PR so that you can review and perhaps that won't break anything

PR #390, opened 2020-01-21
adds PyList support to ForTag

### LINKED ISSUE #321: for tuple not working
<ul class="navigation">
    {% for href, caption in [('index.html', 'Index'), ('downloads.html', 'Downloads'), ('products.html', 'Products')] %}
      <li><a href="{{href|e}}">{{caption|e}}</a></li>
    {% endfor %}
    </ul>

Not working

gabru-md:
Fixes #321 

This pull request is wrt the above-mentioned Issue.
The support for For-Tag was limited to Maps and the else block in ForTag.java file could not handle tuples(list of tuples) types. This PR adds the needed code as well as the supporting tests.

Cheers!

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
@@ -19,7 +19,6 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Map.Entry;
 
 import org.apache.commons.lang3.StringUtils;
@@ -165,10 +164,10 @@ public class ForTag implements Tag {
               interpreter.getContext().put(loopVar, entryVal);
             } else if (List.class.isAssignableFrom(val.getClass())) {
               List<Object> entries = ((PyList) val).toList();
-              String entryVal = null;
+              Object entryVal = null;
               // safety check for size
               if (entries.size() >= loopVarIndex) {
-                entryVal = Objects.toString(entries.get(loopVarIndex));
+                entryVal = entries.get(loopVarIndex);
               }
               interpreter.getContext().put(loopVar, entryVal);
             } else {
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
@@ -19,6 +19,7 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Map.Entry;
 
 import org.apache.commons.lang3.StringUtils;
@@ -32,6 +33,7 @@ import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter.InterpreterScopeClosable;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.objects.DummyObject;
+import com.hubspot.jinjava.objects.collections.PyList;
 import com.hubspot.jinjava.tree.Node;
 import com.hubspot.jinjava.tree.TagNode;
 import com.hubspot.jinjava.util.ForLoop;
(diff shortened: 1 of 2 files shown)


src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
```diff
@@ -205,6 +205,27 @@ public void testForLoopWithDates() {
     assertEquals(new PyishDate(testDate).toString(), rendered);
   }
 
+  @Test
+  public void testTuplesWithPyList() {
+	  String template = "{% for href, caption in [('index.html', 'Index'), ('downloads.html', 'Downloads'), ('products.html', 'Products')] %}" + 
+	  		"<li><a href=\"{{href|e}}\">{{caption|e}}</a></li>\n" +
```
boulter: any reason for running this through the escape filter (e) here?

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
```diff
@@ -205,6 +205,27 @@ public void testForLoopWithDates() {
...
+  @Test
+  public void testTuplesWithPyList() {
+	  String template = "{% for href, caption in [('index.html', 'Index'), ('downloads.html', 'Downloads'), ('products.html', 'Products')] %}" + 
+	  		"<li><a href=\"{{href|e}}\">{{caption|e}}</a></li>\n" + 
+	  		"{% endfor %}";
+	  String expected = "<li><a href=\"index.html\">Index</a></li>\n" +
+	  		"<li><a href=\"downloads.html\">Downloads</a></li>\n" +
+	  		"<li><a href=\"products.html\">Products</a></li>\n";
+	  
+	  String rendered = jinjava.render(template, context);
```
boulter: this could be broken out into a separate test

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
```diff
@@ -205,6 +205,27 @@ public void testForLoopWithDates() {
...
+	  assertEquals(rendered, expected);
+	  
+	  template = "{% for a, b, c in [(1,2,3), (4,5,6)] %}"
+	  		+ "<p>{{a}} {{b}} {{c}}</p>\n"
+	  		+ "{% endfor %}";
+	  expected = "<p>1 2 3</p>\n"
+	  		+ "<p>4 5 6</p>\n";
+	  rendered = jinjava.render(template, context);
+	  assertEquals(rendered, expected);
```
boulter: could you also add a test that confirms that this works for non-string values? Perhaps add 1 + 2 + 3?

src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
```diff
@@ -159,6 +162,14 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
                 entryVal = entry.getValue();
               }
 
+              interpreter.getContext().put(loopVar, entryVal);
+            } else if (List.class.isAssignableFrom(val.getClass())) {
+              List<Object> entries = ((PyList) val).toList();
+              String entryVal = null;
+              // safety check for size
+              if (entries.size() >= loopVarIndex) {
+                entryVal = Objects.toString(entries.get(loopVarIndex));
```
boulter: can you add a test for this case?

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
gabru-md: this was mentioned in issue #321, therefore, I used the exact same thing to test the functionality out. there is no specific reason for choosing this template. :sweat_smile:

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
gabru-md: done :+1:

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
gabru-md: I've added tests for the same.
Also, I've modified the ForTag.java file to accommodate the changes.

src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
gabru-md: This is a safety check only but I've added tests where the `jinjava.render` is expected to throw an exception when there are lesses values to unpack. :+1:

src/test/java/com/hubspot/jinjava/lib/tag/ForTagTest.java
boulter: Ok, can you remove anything that's not relevant to the test then? We don't want to be testing the escape filter here.

gabru-md: done!

PR #392, opened 2020-01-30
Replicate string bug

mattcoley:
Caused by https://github.com/HubSpot/jinjava/pull/385


Joeoh:
Closing. Bug resolved.

PR #394, opened 2020-01-31
Implement safe filter as SafeString and handle SafeString in filters, functions and expressions

Joeoh:
# Safe filter:

Replacement for reverted PR: https://github.com/HubSpot/jinjava/pull/385
Implemenetation is much simplier than previous PR. Any filter which is receiving a SafeString object will now be handled without any changes to the underlying filter. If a function filter receives a SafeString and returns a String it will be wrapped in a SafeString. This is undesirable in some cases such as the ipAddr filter which does not return some direct version of the input. An override has been added to disable this wrapping for such filters. `preserveSafeString` can be set to false to disable the wrapping.

## Documentation

> Marks the value as safe which means that in an environment with automatic escaping enabled the variable will not be escaped.

Jinja docs: http://jinja.palletsprojects.com/en/2.10.x/templates/#working-with-automatic-escaping

## Approach: 

Wrap String with new class `SafeString` when the `|safe` filter is applied to the value. This way we can check the type in `ExpressionNode` and not escape it. I overwrote `toString` to make it work as expected here: https://github.com/HubSpot/jinjava/blob/16679771bf9b81a2435d9140a574ab0d15a05603/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java#L52

## Considerations: 

Since I introduce a new type here we might need to add some implementation for this in `ExpressionResolver` or `JinjavaInterpreterResolver`. The implementation is currently only tested with `|safe` being used at the end of an expression.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsStringExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsStringExpTest.java
@@ -4,6 +4,7 @@ import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.objects.SafeString;
 
 @JinjavaDoc(
     value = "Return true if object is a string",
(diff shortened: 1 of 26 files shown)


Joeoh:
@mattcoley @boulter 

This PR is ready for review when you get a chance. This simplifies the implementation a great deal and gives us more confidence that the behaviour will remain consistent as the individual filters have been changed minimally. This resolves this issue replicated in this PR: https://github.com/HubSpot/jinjava/pull/392

PR #396, opened 2020-02-10
prevent CallStacks from parenting themselves

boulter:
We've seen some instances where `CallStack.pop()` gets itself into a cycle. This could only be possible if it were a parent of itself. 

Adding some error logging here to help track it down.


src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
```diff
@@ -33,7 +33,7 @@
     snippets = {
         @JinjavaSnippet(
             code = "{% set escape_string = \"{{This markup is printed as text}}\" %}\n" +
-                "{{ escape_string|escapeJinjava }}")
+                "{{ escape_string|escape_jinjava }}")
```
boulter: documentation error I noticed

boulter:
Trying to devise a test case for this made me believe this isn't actually possible since you cannot reference self in the constructor before it's created.

PR #398, opened 2020-02-11
Fix implementation for slice filter

mattcoley:
Slice filter was not implemented as according to spec. When doing `list|slice(3)` we should not be padding non-divisible slices with null values. When using `list|slice(3, 'value')` we should pad non-divisible slices with `value`. https://jinja.palletsprojects.com/en/2.11.x/templates/#slice

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
@@ -1,8 +1,10 @@
 package com.hubspot.jinjava.lib.filter;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.apache.commons.lang3.math.NumberUtils;
 
-import com.google.common.collect.Iterators;
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -47,7 +51,30 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
     }
 
     int slices = NumberUtils.toInt(args[0], 3);
```
liamrharwood: Any reason we default to `3` here?

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -47,7 +51,30 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
+        currentList = new ArrayList<>(slices);
+        result.add(currentList);
+      }
+      currentList.add(next);
+      i++;
+    }
+
+    if (args.length > 1 && currentList != null) {
+      while (currentList.size() < slices) {
+        currentList.add(args[1]);
```
liamrharwood: nit: assign `args[1]` to something like `fill_with` for clarity

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
mattcoley: Ideally we would throw an error if `args[0]` wasn't provided, however we can't do that without breaking existing functionality. Somebody could be using a raw `|slice` out there.

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
mattcoley: Done

PR #401, opened 2020-02-14
add a config option to turn off expression tracking

boulter:
For large templates, tracking which functions, expressions, and filters that were used during execution can use a lot of CPU and memory, especially since they are copied all the way up the context tree. 

This adds an option to disable tracking unless it is needed.

For a sufficiently large template, this can be called a lot.

image

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -41,6 +41,7 @@ public class JinjavaConfig {
   private final boolean trimBlocks;
   private final boolean lstripBlocks;
 
+  private final boolean trackResolved;
   private final boolean readOnlyResolver;
   private final boolean enableRecursiveMacroCalls;
   private final int maxMacroRecursionDepth;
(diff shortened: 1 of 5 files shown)


src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -278,8 +283,13 @@ public Builder withMaxStringLength(long maxStringLength) {
       return this;
     }
 
-    public Builder withInterperterFactory(InterpreterFactory interperterFactory) {
```
boulter: fixed typo in Interpreter

hs-lsong:
I think this should make it performant a bit, but our own usage requires checking of those resolved values, expression. I am not sure if we can disable the tracking even if it is available.

boulter:
Rethinking this.

hs-lsong:
Another finding I have was, for the profiling timing, as the number of function calls is large, the profiler's overhead is very large. I tried to just log the timing of those small function calls, especially the `wrap`. Its timing is not signification compare to other big functions.

boulter:
Another thought was to make the tracking pluggable so we can provide an implementation that only tracks the things we care about.

Issue #379, opened 2019-10-29
.update function does not support variables


TheWebTech:
```
{% set dict_var = {} %}
{% set custom_key = "string"%}
{% set test = dict_var.update({ custom_key : 'Jake'}) %}
{{ dict_var }}
```
You would expect
```
{"string":"Jake"}
```
Instead the custom_key is treated as if "custom_key" is a string.
```
{custom_key:"Jake"}
```

This came up as a developer was trying to generate keys in a dict as they don't know what all of the keys will be at all times.

The update function was based on the python update function.
In python this works
```
d = {1: "one", 2: "three"}
varname = "three"
d1 = {varname: "two"}

# updates the value of key 2
d.update(d1)
print(d)
```

CC: @mattcoley

boulter:
See https://github.com/HubSpot/jinjava/pull/403

boulter:
This is reverted for now. https://github.com/HubSpot/jinjava/pull/404

PR #403, opened 2020-02-18
evalulate variables in dict keys

boulter:
While you could previously add variables as elements in list, this was not possible with dict keys. 

So 
```
{% set keyName = "key1" %}
{% set test = {keyName: "value1"} %}
{{ test[keyName]}}
```

will now return "value1"

This could have adverse side effects. We were liberally accepting unquoted key names for dicts. Now they will be evaluated.

This does work in Jinja 2.11.1 for lists and dicts:
```
>>> Template('Hello {% set key = "key1" %}{% set d = [key, "val"] %}{{ d[0] }}').render()
u'Hello key1'
```

```
>>> Template('Hello {% set d = [key, "val"] %}{{ d[0] }}').render(key="key1")
u'Hello key1'
```

For compatibility, key names that evaluate to `null` (meaning probably undefined) will be treated as strings. 

Fixes https://github.com/HubSpot/jinjava/issues/379

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
@@ -33,7 +33,7 @@ public class AstDict extends AstLiteral {
       if (entry.getKey() instanceof AstString) {
         key = Objects.toString(entry.getKey().eval(bindings, context));
       } else if (entry.getKey() instanceof AstIdentifier) {
-        key = ((AstIdentifier) entry.getKey()).getName();
+        key = entry.getKey().eval(bindings, context).toString();
       } else {
         throw new TemplateStateException("Dict key must be a string or identifier, was: " + entry.getKey());
       }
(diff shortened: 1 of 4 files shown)


src/test/java/com/hubspot/jinjava/objects/collections/PyListTest.java
```diff
@@ -105,4 +105,11 @@ public void itReturnsNegativeOneForMissingObjectForIndexWithinBounds() {
     assertThat(jinjava.render("{% set test = [10, 20, 30, 10, 20, 30] %}" +
         "{{ test.index(999, 1, 5) }}", Collections.emptyMap())).isEqualTo("-1");
   }
+
+  @Test
+  public void itInterpretsVariables() {
```
boulter: this just confirms that variables do work in lists.

src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
```diff
@@ -33,7 +33,11 @@ public Object eval(Bindings bindings, ELContext context) {
       if (entry.getKey() instanceof AstString) {
         key = Objects.toString(entry.getKey().eval(bindings, context));
       } else if (entry.getKey() instanceof AstIdentifier) {
-        key = ((AstIdentifier) entry.getKey()).getName();
+        Object result = entry.getKey().eval(bindings, context);
+        if (result == null) {
+          continue;
```
boulter: A possible fallback here to preserve the existing behavior would be to return the identifier name when the key evaluates to `null`.

mattcoley: I think the fallback is safer.

PR #405, opened 2020-02-28
Return empty string for un-evaluated lazy expression.

hs-lsong:


src/main/java/com/hubspot/jinjava/interpret/LazyExpression.java
```diff
@@ -19,13 +19,20 @@ public static LazyExpression of(Supplier supplier, String image) {
...
   }
 
   @Override
   public String toString() {
     return image;
   }
+
+  @JsonValue
+  public Object getJsonValue() {
+    return jsonValue == null ? "" : jsonValue;
```
mattcoley: Could it evaluate the supplier but catch any exception and return empty string?

hs-lsong: That is a good question. It depends what the serialization wants. What I wanted here is to get the current state -- no more evaluation. Maybe there is a need to finalize any un-evaluated expression.

mattcoley: `LazyExpression` was created as a more under-the-hood optimization where we could use it to prevent computing something if it wasn't needed. It would be still useful to compute when debugging information.

hs-lsong: When debugging, you either just use the toString(), and/or with get() to evaluate it.

mattcoley: I am talking from the developer perspective. Developers depend on the developer info button that displays the context.

hs-lsong: How does it work now? It will throw exceptions for serialization. If it is not using the json serialization, it does not affect anything.

mattcoley: I think the serialization breaks if the `get()` method throws an exception (in the case of the NPE). We should also at least fall back on the `image` property.

hs-lsong: I assume when we store the lazyExpression in somewhere, we assume to store the value, only when we need it. However, the serialization should not execute the function because it may have side effect -- which may cause concurrent modification exception. I thought the `image` is the expression itself. I think they should be evaluate to empty, but I am OK if you change to keep it.

PR #406, opened 2020-02-28
Enable prettier-java for formatting

jhaber:
@mattcoley @boulter @hs-lsong


jhaber:
Getting a findbugs error, probably because the basepom upgrade picked up a new version of findbugs

mattcoley:
Do we have a guide for non-hubspot devs contributing to this repo if prettier is enabled?

jhaber:
Code should automatically get reformatted on build so they shouldn't need to do anything special besides building before committing. To just reformat without running a full build you can do `mvn prettier:write`

mattcoley:
👍 We can add that to the readme.

jhaber:
Suppressed the findbugs warning, it might actually be due to a Guava upgrade rather than a findbugs change (seems like they now use a `@NullableDecl` annotation which findbugs isn't treating as nullable)

jhaber:
It's probably lost in the massive diff, but I also slimmed down the checkstyle configuration to remove things that are superseded by prettier

PR #407, opened 2020-03-06
Remove hacky replaceL behavior

mattcoley:
This behavior has been here since initial commit but is pretty strange. Basically if we detect the format is a Java format rather than strftime format, we return the Java format back but replace all `LLLL` instances with `MMMM`. This means we can't manually specify `LLLL` when that is the appropriate format we want.


boulter:
Probably had something do with python compatibility? Worth checking for differences in any case.

mattcoley:
That's my plan. I am not sure how this would affect python compatibility since python would be using the stftime format with `%`, not the Java datetime format that we convert the strftime into.

mattcoley:
I ran the diff report 5 times. The only diffs were expected. Languages that have genitive months are now properly using the genitive format.

Issue #402, opened 2020-02-16
TokenSymbols not customizable


KidsDontPlay:
Hi,
can you make the symbols cuztomizable like in [Nunjucks](https://mozilla.github.io/nunjucks/api.html#customizing-syntax). Currently they are hardcoded in `TokenScannerSymbols`.
Maybe they should be part of the `JinjavaConfig`.

KidsDontPlay:
Or is there another way to use Jinjava and Vue at the same time.

gabru-md:
@mattcoley close this issue?

PR #410, opened 2020-03-10
adds support for Custom Token Scanner Symbols

gabru-md:
This pull request aims at implementing the Custom Token Symbols class for Jinjava. The goal is to allow the user to specify the tokens that he/she wants to use. It can be extending using the TokenScannerSymbols abstract class and then filling the abstract methods.

A few lines inside the tree package have been changed from `switch-case` to `if-else` since `switch-case` allows only constant expressions to be used with them. The changes to the tests which were failing initially due to the change have also been made and it should work fine now. Furthermore a JUnit test file has also been added in support of the changes made.

related issue: https://github.com/HubSpot/jinjava/issues/402 

@KidsDontPlay perhaps this solves your problem.

cheers!

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
@@ -50,14 +50,8 @@ public class ExpressionNode extends Node {
     String result = Objects.toString(var, "");
 
     TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
-    String expressionBegins = new StringBuilder()
-      .append(symbols.getTokenPrefixChar())
-      .append(symbols.getTokenExprStartChar())
-      .toString();
-    String expressionWithTag = new StringBuilder()
-      .append(symbols.getTokenPrefixChar())
-      .append(symbols.getTokenTagChar())
-      .toString();
+    String expressionBegins = symbols.getExpressionStart();
+    String expressionWithTag = symbols.getExpressionStartWithTag();
 
     if (interpreter.getConfig().isNestedInterpretationEnabled()) {
       if (
--- a/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
@@ -48,17 +48,24 @@ public class ExpressionNode extends Node {
     }
 
     String result = Objects.toString(var, "");
-    
+
     TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
-    String expressionBegins = new StringBuilder().append(symbols.getTokenPrefixChar())
-                                  .append(symbols.getTokenExprStartChar()).toString();
-    String expressionWithTag = new StringBuilder().append(symbols.getTokenPrefixChar())
-                                  .append(symbols.getTokenTagChar()).toString();
+    String expressionBegins = new StringBuilder()
+      .append(symbols.getTokenPrefixChar())
+      .append(symbols.getTokenExprStartChar())
+      .toString();
+    String expressionWithTag = new StringBuilder()
+      .append(symbols.getTokenPrefixChar())
+      .append(symbols.getTokenTagChar())
+      .toString();
 
     if (interpreter.getConfig().isNestedInterpretationEnabled()) {
       if (
         !StringUtils.equals(result, master.getImage()) &&
-        (StringUtils.contains(result, expressionBegins) || StringUtils.contains(result, expressionWithTag))
+        (
+          StringUtils.contains(result, expressionBegins) ||
+          StringUtils.contains(result, expressionWithTag)
+        )
       ) {
         try {
           result = interpreter.renderFlat(result);
--- a/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
@@ -22,6 +22,7 @@ import com.hubspot.jinjava.objects.SafeString;
 import com.hubspot.jinjava.tree.output.OutputNode;
 import com.hubspot.jinjava.tree.output.RenderedOutputNode;
 import com.hubspot.jinjava.tree.parse.ExpressionToken;
+import com.hubspot.jinjava.tree.parse.TokenScannerSymbols;
 import com.hubspot.jinjava.util.Logging;
 import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
@@ -47,11 +48,17 @@ public class ExpressionNode extends Node {
     }
 
     String result = Objects.toString(var, "");
+    
+    TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
+    String expressionBegins = new StringBuilder().append(symbols.getTokenPrefixChar())
+                                  .append(symbols.getTokenExprStartChar()).toString();
+    String expressionWithTag = new StringBuilder().append(symbols.getTokenPrefixChar())
+                                  .append(symbols.getTokenTagChar()).toString();
 
     if (interpreter.getConfig().isNestedInterpretationEnabled()) {
       if (
         !StringUtils.equals(result, master.getImage()) &&
-        (StringUtils.contains(result, "{{") || StringUtils.contains(result, "{%"))
+        (StringUtils.contains(result, expressionBegins) || StringUtils.contains(result, expressionWithTag))
       ) {
         try {
           result = interpreter.renderFlat(result);
--- a/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
@@ -50,15 +50,13 @@ public class ExpressionNode extends Node {
     String result = Objects.toString(var, "");
 
     TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
-    String expressionBegins = symbols.getExpressionStart();
-    String expressionWithTag = symbols.getExpressionStartWithTag();
 
     if (interpreter.getConfig().isNestedInterpretationEnabled()) {
       if (
         !StringUtils.equals(result, master.getImage()) &&
         (
-          StringUtils.contains(result, expressionBegins) ||
-          StringUtils.contains(result, expressionWithTag)
+          StringUtils.contains(result, symbols.getExpressionStart()) ||
+          StringUtils.contains(result, symbols.getExpressionStartWithTag())
         )
       ) {
         try {
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -113,7 +113,7 @@ public class TagNode extends Node {
       .getConfig()
       .getTokenScannerSymbols();
     StringBuilder stringBuilder = new StringBuilder();
-    stringBuilder.append(symbols.getExprStartChar()).append(symbols.getTagChar());
+    stringBuilder.append(symbols.getExpressionStartWithTag());
     if (
       getChildren() != null &&
       !getChildren().isEmpty() &&
@@ -126,7 +126,7 @@ public class TagNode extends Node {
       stringBuilder.append(symbols.getTrimChar());
     }
 
-    stringBuilder.append(symbols.getTagChar()).append(symbols.getExprEndChar());
+    stringBuilder.append(symbols.getExpressionEndWithTag());
     return stringBuilder.toString();
   }
 }
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -108,7 +108,10 @@ public class TagNode extends Node {
   }
 
   public String reconstructEnd() {
-    TokenScannerSymbols symbols = JinjavaInterpreter.getCurrent().getConfig().getTokenScannerSymbols();
+    TokenScannerSymbols symbols = JinjavaInterpreter
+      .getCurrent()
+      .getConfig()
+      .getTokenScannerSymbols();
     StringBuilder stringBuilder = new StringBuilder();
     stringBuilder
       .append(symbols.getTokenExprStartChar())
@@ -124,10 +127,8 @@ public class TagNode extends Node {
     if (getMaster().isRightTrimAfterEnd()) {
       stringBuilder.append(symbols.getTokenTrimChar());
     }
-    
-    stringBuilder
-      .append(symbols.getTokenTagChar())
-      .append(symbols.getTokenExprEndChar());
+
+    stringBuilder.append(symbols.getTokenTagChar()).append(symbols.getTokenExprEndChar());
     return stringBuilder.toString();
   }
 }
--- a/src/main/java/com/hubspot/jinjava/tree/TagNode.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TagNode.java
@@ -111,23 +111,23 @@ public class TagNode extends Node {
     TokenScannerSymbols symbols = JinjavaInterpreter.getCurrent().getConfig().getTokenScannerSymbols();
     StringBuilder stringBuilder = new StringBuilder();
     stringBuilder
-      .append(symbols.TOKEN_EXPR_START_CHAR())
-      .append(symbols.TOKEN_TAG_CHAR());
+      .append(symbols.getTokenExprStartChar())
+      .append(symbols.getTokenTagChar());
     if (
       getChildren() != null &&
       !getChildren().isEmpty() &&
       getChildren().getLast().getMaster().isRightTrim()
     ) {
-      stringBuilder.append(symbols.TOKEN_TRIM_CHAR());
+      stringBuilder.append(symbols.getTokenTrimChar());
     }
     stringBuilder.append(" ").append(getEndName()).append(" ");
     if (getMaster().isRightTrimAfterEnd()) {
-      stringBuilder.append(symbols.TOKEN_TRIM_CHAR());
+      stringBuilder.append(symbols.getTokenTrimChar());
     }
     
     stringBuilder
-      .append(symbols.TOKEN_TAG_CHAR())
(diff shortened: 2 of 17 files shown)


gabru-md:
even after running java-prettier on my end travis seems to be failing for the same reason of unformatted code. :man_facepalming:

src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
```diff
@@ -51,7 +51,10 @@ public OutputNode render(JinjavaInterpreter interpreter) {
     if (interpreter.getConfig().isNestedInterpretationEnabled()) {
       if (
         !StringUtils.equals(result, master.getImage()) &&
-        (StringUtils.contains(result, "{{") || StringUtils.contains(result, "{%"))
+        (
+          StringUtils.contains(result, "{{") ||
```
boulter: Looks like this should use the scanner symbols as well.

src/main/java/com/hubspot/jinjava/tree/TreeParser.java
```diff
@@ -89,40 +89,38 @@ public Node buildTree() {
...
+    if (token.getType() == symbols.TOKEN_FIXED()) return text(
+      (TextToken) token
+    ); else if (
+      token.getType() == symbols.TOKEN_EXPR_START()
+    ) return expression((ExpressionToken) token); else if (
+      token.getType() == symbols.TOKEN_TAG()
+    ) return tag((TagToken) token); else if (
+      token.getType() == symbols.TOKEN_NOTE()
+    ) {
+      if (!token.getImage().endsWith("#}")) {
```
boulter: Use `TOKEN_NOTE_CHAR` here?

src/main/java/com/hubspot/jinjava/tree/parse/DefaultTokenScannerSymbols.java
```diff
@@ -0,0 +1,58 @@
...
+  char TOKEN_FIXED_CHAR = 0;
+  char TOKEN_NOTE_CHAR = '#';
+  char TOKEN_TAG_CHAR = '%';
+  char TOKEN_EXPR_START_CHAR = '{';
+  char TOKEN_EXPR_END_CHAR = '}';
+  char TOKEN_NEWLINE_CHAR = '\n';
+  char TOKEN_TRIM_CHAR = '-';
+
+  @Override
+  public char TOKEN_PREFIX_CHAR() {
```
boulter: should be `getTokenPrefix()`. Same for the others below.

src/main/java/com/hubspot/jinjava/tree/parse/Token.java
```diff
@@ -93,22 +88,33 @@ public String toString() {
...
+      image,
+      lineNumber,
+      startPosition
+    ); else if (tokenKind == symbols.TOKEN_NOTE()) return new NoteToken(
+      image,
+      lineNumber,
+      startPosition
+    ); else if (
+      tokenKind == symbols.TOKEN_EXPR_START()
+    ) return new ExpressionToken(image, lineNumber, startPosition); else if (
```
boulter: `else if` should be on a new line

src/main/java/com/hubspot/jinjava/tree/parse/Token.java
```diff
@@ -93,22 +88,33 @@ public String toString() {
...
-        );
-    }
+  static Token newToken(
+    int tokenKind,
+    TokenScannerSymbols symbols,
+    String image,
+    int lineNumber,
+    int startPosition
+  ) {
+    if (tokenKind == symbols.TOKEN_FIXED()) return new TextToken(
```
boulter: please always wrap bodies of if statements in `{ }`

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
```diff
@@ -15,24 +15,59 @@
...
-  int TOKEN_FIXED = TOKEN_FIXED_CHAR;
-  int TOKEN_NOTE = TOKEN_NOTE_CHAR;
-  int TOKEN_TAG = TOKEN_TAG_CHAR;
-  int TOKEN_EXPR_START = TOKEN_EXPR_START_CHAR;
-  int TOKEN_EXPR_END = TOKEN_EXPR_END_CHAR;
-  int TOKEN_NEWLINE = TOKEN_NEWLINE_CHAR;
-  int TOKEN_TRIM = TOKEN_TRIM_CHAR;
+public abstract class TokenScannerSymbols {
+
+  public abstract char TOKEN_PREFIX_CHAR();
```
boulter: same comment as before

src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
gabru-md: I've made the changes. Although I'm not quite sure to if the `{{` is to be replaced by 
`TOKEN_PREFIX_CHAR + TOKEN_EXPR_START_CHAR`
or is it just 
`TOKEN_EXPR_START_CHAR + TOKEN_EXPR_START_CHAR`.

Please review it once again.
Cheers!

src/main/java/com/hubspot/jinjava/tree/TreeParser.java
gabru-md: done

src/main/java/com/hubspot/jinjava/tree/parse/DefaultTokenScannerSymbols.java
gabru-md: changes made!

src/main/java/com/hubspot/jinjava/tree/parse/Token.java
gabru-md: done.

gabru-md:
@boulter 

while debugging the code after making the changes, I noticed something weird [here](https://github.com/HubSpot/jinjava/pull/410/files#diff-09bee8d4b3f12e40cf4f966d5192cc10L44).

```java
this.expr = WhitespaceUtils.unwrap(image, "{{", "}}");
```
```java
// inside the unwrap function
 return s.substring(start + prefix.length(), end - suffix.length() + 1);
```
the unwrap function does not care about the prefix and the suffix being passed into the function. The only purpose it serves is when the function calculates the length of the prefix and the suffix to trim the image and return the expression. Should there not be a validation inside the `unwrap` function that asserts a check that the correct prefix and suffix has been used in the image?

I was just wondering if it should be added as well?

boulter:
@gabru-md, could you run `mvn clean verify` on your build so prettier runs and check it in? It's hard to see what the differences are since there's so many whitespace and formatting changes. Thanks!

gabru-md:
@boulter, I executed `mvn clean verify` and applied the diff created by the prettier. Please see if it is reviewable now.

src/main/java/com/hubspot/jinjava/tree/parse/DefaultTokenScannerSymbols.java
```diff
@@ -0,0 +1,58 @@
...
+  char TOKEN_FIXED_CHAR = 0;
+  char TOKEN_NOTE_CHAR = '#';
+  char TOKEN_TAG_CHAR = '%';
+  char TOKEN_EXPR_START_CHAR = '{';
+  char TOKEN_EXPR_END_CHAR = '}';
+  char TOKEN_NEWLINE_CHAR = '\n';
+  char TOKEN_TRIM_CHAR = '-';
+
+  @Override
+  public char getTokenPrefixChar() {
```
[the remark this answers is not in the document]
gabru-md: okay. I will remove that !

src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
```diff
@@ -48,10 +49,23 @@ public OutputNode render(JinjavaInterpreter interpreter) {
 
     String result = Objects.toString(var, "");
 
+    TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
+    String expressionBegins = new StringBuilder()
```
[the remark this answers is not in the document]
gabru-md: I didn't quite get what you're trying to say. Can you please explain a little bit so that I can make the changes?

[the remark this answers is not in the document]
boulter: Sure. Add `getExpressionStart()` and `getExpressionEnd()` methods to `TokenScannerSymbols` and ensure they're only built once, rather than every time a node is rendered.

gabru-md:
@boulter,
I am facing an error. As you might have noticed that I've introduced a new test file called `CustomTokenScannerSymbolsTest.java`. The problem that I'm facing when I try to run tests on Eclipse. When I run the `CustomTokenScannerSymbolsTest.java` file then the new tests seem to pass whereas when I execute all the tests in the test package then the new tests which I've created starts failing and the reason I could identify is due to the fact that the interpreter still uses
`DefaultTokenScannerSymbols` instead of the newly created `CustomTokens`.

I cannot quite seem to figure out why this behavior is happening.

src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
```diff
@@ -48,10 +49,17 @@ public OutputNode render(JinjavaInterpreter interpreter) {
 
     String result = Objects.toString(var, "");
 
+    TokenScannerSymbols symbols = interpreter.getConfig().getTokenScannerSymbols();
+    String expressionBegins = symbols.getExpressionStart();
+    String expressionWithTag = symbols.getExpressionStartWithTag();
```
boulter: unnecessary variables as they are only used once.

src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -108,24 +108,25 @@ public String reconstructImage() {
...
   public String reconstructEnd() {
+    TokenScannerSymbols symbols = JinjavaInterpreter
+      .getCurrent()
+      .getConfig()
+      .getTokenScannerSymbols();
     StringBuilder stringBuilder = new StringBuilder();
-    stringBuilder
-      .append(TokenScannerSymbols.TOKEN_EXPR_START_CHAR)
-      .append(TokenScannerSymbols.TOKEN_TAG_CHAR);
+    stringBuilder.append(symbols.getExprStartChar()).append(symbols.getTagChar());
```
boulter: can you add a method to construct this string once as well?

src/main/java/com/hubspot/jinjava/tree/TagNode.java
```diff
@@ -108,24 +108,25 @@ public String reconstructImage() {
...
     stringBuilder.append(" ").append(getEndName()).append(" ");
     if (getMaster().isRightTrimAfterEnd()) {
-      stringBuilder.append(TokenScannerSymbols.TOKEN_TRIM_CHAR);
+      stringBuilder.append(symbols.getTrimChar());
     }
-    stringBuilder
-      .append(TokenScannerSymbols.TOKEN_TAG_CHAR)
-      .append(TokenScannerSymbols.TOKEN_EXPR_END_CHAR);
+
+    stringBuilder.append(symbols.getTagChar()).append(symbols.getExprEndChar());
```
boulter: same here

src/main/java/com/hubspot/jinjava/tree/parse/TagToken.java
```diff
@@ -15,20 +15,22 @@
...
 
-import static com.hubspot.jinjava.tree.parse.TokenScannerSymbols.TOKEN_TAG;
-
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.util.WhitespaceUtils;
 
 public class TagToken extends Token {
   private static final long serialVersionUID = -4927751270481832992L;
+  private final int TOKEN_TAG;
```
boulter:
```suggestion
  private final int tokenTag;
```
only static variables should be in upper case.

src/main/java/com/hubspot/jinjava/tree/parse/TagToken.java
```diff
@@ -15,20 +15,22 @@
...
 
   private String tagName;
   private String rawTagName;
   private String helpers;
 
   public TagToken(String image, int lineNumber, int startPosition) {
     super(image, lineNumber, startPosition);
+    TOKEN_TAG =
```
boulter:
```suggestion
    tokenTag =
```

src/main/java/com/hubspot/jinjava/tree/parse/TextToken.java
```diff
@@ -15,15 +15,17 @@
...
 package com.hubspot.jinjava.tree.parse;
 
-import static com.hubspot.jinjava.tree.parse.TokenScannerSymbols.TOKEN_FIXED;
-
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import org.apache.commons.lang3.StringUtils;
 
 public class TextToken extends Token {
   private static final long serialVersionUID = -6168990984496468543L;
+  private final int TOKEN_FIXED;
```
boulter:
```suggestion
  private final int tokenFixed;
```

src/main/java/com/hubspot/jinjava/tree/parse/TextToken.java
```diff
@@ -15,15 +15,17 @@
...
 
   public TextToken(String image, int lineNumber, int startPosition) {
     super(image, lineNumber, startPosition);
+    TOKEN_FIXED =
```
boulter:
```suggestion
    tokenFixed =
```

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
```diff
@@ -15,24 +15,86 @@
...
+
+  public int getNewline() {
+    return getNewlineChar();
+  }
+
+  public int getTrim() {
+    return getTrimChar();
+  }
+
+  public String getExpressionStart() {
```
boulter: it would good to fill this out with all the other starting and closing sequences that we'll eventually need.

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
```diff
@@ -15,24 +15,86 @@
...
+
+  public abstract char getExprStartChar();
+
+  public abstract char getExprEndChar();
+
+  public abstract char getNewlineChar();
+
+  public abstract char getTrimChar();
+
+  public int getPrefix() {
```
boulter: why are all these methods that just call abstract methods needed?

src/test/java/com/hubspot/jinjava/tree/parse/CustomTokenScannerSymbolsTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Before
+  public void setup() {
+    config =
+      JinjavaConfig.newBuilder().withTokenScannerSymbols(new CustomTokens()).build();
+    jinjava = new Jinjava(config);
+    jinjava.getGlobalContext().put("numbers", Lists.newArrayList(1L, 2L, 3L, 4L, 5L));
+  }
+
+  @Test
+  public void testsThatCustomTokensDoesNotFail() {
```
boulter: can you make these rspec-style? `itRendersWithCustomTokens`, etc.

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
gabru-md: there are places when an integer is being required and there are separate places where a character is required during parsing the template. Also, it will be better if the user is allowed to fill in the functions which return `char` and not the ASCII value instead. Somebody who wants custom tokens will only have to fill in the tokens in form of characters and do not need to write down their ASCII values.
Also, I could've used those abstract methods instead since they can be cast to integers as well but the problem lies with the fact that when I was writing the `if..else` conditions then it was becoming a lot unclear since the casting had to be done during the comparison. Therefore I thought It'd be better to not mess up the code readability a lot and simply have users fill out the functions which return `char` and then return their `int` value using the `getPrefix` functions instead. I believe it was a better way to make changes.
I hope my explanation is fine. If you feel that it is not okay then I'd try and make the changes.

gabru-md:
I believe that the command line builds should not fail if the same isn't failing for me in Eclipse. But since they are then I will work on it and try to get to the depth of the problem and commit a fix as soon as I am done with it. Since this does not seem to be a high-priority feature then I wish I can take the time I want to get it all fixed. I will perhaps need some more reviews from your end though and will tag you once changes seem fine to me.

src/main/java/com/hubspot/jinjava/tree/TagNode.java
gabru-md: sure. I might have missed this out in the last commit.
I will add this to the changes as well.

src/test/java/com/hubspot/jinjava/tree/parse/CustomTokenScannerSymbolsTest.java
gabru-md: sure.

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
gabru-md: :+1:

gabru-md:
@boulter,
please review. I've made the changes required and the tests are now passing.
cheers!

src/main/java/com/hubspot/jinjava/tree/ExpressionNode.java
boulter: @gabru-md

src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java
boulter: @gabru-md are all these in there now?

gabru-md: Yes, all the required combinations of sequences are in here right now.

KidsDontPlay:
@gabru-md 
thanks for your effort

Issue #411, opened 2020-03-16
Parser throws only one error


Roobun:
buildTree() method of class TreeParser adds only last found error in template.

E.g. after parsing that template (which has two not closed For tags) interpreter has one error (about not having ending tag for last one for operation)
"{% for item in list %}\n" +
"{% for element in vector %}\n"

It also applies to other tags

Roobun:
Added PR related to that problem: #413

PR #413, opened 2020-03-19
Several parse errors

Roobun:
Allow buildTree() method of TreeParser to save info about several parse errors in template

Fixes https://github.com/HubSpot/jinjava/issues/411


boulter:
Nice job. Thank you!

PR #415, opened 2020-03-24
New release (pending PR merge)

mattcoley:
Waiting on https://github.com/HubSpot/jinjava/pull/410 to be merged but should be good to release after that.


boulter:
Should be GTG @mattcoley

PR #419, opened 2020-04-02
store reference to TokenScannerSymbols in Nodes and Tokens

boulter:
Before, we would parse nodes and tokens using one set of symbols from the config and potentially use a different set when reconstructing a tag. Instead, let's just store a reference to the symbols on each node so we can use it when rendering.

Follow up to https://github.com/HubSpot/jinjava/pull/410

@gabru-md


gabru-md:
I apologise. I did not observe that two seperate set of symbols may be used. I thought I had done it successfully. 
This PR definitely adds more insight to the project for me. Thank You.

boulter:
> I apologise. I did not observe that two seperate set of symbols may be used. I thought I had done it successfully.

Not a problem.

PR #430, opened 2020-04-15
Track Java method invocations, start whitelist

mattcoley:
Tracks usage of `method()` on objects. There are some we should allow (for example `append`), so started a whitelist of allowed methods that skip tracking.


boulter:
How do you plan to aggregate these results? Perhaps you could build in the whitelist now as something configurable (can be updated in realtime) and throw errors whenever you come across a new one, but not fail execution yet. Then have another config to actually block execution when the errors have subsided.

mattcoley:
This tracking does not affect interpretation ATM. I was planning on pulling this information from the `RenderResult` and adding to the internal template response object. Then I can run a report on current templates to find usage. Using this information, I could build a default whitelist setting that could be part of the config.

boulter:
Yeah, what I'm saying is that I think you could put all the code in place now and then build up the whitelist from sentry logs.

PR #431, opened 2020-04-16
Add whitelist for host Java classes

mattcoley:
Replacement for https://github.com/HubSpot/jinjava/pull/430

Adds a class whitelist per recommendation from @jboulter that can be configured.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -55,7 +55,8 @@ public class JinjavaConfig {
   private final long maxStringLength;
   private InterpreterFactory interpreterFactory;
   private TokenScannerSymbols tokenScannerSymbols;
-  private Set<Class> allowedHostClasses;
+  private final Set<Class> allowedHostClasses;
+  private final boolean enforceHostWhitelist;
 
   public static Builder newBuilder() {
     return new Builder();
@@ -85,7 +86,8 @@ public class JinjavaConfig {
       0,
       interpreterFactory,
       new DefaultTokenScannerSymbols(),
-      Collections.emptySet()
+      Collections.emptySet(),
+      false
     );
   }
 
@@ -114,7 +116,8 @@ public class JinjavaConfig {
       0,
       new JinjavaInterpreterFactory(),
       new DefaultTokenScannerSymbols(),
-      Collections.emptySet()
+      Collections.emptySet(),
+      false
     );
   }
 
@@ -137,7 +140,8 @@ public class JinjavaConfig {
     long maxStringLength,
     InterpreterFactory interpreterFactory,
     TokenScannerSymbols tokenScannerSymbols,
-    Set<Class> allowedHostClasses
+    Set<Class> allowedHostClasses,
+    boolean enforceHostWhitelist
   ) {
     this.charset = charset;
     this.locale = locale;
@@ -158,6 +162,7 @@ public class JinjavaConfig {
     this.interpreterFactory = interpreterFactory;
     this.tokenScannerSymbols = tokenScannerSymbols;
     this.allowedHostClasses = allowedHostClasses;
+    this.enforceHostWhitelist = enforceHostWhitelist;
   }
 
   public Charset getCharset() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -229,6 +241,14 @@ public void setTokenScannerSymbols(TokenScannerSymbols tokenScannerSymbols) {
     this.tokenScannerSymbols = tokenScannerSymbols;
   }
 
+  public Set<Class> getAllowedHostClasses() {
```
boulter, line 244: maybe it's worth wrapping these in a `MethodSecurityConfig` or `ClassSecurityConfig` class or something? The names are not terribly easy to understand. 

Then you could just call this allowedClasses.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -229,6 +241,14 @@ public void setTokenScannerSymbols(TokenScannerSymbols tokenScannerSymbols) {
...
+    return allowedHostClasses;
+  }
+
+  public boolean getEnforceHostWhitelist() {
```
boulter, line 248: this could just be `isEnforced` on MethodSecurityConfig.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -126,17 +128,40 @@ private boolean isRestrictedClass(Object o) {
...
+      }
+    }
+
+    JinjavaInterpreter interpreter = JinjavaInterpreter.getCurrent();
+    if (interpreter != null && interpreter.getConfig() != null) {
+      if (!interpreter.getConfig().getAllowedHostClasses().contains(o.getClass())) {
+        if (interpreter.getConfig().getEnforceHostWhitelist()) {
+          return true;
+        } else {
+          Logging.ENGINE_LOG.error(
```
boulter, line 157: Does this actually end up as an error in our logging or do we need to throw an exception?

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -126,17 +128,40 @@ private boolean isRestrictedClass(Object o) {
...
+            "Warning: Template using method on restricted class '{}'",
```
boulter, line 158: Not necessarily a template.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -126,17 +128,40 @@ private boolean isRestrictedClass(Object o) {
...
+      } else if (o.getClass().getPackage().getName().startsWith("com.hubspot.jinjava")) {
+        return false;
+      }
+    }
+
+    JinjavaInterpreter interpreter = JinjavaInterpreter.getCurrent();
+    if (interpreter != null && interpreter.getConfig() != null) {
+      if (!interpreter.getConfig().getAllowedHostClasses().contains(o.getClass())) {
+        if (interpreter.getConfig().getEnforceHostWhitelist()) {
+          return true;
```
boulter, line 155: Will we also need method-level security?

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
mattcoley: Isn't everything technically a template to Jinjava? Trying to keep this copy specific to Jinjava rather than HubSpot.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
mattcoley: I believe it ends up as an error in our logging depending on our logging configuration (we have removed these log statements in the past IIRC to clean up our logs).

jhaber:
Maybe the entire EL resolver chain should be a configuration option?

mattcoley:
I think it makes sense to make the `ELResolver` a configuration option so we can specify our own `JinjavaBeanELResolver` without affecting other users of this repo. Any thoughts on that approach @boulter?

boulter:
Yeah, I think that will simplify things a bit.

PR #432, opened 2020-04-16
Allow ELResolver to be specified in Config

mattcoley:
Alternative to https://github.com/HubSpot/jinjava/pull/431

This will allow us to connect our custom ELResolver.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -15,6 +15,7 @@
  **********************************************************************/
 package com.hubspot.jinjava;
 
+import com.hubspot.jinjava.el.JinjavaInterpreterResolver;
 import com.hubspot.jinjava.interpret.Context;
 import com.hubspot.jinjava.interpret.Context.Library;
 import com.hubspot.jinjava.interpret.InterpreterFactory;
@@ -30,6 +31,7 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import javax.el.ELResolver;
 
 public class JinjavaConfig {
   private final Charset charset;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -307,7 +309,10 @@ public Builder withMaxMacroRecursionDepth(int maxMacroRecursionDepth) {
     }
 
     public Builder withReadOnlyResolver(boolean readOnlyResolver) {
```
mattcoley, line 311: I wanted to keep the existing behavior consistent so kept this method in for the builder.

boulter:
instead of making it part of the config, how about allowing it to be set on the Jinjava instance like it is for `ResourceLocator`. Then you can create whatever implementation you want.

mattcoley:
The ELResolver used right now is determined by the config (through the `readOnlyResolver` setting). Is there any advantage to exposing the ELResolver at the Jinjava level? That would mean you would need a separate instance of Jinjava for every ELResolver you would like to use vs. being able to reuse the same Jinjava object with different ELResolver's specified by the config.

mattcoley:
I realized I forgot to add the Builder method to the config for specifying your own ELResolver.

boulter:
I think it should be possible to provide your own chain of resolvers. The readOnlyResolver seems a little too inflexible. Maybe you could just pass your `CompositeELResolver` to config if you don't want to add a setter to `Jinjava`.

mattcoley:
The config accepts any `ELResolver` so you can pass in your own `CompositeELResolver` to the config. The `readOnlyResolver` boolean was an existing config value that I didn't want to remove for backwards compatibility, I just moved the logic to map the boolean value to one of the two default values to the config from the JinjavaInterpreterResolver

PR #438, opened 2020-05-05
Defer variables used in deferred nodes

Joeoh:
As a follow on to the discussion in the PR here: https://github.com/HubSpot/jinjava/pull/421

This PR proposes a solution for two problems: Handling variables used within a deferred block and handling variables used in deferred blocks that occur at lower scopes.

###  Handling variables used within a deferred block
As discussed in https://github.com/HubSpot/jinjava/pull/421 we want a way to mark variables as deferred _while_ we are processing so that any subsequent uses of the variable are not processed. This is achieved here by scanning the deferred node each time and finding variables used within it. These variables are then put back into the context as DeferredValues with their original value. Any subsequent uses of these variables will result in a DeferredValueException and be handled like any other deferred variable. Before adding DeferredValuesUtils this kind of logic only existed in internal HS code -I believe because it is a small bit hacky. Before now, anyone else making use of the deferred functionality would have had to have their own way to rebuild a Deferred Render Context for their 2nd stage render.

I have also added support for finding yet undefined variables used within deferred blocks and marking them as deferred for example
```
{% if deferred = 'foo' %} 
{% set newVariable = 'deferred was foo' %}
{% else %}
{% set newVariable = 'deferred was not foo' %}
{% endif %}
{{ newVariable }}
```
Previously this would have output newVariable as an empty undefined var. Now it will be put in the context as deferred and so will be re-output as {{ newVariable }}
 

### handling variables used in deferred blocks that occur at lower scopes
I have observed issues in using the previous implementation of deferred variables in a HS setup where things are rendered in lower contexts (eg modules) and any deferred values are lost. These changes copy any deferred values to the parent scope, again with their original value. 

I can think of some cases where this might go wrong. If you had a template, that imported the same module multiple times, the first instance of the module would muddle the global scope with its values. I don't think this more global scope should cause issues as any code making use of deferred values should be preserved and the order of execution should be consistent. I added handling for deferred values to SetTag to handle this case.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
@@ -6,6 +6,7 @@ import com.google.common.collect.Sets;
 import com.hubspot.jinjava.interpret.Context;
 import com.hubspot.jinjava.interpret.DeferredValue;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.lib.tag.SetTag;
 import com.hubspot.jinjava.tree.ExpressionNode;
 import com.hubspot.jinjava.tree.Node;
 import com.hubspot.jinjava.tree.TagNode;
@@ -26,7 +27,7 @@ public class DeferredValueUtils {
   private static final Pattern TEMPLATE_TAG_PATTERN = Pattern.compile(TEMPLATE_TAG_REGEX);
 
   private static final Pattern SET_TAG_PATTERN = Pattern.compile(
-    "set " + TEMPLATE_TAG_REGEX
+    SetTag.TAG_NAME + " " + TEMPLATE_TAG_REGEX
   );
 
   public static HashMap<String, Object> getDeferredContextWithOriginalValues(
--- a/src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
@@ -1,0 +1,174 @@
+package com.hubspot.jinjava.util;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import com.hubspot.jinjava.interpret.Context;
+import com.hubspot.jinjava.interpret.DeferredValue;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.tree.ExpressionNode;
+import com.hubspot.jinjava.tree.Node;
+import com.hubspot.jinjava.tree.TagNode;
+import com.hubspot.jinjava.tree.TextNode;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.StringJoiner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DeferredValueUtils {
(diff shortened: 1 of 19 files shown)


src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DeferredValueUtils {
+  private static final String TEMPLATE_TAG_REGEX = "(\\w+(?:\\.\\w+)*)";
+  private static final Pattern TEMPLATE_TAG_PATTERN = Pattern.compile(TEMPLATE_TAG_REGEX);
+
+  private static final Pattern SET_TAG_PATTERN = Pattern.compile(
+    "set " + TEMPLATE_TAG_REGEX
```
boulter, line 29: Could you use the name defined here? https://github.com/HubSpot/jinjava/blob/5cb246370d8f2bdb304fdcfa22cfa57dfab2c14f/src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java#L65

src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.StringJoiner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class DeferredValueUtils {
+  private static final String TEMPLATE_TAG_REGEX = "(\\w+(?:\\.\\w+)*)";
```
boulter, line 25: I don't think this will work with multivariable assignments like `set myvar1, myvar2, myvar3, myvar4 = ...` I'm not sure you'd want to though as that might make things very complex.

src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
```diff
@@ -183,4 +187,169 @@ public void itDefersMacro() {
...
+    interpreter.render(template);
+    Object varInScope = interpreter.getContext().get("testVar");
+    assertThat(varInScope).isInstanceOf(DeferredValue.class);
+    DeferredValue varInScopeDeferred = (DeferredValue) varInScope;
+    assertThat(varInScopeDeferred.getOriginalValue()).isEqualTo("testvalue");
+  }
+
+  @Test
+  public void itPutsDeferredVariablesOnParentScopes() {
+    String template = "";
```
boulter, line 248: Might be nicer to put these in fixture files like some of the other tests.

src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
```diff
@@ -183,4 +187,169 @@ public void itDefersMacro() {
...
+    String output = interpreter.render(template);
+    assertThat(interpreter.getContext()).containsKey("varSetInside");
+    Object varSetInside = interpreter.getContext().get("varSetInside");
+    assertThat(varSetInside).isInstanceOf(DeferredValue.class);
+    assertThat(output).contains("{{ varSetInside }}");
+  }
+
+  @Test
+  public void itMarksVariablesUsedAsMapKeysAsDeferred() {
+    /*    {% set week_number = contact.MOweekno %}
```
boulter, line 319: comment needed?

src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
Joeoh: Nice catch. I added this regex to be more specific for looking for sets as we don't have the existing context to compare matches to filter out noise. I think I will change this (and the other usages) to filter out any reserved tokens instead.

Issue #444, opened 2020-05-21
missing operator expression tests


jfudman:
There are missing expression tests for most operators. I came across this trying to use a **greater than** expression test for the [select attribute filter's exp_test parameter](https://designers.hubspot.com/docs/hubl/filters#selectattr). 
```
|selectattr( attribute_name, exp_test, val )
```

I see expression test names like [`equalto`](https://designers.hubspot.com/docs/hubl/operators-and-expression-tests#equalto), but nothing equivalent to any other [comparison operator](https://designers.hubspot.com/docs/hubl/operators-and-expression-tests#comparison). For example, the following operators do not have expression tests:
- `<`
- `<=`
- `>`
- '>=`

I'm not quite sure of the best way to implement it. I see one of 3 ways.
1. use lowercase syntax such as `equalto` (i.e.`greaterthan`). But `greaterthanorequalto` might be a little long
2. use api query syntax similar to [hubdb filtering](https://developers.hubspot.com/docs/methods/hubdb/v2/get_table_rows) (i.e. `gt`, `gte`, `lt`, `lte`, etc).
3. Use the actual comparison symbol

I would prefer option 2 or 3.

mattcoley:
We are missing some built-in tests that exist in Jinja https://jinja.palletsprojects.com/en/2.11.x/templates/#list-of-builtin-tests. We should add any missing tests.

PR #446, opened 2020-05-25
Jinja expression test parity

tectonic8:
For issue #444. 

This PR adds the missing [expression tests from Jinja](itTruncatesStringToConfigLimit) along with their aliases, namely `eq`, `ne`, `lt`, `lte`, `gt`, `gte`, `boolean`, `integer`, `float`, `true`, `false`, `escaped`, `in`. The only expression test that I've omitted is `callable` which returns true for built-in tests and user-defined macros. We have little existing support for this sort of type checking in Jinjava.

Although there are risks, I've changed some existing Jinjava behavior to match Jinja functionality. 
- `is iterable` and `is sequence` should return `true` for Maps and Strings
- A PyMap should implement Iterable and iterate over the keys. This is needed for `is in` and `is containing` to work with dicts. I did not change how `ObjectIterator` used by the Jinjava `for` loop iterates over the values, which is not to spec. The `ObjectIterator` also fails to iterate over a string one character at a time. 
-  The `escape` filter now returns a SafeString. This is needed for `is escaped` to match the spec. I'm unsure of this change because it adds a level of indirection in the unit tests.
- Membership testing should handle types more permissively. More details in the comments.

cc @TheWebTech for potential [doc](https://designers.hubspot.com/docs/hubl/operators-and-expression-tests) changes


src/main/java/com/hubspot/jinjava/el/ext/CollectionMembershipOperator.java
```diff
@@ -7,39 +7,43 @@
...
-            return Boolean.TRUE;
+    if (maybeIterable instanceof Iterable) {
+      for (Object element : (Iterable) maybeIterable) {
+        if (element == null) {
+          if (value == null) {
+            return true;
           }
         } else {
           try {
-            return collection.contains(converter.convert(o1, value.getClass()));
```
tectonic8, line 37: What we had before was that we would cast the value to the type of the first element of the collection, and then return on that result. So `2 is in ['one', 2]` would fail because `2` gets casted to `'2'`.

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -53,6 +64,19 @@
...
+    IDENTIFIER,
+    EQ,
+    NE,
+    LT,
+    LE,
+    GT,
+    GE,
+    TRUE,
+    FALSE,
+    CollectionMembershipOperator.TOKEN.getSymbol()
```
tectonic8, line 77: CollectionMembershipOperator.TOKEN.getSymbol() is just `EXTENSION` which is non-ideal because we really just want to add `IN` to this set, but we can't add symbols.

src/main/java/com/hubspot/jinjava/lib/exptest/IsSequenceExpTest.java
```diff
@@ -25,9 +27,6 @@ public String getName() {
 
   @Override
   public boolean evaluate(Object var, JinjavaInterpreter interpreter, Object... args) {
-    return (
-      var != null &&
-      (var.getClass().isArray() || Iterable.class.isAssignableFrom(var.getClass()))
-    );
+    return isIterable(var);
```
tectonic8, line 30: It is questionable to define iterables and sequences as the same thing, but this follows the Jinja spec.

src/test/java/com/hubspot/jinjava/lib/exptest/IsContainingExpTestTest.java
```diff
@@ -30,14 +30,14 @@ public void itPassesOnContainedValue() {
...
-  public void itFailsOnNullContainedValue() {
+  public void itPassesOnNullContainedValue() {
     assertThat(
         jinjava.render(
           String.format(CONTAINING_TEMPLATE, "[1, 2, null]", "null"),
           new HashMap<>()
         )
       )
-      .isEqualTo("fail");
+      .isEqualTo("pass");
```
tectonic8, line 40: This is a breaking change. I don't see why this should fail before, and it does not fail in Jinja.

src/test/java/com/hubspot/jinjava/lib/exptest/IsInExpTestTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  public void testNull() {
+    assertThat(jinjava.render("{{ null is in [null] }}", new HashMap<>()))
+      .isEqualTo("true");
+    assertThat(jinjava.render("{{ null is in [2] }}", new HashMap<>()))
+      .isEqualTo("false");
+    assertThat(jinjava.render("{{ 2 is in [null] }}", new HashMap<>()))
+      .isEqualTo("false");
+    assertThatThrownBy(() -> jinjava.render("{{ 2 is in null }}", new HashMap<>()))
+      .hasMessageContaining("1st argument with value 'null' must be iterable");
```
tectonic8, line 51: Should passing in an object that is not an iterable simply return false or throw an exception?

src/test/java/com/hubspot/jinjava/lib/exptest/IsWithinExpTestTest.java
```diff
@@ -53,12 +53,6 @@ public void itFailsOnNullValueNotInSequence() {
       .isEqualTo("fail");
   }
 
-  @Test
-  public void itFailsOnNullSequence() {
-    assertThat(jinjava.render(String.format(IN_TEMPLATE, "2", "null"), new HashMap<>()))
-      .isEqualTo("fail");
-  }
-
```
tectonic8, line 61: This is related to whether we should return false or throw an exception. In Python, `2 in None` would throw an exception.

src/test/java/com/hubspot/jinjava/lib/filter/EscapeFilterTest.java
```diff
@@ -20,11 +20,11 @@ public void setup() {
...
   public void testEscape() {
-    assertThat(f.filter("", interpreter)).isEqualTo("");
-    assertThat(f.filter("me & you", interpreter)).isEqualTo("me &amp; you");
-    assertThat(f.filter("jared's & ted's bogus journey", interpreter))
+    assertThat(f.filter("", interpreter).toString()).isEqualTo("");
+    assertThat(f.filter("me & you", interpreter).toString()).isEqualTo("me &amp; you");
+    assertThat(f.filter("jared's & ted's bogus journey", interpreter).toString())
       .isEqualTo("jared&#39;s &amp; ted&#39;s bogus journey");
-    assertThat(f.filter(1, interpreter)).isEqualTo("1");
+    assertThat(f.filter(1, interpreter).toString()).isEqualTo("1");
```
tectonic8, line 27: This is because the escape filter now returns a `SafeString`. Is this fine?

mattcoley:
This is exciting! I am going to go through the changes tomorrow.

src/main/java/com/hubspot/jinjava/lib/exptest/IsEscapedExpTest.java
```diff
@@ -0,0 +1,41 @@
...
+public class IsEscapedExpTest implements ExpTest {
+
+  @Override
+  public String getName() {
+    return "escaped";
+  }
+
+  @Override
+  public boolean evaluate(Object var, JinjavaInterpreter interpreter, Object... args) {
+    return var instanceof SafeString;
```
Joeoh, line 39: SafeString is kind of confusing, a string being "safe" means that you trust the contents and that it won't be escaped. 
I'm wondering if we should also check the context for interpreter.getContext().isAutoEscape . 

Something like !(var instanceof SafeString) && interpreter.getContext().isAutoEscape()

tectonic8: Hi, thanks for reviewing!

I have it this way because the Jinja implementation just checks for whether the value in the variable is marked safe rather than whether it actually is safe, so whether or not the output enters that `if` statement you linked might not actually be relevant. Also, the expression test is evaluated before that condition evaluates, so I think it makes sense that the test returns whether the value is marked safe rather than whether its value would get escaped.

tectonic8: Screen Shot 2020-05-26 at 9 52 28 AM
Here's the Jinja behavior that I modeled this off of.

Joeoh: Ok thanks for explaining. I guess it makes sense otherwise we would end up with double escaped strings.

src/main/java/com/hubspot/jinjava/el/ext/CollectionMembershipOperator.java
mattcoley: Can you check the performance of this? I believe I did this back in https://github.com/HubSpot/jinjava/pull/314/files for performance reasons but would be good to double check. We could also split this into a separate PR to be safe.

src/main/java/com/hubspot/jinjava/lib/exptest/IsFloatExpTest.java
```diff
@@ -0,0 +1,30 @@
...
+public class IsFloatExpTest implements ExpTest {
+
+  @Override
+  public String getName() {
+    return "float";
+  }
+
+  @Override
+  public boolean evaluate(Object var, JinjavaInterpreter interpreter, Object... args) {
+    return var instanceof Double || var instanceof Float;
```
mattcoley, line 28: I think we need to check `BigDecimal` as well (see the `AddFilter`). BigDecimal support was added in the math filters since python supports arbitrary precision and we had to come up with a workaround for Java.

src/main/java/com/hubspot/jinjava/lib/exptest/IsIntegerExpTest.java
```diff
@@ -0,0 +1,30 @@
...
+public class IsIntegerExpTest implements ExpTest {
+
+  @Override
+  public String getName() {
+    return "integer";
+  }
+
+  @Override
+  public boolean evaluate(Object var, JinjavaInterpreter interpreter, Object... args) {
+    return var instanceof Integer || var instanceof Long;
```
mattcoley, line 28: Same here, we will have to handle arbitrarily long numbers.

src/test/java/com/hubspot/jinjava/lib/exptest/IsWithinExpTestTest.java
mattcoley: Escaping is tricky. I would leave out the changes to the escape filter and the addition of the escaped expression test in a separate PR to be safe.

tectonic8: That makes sense.

tectonic8: I removed the escaping, but also this code snippet here is related to the whether searching something that isn't an iterable should return false as it does currently or throw an exception. Can you comment on that?

src/main/java/com/hubspot/jinjava/el/ext/CollectionMembershipOperator.java
tectonic8: There is a slight performance hit. For checking the existence of a float in a list of 1,000,0000 strings, it takes 2-6 seconds with the current code and 1-2 seconds in the master branch. This code can be slightly optimized by storing the converted values in a type->object map. With that we see closer to the 2 second range. For a list of 100,000 strings, the difference is about 400 ms vs 300 ms. It seems to me that there isn't a huge difference until we reach very large scales. What do you think we should do?

tectonic8:
Hey Matt, is there still interest in this feature? It was last pending your response [here](https://github.com/HubSpot/jinjava/pull/446#discussion_r439574413) and [here](https://github.com/HubSpot/jinjava/pull/446#discussion_r439577460). I understand if you prefer that I close the PR.

bszabo:
Hi Hartek,

 

I think the distribution list on this PR got inadvertently crossed. Matt may not have seen your missive.

 

Bernard

tectonic8:
Thanks for informing me, Bernard! Sorry for the @ and no rush, but @mattcoley is there still interest in this feature?

Hartek

mattcoley:
I get all the emails :) I was out yesterday and today moving apartments. I have this on my list to finish as well as your other PR @tectonic8. I am finishing up one major issue before working on these two.

PR #448, opened 2020-05-29
Add global prop to context

Joeoh:
Changes to mark context as global when created


hs-lsong:
Since global context has no parent, is this redundant to checking `parent == null`?

src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -94,25 +95,38 @@ public Context setValidationMode(boolean validationMode) {
   private boolean validationMode = false;
 
   public Context() {
-    this(null, null, null);
+    this(null, null, null, false);
+  }
+
+  public Context(boolean global) {
```
boulter, line 101: Is this constructor necessary? I think you could just use one of the others and pass in nulls. I find it confusing when a method with one parameter could be take two types of arguments.

Joeoh:
Going to close this out and just check the parent's parent to determine global instead here: https://github.com/HubSpot/jinjava/pull/449

PR #454, opened 2020-06-11
Support expressions in selectattr/rejectattr again, more efficiently v2

mattcoley:
Take 2. I will fix this up.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
@@ -10,11 +10,11 @@ import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.lib.exptest.ExpTest;
 import com.hubspot.jinjava.util.ForLoop;
 import com.hubspot.jinjava.util.ObjectIterator;
-import com.hubspot.jinjava.util.Variable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 
 @JinjavaDoc(
   value = "Filters a sequence of objects by applying a test to an attribute of an object and only selecting the ones with the test succeeding.",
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
```diff
@@ -110,20 +110,32 @@ protected Object applyFilter(
...
-      )
-      .resolve(val);
       if (acceptObjects == expTest.evaluate(attrVal, interpreter, expArgs)) {
         result.add(val);
       }
     }
+    interpreter.getContext().removeResolvedExpression(expression);
 
     return result;
   }
```
samukce, line 132: There is an unused method parameter in this method. Should we remove it? `Map<String, Object> kwargs`

src/main/java/com/hubspot/jinjava/lib/filter/SelectAttrFilter.java
```diff
@@ -110,20 +110,32 @@ protected Object applyFilter(
...
+
+  private String generateTempVariable() {
+    return "jj_temp_" + Math.abs(ThreadLocalRandom.current().nextInt());
+  }
+
+  private String generateTempVariable(String tempValue, String expression) {
+    return String.format("%s.%s", tempValue, expression).trim();
```
samukce, line 139: The `String.format` has some performance issues. So, for curiosity, I ran a [benchmark](https://gist.github.com/samukce/c12a520c2f5ca8db19f560a9572bc39f) to compare the approaches.

Those are the results:
```bash
Benchmark                                                Mode  Cnt   Score   Error  Units
StringConcatSimulation.concatWithConcatFunction            ss   25   1.900 ± 0.120  ms/op
StringConcatSimulation.concatWithPlus                      ss   25   2.079 ± 0.438  ms/op
StringConcatSimulation.concatWithStringBuilder             ss   25   1.871 ± 0.436  ms/op
StringConcatSimulation.concatWithStringBuilderWithCount    ss   25   1.720 ± 0.255  ms/op
StringConcatSimulation.concatWithStringFormat              ss   25  15.145 ± 2.836  ms/op
```

I would suggest using StringBuilder instead.
```suggestion
    return new StringBuilder()
                .append(tempValue)
                .append(".")
                .append(expression)
                .toString()
                .trim();
```

Issue #414, opened 2020-03-23
Equalto operator doesn't work in "or" statement (== does)


Zoggy-Stardust:
Looks like `== `works in an `or` statement, but `equalto` doesn't. They're documented as doing the same thing [here](https://designers.hubspot.com/docs/hubl/operators-and-expression-tests#equalto), which could also be incorrect.
This doesn't work:
```
{% set myStr = 'world' %}
{% if myStr is equalto 'hey' -%}
{% elif myStr is equalto 'hello' or myStr is equalto 'world' -%}
  should work, doesn't
{% endif %}
```
this does:
```
{% set myStr = 'world' %}
{% if myStr is equalto 'hey' -%}
{% elif myStr == 'hello' or myStr == 'world' -%}
  does work
{% endif %}
```

samukce:
hello @Zoggy-Stardust , you can find the fix here https://github.com/HubSpot/jinjava/pull/455;)

samukce:
@Joeoh  and @mattcoley should we close this issue?

Joeoh:
Yes thanks for resolving @samukce

PR #455, opened 2020-06-12
Fix for "Equalto operator doesn't work in "or" statement (== does)"

samukce:
This is a fix to be able to support multiple equals when using multiple conditions.

This issue was related here #414 

This commit is the main resolution for this fix https://github.com/HubSpot/jinjava/commit/8cb5142832a310f22cab8c1c2a9650a4cf4f8334#diff-dfa299c9356b094d756062285eaf6efdL426-R426

Also, after applied I could notice that for `is no`" expression had the same issue, so I applied too.

Feedback is welcome.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
+++ b/src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
@@ -122,7 +122,7 @@ public class DeferredTest {
   @Test
   public void itResolveEqualToInOrCondition() {
     String output = interpreter.render(
-            "{% if 'a' is equalto 'b' or 'a' is equalto 'a' %}{{deferred}}{% endif %}"
+      "{% if 'a' is equalto 'b' or 'a' is equalto 'a' %}{{deferred}}{% endif %}"
     );
     assertThat(output).isEqualTo("{{deferred}}");
   }
--- a/src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
+++ b/src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
@@ -120,6 +120,14 @@ public class DeferredTest {
   }
 
   @Test
+  public void itResolveEqualToInOrCondition() {
+    String output = interpreter.render(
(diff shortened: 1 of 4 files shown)


samukce:
hello @mattcoley , could you take a look at this PR?

src/test/java/com/hubspot/jinjava/interpret/DeferredTest.java
```diff
@@ -119,6 +119,14 @@ public void itResolvesIfTagWherePossible() {
     assertThat(interpreter.getErrors()).isEmpty();
   }
 
+  @Test
+  public void itResolveEqualToInOrCondition() {
```
Joeoh, line 123: Thanks for adding tests here too. Would you mind adding another for a deferred value on the RHS of an if?

samukce: Sure @Joeoh. Before that, when you say RHS you mean right side of the equation, right? If so, can you give me an example about this test scenario to make clear I got your test idea?

Joeoh: Yes the right hand side. Similar to itPreservesIfTag above we want to ensure that the if is preserved if a deferred variable is used in the Or condition. so for example: `{% if 'a' is equalto 'b' or 'a' is equalto deferred %}preserved{% endif %}

samukce: done @Joeoh .

Joeoh: Thanks! Can you use a var for the string instead

samukce: pls, let me know if you have any other suggestion @Joeoh

samukce:
If everything looks good for you could you merge this pr @Joeoh

Joeoh:
Be good to get another set of 👀  on this @mattcoley

Issue #457, opened 2020-06-23
Fix `fromjson` filter to accept a JSON array


mattcoley:
`fromjson` https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/lib/filter/FromJsonFilter.java#L33 only accepts a map. It should allow a JSON array as well.`

PR #458, opened 2020-06-23
Map Json string to Object instead of HashMap

AlpriElse:
Resolves https://github.com/HubSpot/jinjava/issues/457
Related to https://hubspotdev.slack.com/archives/C2YH119N2/p1592854855031000

This pr changes `jinjava.lib.filter.FromJsonFilter` to return a `Object` instead of a `HashMap` in order to also support deserializing JSON array strings [(see issue)]( https://github.com/HubSpot/jinjava/issues/457).

Now, we are able to use array JSON strings:
```
// template
{% test = "[one, two, three]" | fromjson %}
{{  test[0] }}

// output
one

/* Before this would raise an InvalidInputException */
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ notifications:
 before_install:
   - pip install --user codecov
 after_success:
-  - mvn validate -e
+  - travis_wait mvn validate -e
   - codecov
 
--- a/src/test/java/com/hubspot/jinjava/lib/filter/FromJsonFilterTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/filter/FromJsonFilterTest.java
@@ -1,120 +1,103 @@
 package com.hubspot.jinjava.lib.filter;
 
-import static java.util.Arrays.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableMap;
 import com.hubspot.jinjava.Jinjava;
 import com.hubspot.jinjava.interpret.InvalidInputException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
-import com.hubspot.jinjava.objects.SafeString;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
+import java.util.Map;
 import org.junit.Before;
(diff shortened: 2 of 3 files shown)


src/test/java/com/hubspot/jinjava/lib/filter/FromJsonFilterTest.java
```diff
@@ -46,21 +48,73 @@ public void itFailsWhenParameterIsNotString() {
...
-    checkedNestJson(node);
+    JsonNode node = (JsonNode) filter.filter(nestedJson, interpreter);
+
+    checkNestedJson(node);
   }
 
-  private void checkedNestJson(HashMap<String, Object> node) {
-    assertThat(node.get("first")).isEqualTo(Arrays.asList(1, 2, 3));
+  @Test
+  public void itReadsEmptyJsonArrayString() {
```
mattcoley, line 73: I think a better option for these tests is to render a jinjava template directly. See https://github.com/HubSpot/jinjava/blob/master/src/test/java/com/hubspot/jinjava/lib/filter/MinusTimeFilterTest.java#L37 as an example. That way we can make sure the API from Jinjava perspective is the same: for example: `{% set obj = test|fromjson %} {{ obj.property }}`

AlpriElse:
@mattcoley I reworked the testing to check rendered jinjava. Also, I removed the old tests because we were testing implementation and not behavior... which isn't ideal. Also, it looks like using `Object` instead of `JsonNode` gives us the intended functionality we want.

One thing I did notice that you should be aware of is the behavior of using the `pprint` filter with this:
```
// template
{% set obj = "{a: 100, b: [1, 2, 3, 4]}" %}
{{ obj | fromjson | pprint }}

// output on Design Manager:
(String: {a: 100, b: [1, 2, 3, 4]})

// output now:
(PyMap: {a=100, b=[1, 2, 3, 4]})
```

.travis.yml
```diff
@@ -21,6 +21,6 @@ notifications:
 before_install:
   - pip install --user codecov
 after_success:
-  - mvn validate -e
+  - travis_wait mvn validate -e
```
AlpriElse, line 24: I've been getting build failures because some packages take a while to install.
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

AlpriElse:
Build is succeeding in Blazar, going to go ahead with merge.

PR #461, opened 2020-06-30
Fix npe in list filter

boulter:
When a null list was passed to the ListFilter, it would throw an NPE. I fixed that and also cleaned up a few warnings along the way as I looked for other potential NPEs in filters.


src/main/java/com/hubspot/jinjava/lib/filter/ListFilter.java
```diff
@@ -37,6 +37,10 @@ public String getName() {
   public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
     List<?> result;
 
+    if (var == null) {
```
boulter, line 40: this is the actual fix

PR #462, opened 2020-07-06
avoid IndexOutOfBoundsExceptions

boulter:
Jinjava code could throw unchecked `IndexOutOfBoundsExceptions` when using indices that were out of range. This now throws an `IndexOutOfRangeException` which is captured in a `TemplateError` instead.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
@@ -29,7 +29,7 @@ public class PyList extends ForwardingList<Object> implements PyWrapper {
   }
 
   public void insert(int i, Object e) {
-    if (i >= list.size()) {
+    if (i < 0 || i >= list.size()) {
       throw createOutOfRangeException(i);
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
```diff
@@ -28,6 +29,10 @@ public boolean append(Object e) {
   }
 
   public void insert(int i, Object e) {
+    if (i >= list.size()) {
```
hs-lsong, line 32: Do you want to check if `i < 0`?

src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
```diff
@@ -28,6 +29,10 @@ public boolean append(Object e) {
   }
 
   public void insert(int i, Object e) {
+    if (i < 0 || i >= list.size()) {
```
Joeoh, line 32: Just checking that we don't support negative indexing as in Python? From testing in a Jinja interpreter this seems to be supported.

PR #463, opened 2020-07-09
log error on parent loop

kacperadach:
Still not really sure how this could happen but we've been encountering it. This code runs up the parent tree and logs an error if a parent loop is found.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ScopeMap.java
+++ b/src/main/java/com/hubspot/jinjava/util/ScopeMap.java
@@ -2,13 +2,19 @@ package com.hubspot.jinjava.util;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ScopeMap<K, V> implements Map<K, V> {
+  private static final Logger LOG = LoggerFactory.getLogger(ScopeMap.class);
+
   private final Map<K, V> scope;
   private final ScopeMap<K, V> parent;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/hubspot/jinjava/util/ScopeMap.java
```diff
@@ -19,6 +25,23 @@ public ScopeMap() {
   public ScopeMap(ScopeMap<K, V> parent) {
     this.scope = new HashMap<K, V>();
     this.parent = parent;
+
+    Set<ScopeMap<K, V>> parents = new HashSet<>();
```
boulter, line 29: I think you only need to check new parent against the existing parents, right? If we're running this every time a new ScopeMap is created, it will get them all.

kacperadach: So are you suggesting storing all the parents in a Set on the ScopeMap? Either way we have to run up the parent tree in the constructor.

src/main/java/com/hubspot/jinjava/util/ScopeMap.java
```diff
@@ -2,13 +2,19 @@
...
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ScopeMap<K, V> implements Map<K, V> {
+  private static final Logger LOG = LoggerFactory.getLogger(ScopeMap.class);
```
hs-lsong, line 16: I see we use `ENGINE_LOG` everywhere else in jinjava, should we keep using that?

kacperadach: Good call

src/main/java/com/hubspot/jinjava/util/ScopeMap.java
boulter: yeah, I was thinking just run up the whole parent tree in the constructor. I think that will take a trivial amount of time.

kacperadach: Should I add a config to enable/disable this check?

boulter: I can't think of a reason why you'd want to turn it off.

PR #472, opened 2020-08-08
Divide by zero

bszabo:
Fix IsDivisibleByExpTest to throw a meaningful exception when asked to check for divisibility by zero.
Add unit tests for the IsDIvisibleByExpTest class

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
@@ -74,8 +74,6 @@ public class IsDivisibleByExpTest implements ExpTest {
     }
 
     Number freeFormDivisor = (Number) args[0];
-    double floor = Math.floor(freeFormDivisor.doubleValue());
-    double ceiling = Math.ceil(freeFormDivisor.doubleValue());
     if (
       Math.floor(freeFormDivisor.doubleValue()) !=
       Math.ceil(freeFormDivisor.doubleValue())
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
@@ -24,7 +24,7 @@ import com.hubspot.jinjava.interpret.TemplateSyntaxException;
       "{% else %}\n" +
       "   <!--code to render if variable cannot be divided by 5-->\n" +
       "{% endif %}"
-    ),
+    )
   }
 )
 public class IsDivisibleByExpTest implements ExpTest {
@@ -42,7 +42,14 @@ public class IsDivisibleByExpTest implements ExpTest {
     if (!Number.class.isAssignableFrom(var.getClass())) {
       return false;
     }
-    int dividend = ((Number) var).intValue();
+    Number freeFormDividend = (Number) var;
+    if (
+      Math.ceil(freeFormDividend.doubleValue()) !=
+      Math.floor(freeFormDividend.doubleValue())
+    ) {
+      return false;
+    }
+    int dividend = freeFormDividend.intValue();
 
     if (args.length == 0) {
       throw new TemplateSyntaxException(
@@ -61,6 +68,22 @@ public class IsDivisibleByExpTest implements ExpTest {
         interpreter,
         this,
         InvalidReason.NUMBER_FORMAT,
+        0,
+        args[0].toString()
+      );
+    }
+
+    Number freeFormDivisor = (Number) args[0];
+    double floor = Math.floor(freeFormDivisor.doubleValue());
+    double ceiling = Math.ceil(freeFormDivisor.doubleValue());
+    if (
+      Math.floor(freeFormDivisor.doubleValue()) !=
+      Math.ceil(freeFormDivisor.doubleValue())
+    ) {
+      throw new InvalidArgumentException(
+        interpreter,
+        this,
+        InvalidReason.NON_ZERO_NUMBER,
         0,
         args[0].toString()
       );
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
@@ -19,7 +19,7 @@ import com.hubspot.jinjava.interpret.TemplateSyntaxException;
   ),
   snippets = {
     @JinjavaSnippet(
-      code = "{[[[[[[% if variable is divisibleby 5 %}]]]]]]\n" +
+      code = "% if variable is divisibleby 5 %\n" +
       "   <!--code to render if variable can be divided by 5-->\n" +
       "{% else %}\n" +
       "   <!--code to render if variable cannot be divided by 5-->\n" +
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
```diff
@@ -19,7 +19,7 @@
   ),
   snippets = {
     @JinjavaSnippet(
-      code = "{% if variable is divisbleby 5 %}\n" +
+      code = "% if variable is divisibleby 5 %\n" +
```
boulter, line 22: good catch on the typo here

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Test
+  public void itRequiresNonZeroDivisor() {
+    assertThatThrownBy(
+        () ->
+          jinjava.render(String.format(DIVISIBLE_BY_TEMPLATE, "10", "0"), new HashMap<>())
+      )
+      .hasMessageContaining("1st argument with value 0 must be non-zero");
+  }
```
boulter, line 77: Thanks for adding test coverage here. Can you check for the NUMBER_FORMAT exception as well?

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
```diff
@@ -0,0 +1,78 @@
...
+  private static final String DIVISIBLE_BY_TEMPLATE = "{{ %s is divisibleby %s }}";
+
+  private Jinjava jinjava;
+
+  @Before
+  public void setup() {
+    jinjava = new Jinjava();
+  }
+
+  @Test
```
boulter, line 21: Might be good to add at least one positive test case as well to confirm we didn't break anything.

src/main/java/com/hubspot/jinjava/lib/exptest/IsDivisibleByExpTest.java
boulter: I think we still need the { and } though.

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
```diff
@@ -0,0 +1,78 @@
...
+  public void itRequiresDivisor() {
+    assertEquals(
+      jinjava.render(String.format(DIVISIBLE_BY_TEMPLATE, "10", "null"), new HashMap<>()),
+      "false"
+    );
+  }
+
+  @Test
+  public void itRequiresNumericalDivisor() {
+    // Question for PR: It seems that InvalidArgumentExceptions are caught before my middleware. How to know it failed for the right reason?
```
boulter, line 50: you can look at the errors on the interpreter by using `jinjava.renderForResult`. [Example](https://github.com/HubSpot/jinjava/blob/c2d7df73cc9b91c90ee0ea4c89f85dcb3abbe532/src/test/java/com/hubspot/jinjava/lib/filter/SortFilterTest.java#L92)

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
bszabo: I'll break out the one positive test on line 59 (currently embedded in a negative test) to be its own test

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
bszabo: I don't know how to get the jinjava interpreter to take input provided as a string template into an object whose type falls outside the Number inheritance tree. That is, I don't know how to structure the test call to induce this condition.

In the process of trying I found that a template of  { 123.4 is divisibleby 3} returns "true". I would have expected a return value of false. This seems like a logic error in jinjava, though fixing it may well prove to be a breaking change, in case any client has assumed the opposite.

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
bszabo: Yes. I followed your suggestion and this addresses my need.

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
```diff
@@ -46,21 +42,15 @@ public void itRequiresDivisor() {
...
-  public void itRequiresDividendIsMultipleOfDivisor() {
+  public void itPassesWhenDividendIsDivisibleByDivisor() {
     assertEquals(
       jinjava.render(String.format(DIVISIBLE_BY_TEMPLATE, "10", "5"), new HashMap<>()),
       "true"
     );
+  }
 
+  @Test
+  public void itFailsWhenDividendIsNotMultipleOfDivisor() {
```
boulter, line 53: nit: doesn't really fail, just returns false. It works correctly!

src/test/java/com/hubspot/jinjava/lib/exptest/isDivisibleByExpTestTest.java
boulter: not sure what you mean in the first paragraph. What have you tried?

As for the second one, I think it's worth fixing.

bszabo: I'm fixing the non-integer case to return false if it's the dividend that's fractional, and to throw an `InvalidArgumentException` if it's the divisor that's fractional. (Testing for fractional nature by comparing Math.ceil() with Math.floor() for equality).

For the first paragraph, what I mean is that all my tests invoke the system under test by passing down a string like "{10 is divisibleby 5}" or "{10 is divisibleby null}". I don't know what I might pass just before the closing "}" for a test case that violates the NUMBER_FORMAT requirement. I don't think that calling renderForResult() instead of render() in the test would buy me anything. It would appear that testing this condition requires mocking some internal jinjava interpreter class. What am I missing?

boulter: Can you try `10 is divisibleby 'foo'`?

boulter:
Code looks good, but the build is failing

[INFO] Dead store to ceiling in com.hubspot.jinjava.lib.exptest.IsDivisibleByExpTest.evaluate(Object, JinjavaInterpreter, Object[]) [com.hubspot.jinjava.lib.exptest.IsDivisibleByExpTest] At IsDivisibleByExpTest.java:[line 78] DLS_DEAD_LOCAL_STORE
[INFO] Dead store to floor in com.hubspot.jinjava.lib.exptest.IsDivisibleByExpTest.evaluate(Object, JinjavaInterpreter, Object[]) [com.hubspot.jinjava.lib.exptest.IsDivisibleByExpTest] At IsDivisibleByExpTest.java:[line 77] DLS_DEAD_LOCAL_STORE

bszabo:
"Is divisible by" functionality doesn't make sense for either fractional numbers or for zero divisor. This change catches these conditions and adds unit testing to "is divisible by"

PR #473, opened 2020-08-13
Divide by zero

bszabo:
Handle division operations within expressions via extended operation operator

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
@@ -18,16 +18,9 @@ public class RegularDivOperator extends SimpleOperator {
       throw new IllegalArgumentException("Division argument may not be null");
     }
 
-    // Debugging support only -- Delete me
-    //    Type typeA = a.getClass();
-    //    String typeNameA = typeA.getTypeName();
-    //    Type typeB = b.getClass();
-    //    String typeNameB = typeB.getTypeName();
-    // End of debugging support
-
     Number numA = (Number) a;
     Number numB = (Number) b;
-    if (numB.longValue() == 0 || numB.doubleValue() == 0.0) {
+    if (numB.doubleValue() == 0.0) {
       throw new IllegalArgumentException("Divisor may not be zero");
     }
 
--- a/src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
@@ -1,0 +1,46 @@
+package com.hubspot.jinjava.el.ext;
+
+import de.odysseus.el.misc.TypeConverter;
+import de.odysseus.el.tree.impl.Parser.ExtensionHandler;
+import de.odysseus.el.tree.impl.Parser.ExtensionPoint;
+import de.odysseus.el.tree.impl.Scanner;
+import de.odysseus.el.tree.impl.ast.AstBinary;
+import de.odysseus.el.tree.impl.ast.AstBinary.SimpleOperator;
+import de.odysseus.el.tree.impl.ast.AstNode;
+
+public class RegularDivOperator extends SimpleOperator {
+  public static final Scanner.ExtensionToken TOKEN = new Scanner.ExtensionToken("/");
+  public static final RegularDivOperator OP = new RegularDivOperator();
+
+  @Override
+  protected Object apply(TypeConverter converter, Object a, Object b) {
+    if (a == null || b == null) {
+      throw new IllegalArgumentException("Division argument may not be null");
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
```diff
@@ -0,0 +1,46 @@
...
+  public static final Scanner.ExtensionToken TOKEN = new Scanner.ExtensionToken("/");
+  public static final RegularDivOperator OP = new RegularDivOperator();
+
+  @Override
+  protected Object apply(TypeConverter converter, Object a, Object b) {
+    if (a == null || b == null) {
+      throw new IllegalArgumentException("Division argument may not be null");
+    }
+
+    // Debugging support only -- Delete me
```
boulter, line 21: TODO: don't forget to delete!

src/test/java/com/hubspot/jinjava/el/ext/RegularDivTest.java
```diff
@@ -0,0 +1,27 @@
...
+public class RegularDivTest {
+  private Jinjava jinja;
+
+  @Before
+  public void setUp() {
+    jinja = new Jinjava();
+  }
+
+  @Test
+  public void okOnLongDivisionWithNonZeroDivisor() {
```
boulter, line 19: we favor rspec-style test names like `itDividesLongsWithNonZeroDivisor`

src/test/java/com/hubspot/jinjava/el/ext/RegularDivTest.java
```diff
@@ -0,0 +1,27 @@
...
+    assertEquals(jinja.render("{% set x = 10 / 2%}{{x}}", new HashMap<>()), "5.0");
```
boulter, line 20: Does the normal division operator return division of longs with a decimal point?

src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
```diff
@@ -0,0 +1,46 @@
...
+    //    Type typeA = a.getClass();
+    //    String typeNameA = typeA.getTypeName();
+    //    Type typeB = b.getClass();
+    //    String typeNameB = typeB.getTypeName();
+    // End of debugging support
+
+    Number numA = (Number) a;
+    Number numB = (Number) b;
+    if (numB.longValue() == 0 || numB.doubleValue() == 0.0) {
```
boulter, line 30: don't you only need to check the doubleValue?

src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
bszabo: Deleted

bszabo: I did have a question earlier on, when I was assuming that` RegularDivOperator ` should return a long when dividing long arguments and that it should return a double when dividing double or mixed arguments. When I tried coding `RegularDivOperator` that way I got a regression failure on the `ExpressionResolverTest.itResolvesMathVals()`, which was expecting a result of "5.0" when dividing 15L by 3L. Switching to having `RegularDivOperator` always return a double made the question about how to simply and reliably format the result vanish. It also fixed the regression and yielded a clean build.

src/test/java/com/hubspot/jinjava/el/ext/RegularDivTest.java
bszabo: (Cut-and-pasted from my reply to "Did you have a question ..." comment above). When I tried coding RegularDivOperator to return a long on division of long arguments I got a regression failure on the ExpressionResolverTest.itResolvesMathVals(), which was expecting a result of "5.0" when dividing 15L by 3L. Switching to having RegularDivOperator always return a double fixed the regression and yielded a clean build.

src/main/java/com/hubspot/jinjava/el/ext/RegularDivOperator.java
```diff
@@ -0,0 +1,39 @@
...
+
+import de.odysseus.el.misc.TypeConverter;
+import de.odysseus.el.tree.impl.Parser.ExtensionHandler;
+import de.odysseus.el.tree.impl.Parser.ExtensionPoint;
+import de.odysseus.el.tree.impl.Scanner;
+import de.odysseus.el.tree.impl.ast.AstBinary;
+import de.odysseus.el.tree.impl.ast.AstBinary.SimpleOperator;
+import de.odysseus.el.tree.impl.ast.AstNode;
+
+public class RegularDivOperator extends SimpleOperator {
```
boulter, line 11: I think this could simply be called `DivOperator`.

Also I think it would be worth a comment on why this was overridden.

PR #474, opened 2020-08-14
disallow adding a map to itself

boulter:
Since serializing a map with references to itself is disasterous, let's try to prevent that when adding things to the map.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ScopeMap.java
+++ b/src/main/java/com/hubspot/jinjava/util/ScopeMap.java
@@ -11,6 +11,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ScopeMap.java
```diff
@@ -120,7 +126,18 @@ public V remove(Object key) {
   }
 
   @Override
-  public void putAll(Map<? extends K, ? extends V> m) {
+  public void putAll(@Nonnull Map<? extends K, ? extends V> m) {
+    for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
```
hs-lsong, line 130: Since you are iterating the entries, can you just call `put(key, value)` for each of the entry, and skip the `scope.putAll(m)`?

boulter: The `HashMap` implementation of `putAll` has some optimizations for inserting multiple entries, so I didn't want to override that.

PR #480, opened 2020-08-17
Revert "Revert "Divide by zero"" (New PR)

phalaphone:
Reverts HubSpot/jinjava#479 to behave as the new PR for Divide by Zero


src/main/java/com/hubspot/jinjava/el/ext/DivOperator.java
```diff
@@ -0,0 +1,43 @@
...
+ * (see PR 473 @ https://github.com/HubSpot/jinjava/pull/473)
+ */
+public class DivOperator extends SimpleOperator {
+  public static final Scanner.ExtensionToken TOKEN = new Scanner.ExtensionToken("/");
+  public static final DivOperator OP = new DivOperator();
+
+  @Override
+  protected Object apply(TypeConverter converter, Object a, Object b) {
+    if (a == null || b == null) {
+      throw new IllegalArgumentException("Division argument may not be null");
```
boulter, line 22: @bszabo, we should be throwing an `InvalidArgumentException` here instead of `IllegalArgumentException`. Sorry for not catching that earlier.

bszabo: I've made the change both here and lower down in the file, where detection of a zero divisor occurs

PR #484, opened 2020-08-20
Detect null pointer expression when evaluating array expression with …

bszabo:
In progress pull request for ability to detect array expressions missing an index. Looking for feedback as to whether detection/reporting could be improved upon.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -75,7 +75,26 @@ public class ExpressionResolver {
         elExpression,
         Object.class
       );
-      Object result = valueExp.getValue(elContext);
+      Object result = null;
+      try {
+        result = valueExp.getValue(elContext);
+      } catch (NullPointerException e) {
+        interpreter.addError(
+          new TemplateError(
+            ErrorType.FATAL,
+            ErrorReason.EXCEPTION,
+            ErrorItem.PROPERTY,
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
```diff
@@ -75,7 +75,26 @@ public Object resolveExpression(String expression) {
         elExpression,
         Object.class
       );
-      Object result = valueExp.getValue(elContext);
+      Object result = null;
+      try {
+        result = valueExp.getValue(elContext);
+      } catch (NullPointerException e) {
```
boulter, line 81: wouldn't this catch any NullPointerException in an expression, not just those from arrays? I think we may need to find a place that's bit lower to catch this.

bszabo: I'm closing this effort due to lack of progress on this front. I tried using a technique akin to the override for `get(int index)` in `PyList.java` but this does not work on two counts: 1) the call happens to late relative to when the exception is trhown; and 2) the method signature does not work with a null index.

bszabo: The second effort had to do with detecting a node referencing a left bracket and null property. This is discussed towards the bottom of this slack thread: https://app.slack.com/client/T024G0P55/CSWBJ5UH1/thread/CSWBJ5UH1-1597185380.177200

PR #489, opened 2020-08-24
- Prevent adding PyMap to itself; add unit tests

bszabo:


src/test/java/com/hubspot/jinjava/objects/collections/PyMapTest.java
```diff
@@ -0,0 +1,262 @@
...
+  public void setup() {
+    jinjava = new Jinjava();
+  }
+
+  @Test
+  public void itSupportsAppendOperation() {
+    RenderResult result = jinjava.renderForResult(
+      "{% set mymap = {} %} {% do mymap.put('map', mymap) %} {{ mymap }}",
+      Collections.emptyMap()
+    );
```
boulter, line 29: this doesn't assert anything. What should it do?

bszabo: I was using this test for debugging purposes and forgot to reinstate it to its proper form. Thanks for catching this.

PR #490, opened 2020-08-25
Use kwargs in truncatehtml filter

liamrharwood:
The [`truncatehtml`](https://developers.hubspot.com/docs/cms/hubl/filters#truncatehtml) filter has two optional arguments, `end` and `breakwords`. The `end` argument is a string which specifies the trailing characters after truncation, while `breakwords` determines if the last word of text is discarded.

Currently, these two optional arguments are recognized purely by their position in the array. This means that the following code:
```
{% set html_text = "<p>I want to truncate this text without breaking my HTML<p>" %} 
{{ html_text|truncatehtml(28, false) }}
```
Renders this:
```
<p>I want to truncate thisfalse</p>
```

My expectation would be that the `false` passed to the filter should be recognized as `breakwords` -- since the two arguments are different types -- but instead it is used as `end`.

This PR changes this behavior so that if there is only one optional argument present, the filter will deduce which one it is based on the type. First we try and parse it as a boolean, and if it is not a recognized boolean value, we consider it to be `end`. If it is a recognized boolean value, we consider it to be `breakwords`.

If we don't like this new behavior, we should instead update the docs so that the current behavior is more clear.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
@@ -7,6 +7,7 @@ import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.lib.fn.Functions;
+import java.util.Map;
 import java.util.Objects;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


mattcoley:
How about we extend `AdvancedFilter` instead and support kwargs in the filter? That way you could `{{ html_text|truncatehtml(28, breakwords=false) }}`

liamrharwood:
Ah, that's a good point -- I'll try that instead.

src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
```diff
@@ -55,6 +56,47 @@ public String getName() {
...
+    JinjavaInterpreter interpreter,
+    Object[] args,
+    Map<String, Object> kwargs
+  ) {
+    String length = null;
+    if (kwargs.containsKey(("length"))) {
+      length = Objects.toString(kwargs.get("length"));
+    }
+    String end = null;
+    if (kwargs.containsKey("end")) {
```
liamrharwood, line 71: Changed this to use kwargs - there's probably a more generic way of doing this, but I figure that can be tackled when trying to bring this functionality to our other filters.

src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
```diff
@@ -55,6 +56,47 @@ public String getName() {
...
+      end = Objects.toString(kwargs.get("end"));
+    }
+    String breakword = null;
+    if (kwargs.containsKey("breakword")) {
```
mattcoley, line 75: nit: Can these strings be static?

PR #503, opened 2020-09-08
Pass function mapper to new interpreter

Joeoh:
There was a problem where if we created a new interpreter eg. to set the config

```
         renderContext.setAutoEscape(true);
        JinjavaInterpreter interpreter = new JinjavaInterpreter(
          jinjava,
          renderContext,
          jinjava.getGlobalConfig()
        );

```

The function resolver would not be copied from the parent to child interpreter so all custom/built in functions would be unavailable. This was because JinjavaInterpreter.getCurrent() would return null from within MacroFunctionMapper

I now pass the interpreter down when creating the JinjavaELContext-> MacroFunctionMapper


Joeoh:
Internal build is fine for this. Is there something weird about the locale env in the public Travis @mattcoley ?

Joeoh:
Can I get some 👀  please?

boulter (approved):

PR #507, opened 2020-09-17
Issue 492: For loop to iterate over keys of dictionary

AnnaKhismatullina:
Make possible to configure iteration mode for dictionaries

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ObjectIterator.java
+++ b/src/main/java/com/hubspot/jinjava/util/ObjectIterator.java
@@ -16,6 +16,7 @@
 package com.hubspot.jinjava.util;
 
 import com.google.common.collect.Iterators;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/util/ObjectIterator.java
```diff
@@ -55,4 +50,16 @@ public static ForLoop getLoop(Object obj) {
     // others
     return new ForLoop(Iterators.singletonIterator(obj), 1);
   }
+
+  @SuppressWarnings("unchecked")
+  public static ForLoop getLoop(Object obj, boolean iterateOverMapKeys) {
```
boulter, line 55: instead of passing this down through all these methods, could you just use the static `JinjavaInterpreter.getCurrent()` to get a hold of the interpreter and the config?

src/test/java/com/hubspot/jinjava/util/ObjectIteratorTest.java
```diff
@@ -73,20 +74,33 @@ public void test6() {
...
     items.add("asfun");
-    loop = ObjectIterator.getLoop(items);
+    loop = ObjectIterator.getLoop(items, false);
     assertEquals(4, loop.getLength());
     assertEquals("hello", loop.next());
   }
 
   @Test
-  public void test7() {
+  public void testItIteratesOverValues() {
```
boulter, line 83: much better name!

boulter (changes requested):
Thanks for doing this. I have a suggestion for making this simpler.

boulter (approved):
This is much simpler. Thanks!

PR #512, opened 2020-09-28
Accept ip address without network prefix in ipaddr('address') filter

Jolmberg:
In jinja2 with netaddr, the ipaddr('address') filter accepts a plain ip address, like "192.168.0.1". Jinjava currently rejects this because it has no network prefix. This patch fixes the jinjava behaviour for these cases.


Jolmberg:
The failing test case is DateTimeFormatFilterTest.itDefaultsToEnglishForBadLocaleValues, and it fails in a clean master too.

mattcoley (approved):

mattcoley:
Hmm that is because the build is targeting Java 8 vs. 11.

PR #519, opened 2020-10-16
Eager Execution for deferred values

jasmith-hs:
I'm working on a way to allow nodes to continue to render when `DeferredValue`'s are encountered. I am calling this approach eager execution because the principle is similar to [Eager Execution](https://en.wikipedia.org/wiki/Speculative_execution#Eager_execution), a type of Speculative Execution.
The goal is to pre-render as many nodes as possible, and only defer the nodes which depend on a DeferredValue, while evaluating the children if possible.

For pre-rendering usages it may be helpful to know which types of nodes were deferred. Without eager execution, if a `TagNode` such as a `ForTag` or `IfTag` gets deferred, there may be nodes that cannot be evaluated during a second render (such as custom widgets in HubSpot). Therefore a second render can not be performed, and the pre-rendered jinjava template must be scrapped. However, with eager execution, custom widgets could be eagerly evaluated inside of `if` statements or `for` loops, so long as they don't directly depend on a deferred value, which would allow the pre-rendered version to be interpreted with a second pass.

Code is pretty WIP, but wanted to get a concept going to figure out if this idea is feasible, and get some feedback on the concept.

---
A couple quick examples to demonstrate the concept:
```
{% if contact.registered %}
{% custom thing %}
{% else %}
{% other custom thing %}
{% endif %}
```
This is a simple case, where eagerly executing the `if` statement (assuming `contact` is deferred) would cause the output to be:
```
{% if contact.registered %}
<video HTML>
{% else %}
<registration form HTML>
{% endif %}
```
A second pass with the `contact` set would either give the video or the form, without needing to have those widgets info.

---
```
{% for item in deferred %}
{{ item }}
{% custom thing %}
{% endfor %}
```
If `{% custom thing %}` doesn't depend on the deferred value, with eager execution it can be pre-rendered like:
```
{% for item in deferred %}
{{ item }}
result
{% endfor %}
```
Of course, if `{% custom thing %}` does depend on the deferred value, it won't get evaluated.

---
Also with `resolved = ['piece', 'chunk', 'bite']`:
```
{% for item in resolved %}
A {{ item }} of {{ deferred }}.
{% endfor %}
```
could get pre-rendered to:
```
A piece of {{ deferred }}.
A chunk of {{ deferred }}.
A bite of {{ deferred }}.
```


boulter:
The concept make sense to me. I didn't realize we we immediately deferring all children of a node where a deferred value was encountered. Rendering as much as possible in the first pass that doesn't depend on deferred value seems like a big win.

Joeoh:
I'm going to look at this in more detail but one thing I wanted to flag:
We originally had similar behaviour to what you've implemented for for loops. I needed to change it to handle cases where we set use the loop variable against a deferred value inside the for loop. For example:


```
items = ['item1', 'item2', 'item3']:

{% for item in items %}
    {{ item }} {{ deferred }}.
   {% set temp = item ~ deferred %}
   {{ temp }}
{% endfor %}
```


First render would result in this:

```
    item1 {{ deferred }}.
    {% set temp = item ~ deferred %}
    {{ temp }}
    item2 {{ deferred }}.
    {% set temp = item ~ deferred %}
    {{ temp }}
    item3 {{ deferred }}.
    {% set temp = item ~ deferred %}
    {{ temp }}

```

 
By the current behaviour, item will be marked as deferred and stored in the context. But we will just store the last value it had so "item3"
When rendering again with `deferred = "resolved"` we will end up with this:

```
    item1 resolved.
    item3resolved
    item2 resolved.
    item3resolved
    item3 resolved.
    item3resolved

```

The desired output would be:
```
    item1 resolved.
    item1resolved
    item2 resolved.
    item2resolved
    item3 resolved.
    item3resolved
```



Heres a PR I did to change this behaviour with another explanation:
https://github.com/HubSpot/jinjava/pull/467

jasmith-hs:
@Joeoh Thanks for mentioning that. I had found that PR earlier so I have that case covered. I have been working on this on a separate branch `eager-execution-a`. And I have accounted for such circumstances by introducing a [Chunk Resolver](https://github.com/jasmith-hs/jinjava/blob/eager-execution-a/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java). (I will open a PR with just the chunk resolver code soon).

> [The Chunk Resolver] takes a string and resolves it in chunks. This allows for strings with deferred values within them to be partially resolved, as much as they can be with a deferred value.

It evaluates the value of `item` at each step in the for loop so that it can expand the children.

Here's a test showing that it handles that scenario: https://github.com/jasmith-hs/jinjava/commit/9d26b06584de0d32b02802dd509638e399463fc5

jasmith-hs:
Closing in favor of using this issue: https://github.com/HubSpot/jinjava/issues/532 as the point of contact

PR #522, opened 2020-10-28
Add abstract base test classes

jasmith-hs:
I found that there's a lot of code duplication in most of the tests, whether it be setting up a `Jinjava` object or `JinjavaInterpreter` instance variable in the `@Before setup()` method.

This PR adds 2 abstract classes: `BaseJinjavaTest` and `BaseInterpretingTest`, the latter of which extends the former.
These base test classes help remove a lot of duplicate logic from the setup of the test classes.


boulter (approved):
nice!

PR #523, opened 2020-10-30
Python Booleans and Filter base with parsing

michaelpro1:
This change includes the following

- Support of python style booleans for portability of templates - `True` and `False` are added into `ExtendedParser`
- `AbstractFilter` that can be used for simpler filters which works by introspecting `Jinjavadoc.params` section with some automatic conversions


boulter:
Thanks for the contribution. Could you break out the YAML filters into their own PR?

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+    throw new InvalidInputException(
+            interpreter,
+            "INVALID_ARG_NAME",
+            String.format("Argument named '%s' with value '%s' cannot be parsed for filter %s", jinjavaParamMetadata.value(), getName())
+    );
+  }
+
+  public void validateArgs(JinjavaInterpreter interpreter, Map<String, Object> parsedArgs) {
+    for(JinjavaParam jinjavaParam: getNamedArguments().values()) {
```
boulter, line 144: style nit:
```suggestion
    for (JinjavaParam jinjavaParam : getNamedArguments().values()) {
```

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+      if(jinjavaParam.required() && !parsedArgs.containsKey(jinjavaParam.value())) {
```
boulter, line 145:
```suggestion
      if (jinjavaParam.required() && !parsedArgs.containsKey(jinjavaParam.value())) {
```

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  public Map<String, JinjavaParam> getNamedArguments() {
+    if (isNamedArgumentCheckDone) {
+      return namedArguments;
+    }
+    JinjavaDoc jinjavaDoc = this.getClass().getAnnotation(JinjavaDoc.class);
+    if (jinjavaDoc != null) {
+      namedArguments = new LinkedHashMap<>();
+      for (JinjavaParam jinjavaParam: jinjavaDoc.params()) {
```
boulter, line 162:
```suggestion
      for (JinjavaParam jinjavaParam : jinjavaDoc.params()) {
```

src/main/java/com/hubspot/jinjava/lib/filter/IndentFilter.java
```diff
@@ -40,24 +41,20 @@
...
   @Override
-  public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
-    int width = 4;
-    if (args.length > 0) {
-      width = NumberUtils.toInt(args[0], 4);
-    }
+  public Object filter(Object var,
+                       JinjavaInterpreter interpreter,
+                       Map<String, Object> parsedArgs) {
+    int width = ((Number)parsedArgs.get("width")).intValue();
```
boulter, line 55:
```suggestion
    int width = ((Number) parsedArgs.get("width")).intValue();
```

src/main/java/com/hubspot/jinjava/lib/filter/IndentFilter.java
```diff
@@ -40,24 +41,20 @@
...
 
-    boolean indentFirst = false;
-    if (args.length > 1) {
-      indentFirst = BooleanUtils.toBoolean(args[1]);
-    }
+    boolean indentFirst = (boolean)parsedArgs.get("indentfirst");
```
boulter, line 57:
```suggestion
    boolean indentFirst = (boolean) parsedArgs.get("indentfirst");
```

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+        namedArguments.put(jinjavaParam.value(), jinjavaParam);
+      }
+      namedArguments = Collections.unmodifiableMap(namedArguments);
+    }
+    isNamedArgumentCheckDone = true;
+    return namedArguments;
+  }
+
+  public int getNamedArgumentPosition(String argName) {
+    getNamedArguments();
```
boulter, line 172: You could rewrite this in funcional style with something like

`return Optional.ofNullable(getNamedArguments()).map(a -> a.keySet()).filter(argNames -> a.contains(argName)).map(a -> a.indexOf(argName)).orElse(-1);`

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+    if (namedArguments != null) {
+      List<String> argNames = new ArrayList<>(namedArguments.keySet());
+      return argNames.contains(argName) ? argNames.indexOf(argName) : -1;
+    }
+    return -1;
+  }
+
+  public String getIndexedArgumentName(int position) {
+    getNamedArguments();
```
boulter, line 181: same comment as above.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,189 @@
...
+import org.apache.commons.lang3.math.NumberUtils;
+
+/***
+ * Filter base that uses Filter Jinjavadoc to construct named argument parameters.
+ * Only filters that specify name, type and defaults correctly should use this as a base
+ *
+ * @see JinjavaDoc
+ * @see JinjavaParam
+ */
+public abstract class AbstractFilter implements Filter {
```
boulter, line 44: Can you add some more test coverage for this class?

boulter (changes requested):
Thanks!

michaelpro1:
Take a look at this, I'll open PR for YAML after this

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,225 @@
...
+  private final Map<String, JinjavaParam> NAMED_ARGUMENTS;
```
boulter, line 44:
```suggestion
  private final Map<String, JinjavaParam> namedArguments;
```
because it's not static

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,225 @@
...
+  ) {
+    Map<String, Object> namedArgs = new HashMap<>();
+    //Set defaults
+    NAMED_ARGUMENTS.forEach(
+      (k, v) -> {
+        if (StringUtils.isNotEmpty(v.defaultValue())) {
+          namedArgs.put(k, v.defaultValue());
+        }
+      }
+    );
```
boulter, line 74: Might as well do this in the initializer with a `stream()` and `collect()`.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,225 @@
...
+        );
+      }
+      namedArgs.put(argName, arg);
+    }
+
+    //Parse args based on their declared types
+    Map<String, Object> parsedArgs = new HashMap<>();
+    namedArgs.forEach(
+      (k, v) -> parsedArgs.put(k, parseArg(interpreter, NAMED_ARGUMENTS.get(k), v))
+    );
```
boulter, line 112: Similar comment as above.

src/main/java/com/hubspot/jinjava/lib/filter/IndentFilter.java
```diff
@@ -40,24 +39,22 @@
...
+    JinjavaInterpreter interpreter,
+    Map<String, Object> parsedArgs
+  ) {
+    int width = ((Number) parsedArgs.get("width")).intValue();
 
-    boolean indentFirst = false;
-    if (args.length > 1) {
-      indentFirst = BooleanUtils.toBoolean(args[1]);
-    }
+    boolean indentFirst = (boolean) parsedArgs.get("indentfirst");
```
boulter, line 57: could you create a static constant for "indentfirst" and "width"?

src/test/java/com/hubspot/jinjava/lib/filter/AbstractFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import java.math.BigDecimal;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Test;
+
+public class AbstractFilterTest extends BaseInterpretingTest {
```
boulter, line 17: nice tests, thank you.

src/test/java/com/hubspot/jinjava/lib/filter/AbstractFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+      @JinjavaParam(value = "dict", type = "dict", desc = "dict")
+    }
+  )
+  public static class AllParamTypesFilter extends ArgCapturingFilter {}
+
+  @Test
+  public void itParsesNumericAndBooleanInput() {
+    filter = new AllParamTypesFilter();
+
+    Map<String, Object> kwArgs = new HashMap<>();
```
boulter, line 77: You could use `ImmutableMap` here for a more fluent interface and less repetition.

boulter (changes requested):
Looks pretty good. I have a few small comments. 

It is a pretty low-level change though so we will have to be careful about rolling this out.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
michaelpro1: Done

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
michaelpro1: Done

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
michaelpro1: These are runtime arguments so can't be initialized at init

src/main/java/com/hubspot/jinjava/lib/filter/IndentFilter.java
michaelpro1: Done

src/test/java/com/hubspot/jinjava/lib/filter/AbstractFilterTest.java
michaelpro1: Unfortunately `ImmutableMap.of` has a limit of 5 keys so forced to use this

michaelpro1:
> 
> 
> Looks pretty good. I have a few small comments.
> 
> It is a pretty low-level change though so we will have to be careful about rolling this out.

Absolutely, if you have a perf test to compare before/after that would be good

src/test/java/com/hubspot/jinjava/lib/filter/AbstractFilterTest.java
boulter: You could use `ImmutableMap.builder()`.

boulter:
> Absolutely, if you have a perf test to compare before/after that would be good

I'm less concerned about performance than compatibility with existing templates and unknowns. We can see how it goes.

boulter:
> FYI prettier-java errors out on my machine with posix error, could you do that once you are happy with the code?

Prettier reported no changes for me.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,238 @@
+/**********************************************************************
+ Copyright (c) 2014 HubSpot Inc.
```
boulter, line 2:
```suggestion
 Copyright (c) 2020 HubSpot Inc.
```

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  public Map<String, JinjavaParam> initNamedArguments() {
+    JinjavaDoc jinjavaDoc = this.getClass().getAnnotation(JinjavaDoc.class);
+    if (jinjavaDoc != null) {
+      Map<String, JinjavaParam> namedArgs = new LinkedHashMap<>();
+      for (JinjavaParam jinjavaParam : jinjavaDoc.params()) {
+        namedArgs.put(jinjavaParam.value(), jinjavaParam);
+      }
+      return Collections.unmodifiableMap(namedArgs);
```
boulter, line 218:
```suggestion
return jinjavaDoc.params().stream().collect(ImmutableMap.toImmutableMap(p -> p.value(), p -> p));
```

boulter: Note that this won't preserve the order though. If you do need that, you could collect to a LinkedHashMap and then wrap the whole thing in `Collections.unmodifiableMap`

boulter (changes requested):
just a few more nits. Thanks again.

src/test/java/com/hubspot/jinjava/lib/filter/AbstractFilterTest.java
michaelpro1: Done

src/main/java/com/hubspot/jinjava/lib/filter/AbstractFilter.java
michaelpro1: The order is important, just to be sure I used forEachOrdered with Builder

michaelpro1:
> Prettier reported no changes for me.

Doesn't seem to work on Windows but OK with WSL. NodeJS executable is an odd dependency.

> I'm less concerned about performance than compatibility with existing templates and unknowns. We can see how it goes.

I've implemented the `.filter(String[])`  method as well which should behave as before and how most filters retrieve their params. 

As this change is only applied to 2 filters, the rest can be modified when needed. e.g. Filters with single param will work without an issue, however where >1 params are involved further testing is warranted.

boulter (approved):

boulter:
@mattcoley any concerns?

mattcoley (approved):
🆒  LGTM

PR #527, opened 2020-11-06
Implement ChunkResolver by extending ExtendedParser

jasmith-hs:
@jboulter Because of your comment about the parser that exists in Jinjava, I decided to look into it as an alternative to the ChunkResolver PR that I opened https://github.com/HubSpot/jinjava/pull/525.
It essentially does the same thing that the ChunkResolver does but in a more low-level (and probably faster) way.

Rather than introducing a new parser that works on top of the `interpreter.resolveELExpression()`, this approach instead extends the Jinjava parser itself by providing decorators for all the different types of AstNodes to allow for deferred values to be handled when they are evaluated.

I created a new Exception type that passes through the Abstract Syntax Tree as it gets evaluated, and when one of the decorated nodes sees this exception, it partially evaluates its children (or whatever they are specifically for each type of AstNode).

This method for partially resolving expressions is more complex than the initial ChunkResolver PR because I had to implement custom overrides for all the different AstNodes.

This is a draft PR because idk if I want to go with this approach. It's a lot more complex than the ChunkResolver, and I haven't fully tested out all the code (I only have it tested against the tests that I wrote for the ChunkResolver PR). A pro to this approach is that I think it is faster than the other PR because the expression gets evaluated a single time, rather than being evaluated several times in sub-expressions.

I don't intend for anyone to read through this code, it's currently just a proof of concept, unless we decide to go with this approach.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/Jinjava.java
+++ b/src/main/java/com/hubspot/jinjava/Jinjava.java
@@ -19,6 +19,7 @@ import com.hubspot.jinjava.doc.JinjavaDoc;
 import com.hubspot.jinjava.doc.JinjavaDocFactory;
 import com.hubspot.jinjava.el.ExtendedSyntaxBuilder;
 import com.hubspot.jinjava.el.TruthyTypeConverter;
+import com.hubspot.jinjava.el.ext.eager.EagerExtendedSyntaxBuilder;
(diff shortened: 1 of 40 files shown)


boulter:
Thanks for trying this out. It's a bummer that it would require overriding all those classes. I don't think performance is a big concern here so I guess the original approach makes more sense.

PR #528, opened 2020-11-10
Simply JinjavaConfig construction

jasmith-hs:
The JinjavaConfig constructor had too many parameters. This replaces that constructor with one that just takes in a `Builder`. It is both easier to read and it doesn't violate checkstyle (which throws a style violation after 20+ parameters).

Also, many of the values in that constructor are the default values in the Builder class so they can be omitted, which removes a fair bit of duplicate code.


boulter (approved):
This is great. thanks.

PR #529, opened 2020-11-10
Introduce Eager Tokens

jasmith-hs:
---
For the Eager Execution project (see https://github.com/HubSpot/jinjava/pull/519), the variables, functions, and filters (together referred to as "words") need to be marked so it is known which words need to be resolved/evaluated for a second rendering pass. This PR sets up this concept, and the actual eager execution functionality for different tags will come in later PRs.

The `eagerExecutionEnabled` flag allows for the option to have different interpreters to be specified with a simple switch to do eager execution on one of them (or neither of them). This would generally only be used when doing a multiple render approach with an initial pre-render and then a final render pass. The pre-render interpreter would have `eagerExecutionEnabled=true`, and the final render interpreter would have it set to false.

The Eager Execution project executes all the children of a node that would be deferred due to the encountering of a deferred value in the context _eagerly_ (see https://en.wikipedia.org/wiki/Speculative_execution#Eager_execution for where that term is derived). By executing all the children nodes, as well as everything in an expression that needs to be deferred (see the [ChunkResolver](https://github.com/HubSpot/jinjava/pull/525) for partial evaluation of deferred expressions), we can determine which words were not able to be resolved/evaluated in the initial pre-render pass. These words will be stored in the `EagerToken` structure.

An `EagerToken` consists of __three__ parts:

1. `token`: this is the `tree.parse.Token` that gets "reconstructed" and put into the output. It is used to determine which types of tokens are deferred (be it a `SetTag`'s token, an `ExpressionNode`'s token, etc.), as some types of tokens may not be possible to evaluate in a second render. It is useful to know ahead of time which tokens will still need to be evaluated in a second render. (If there are none, then a second render isn't required).
2. `setDeferredWords`: this is a set of strings that are for when a word (namely a variable) gets set within the token and the setting needs to be deferred. This can happen in a ForTag, CycleTag, FromTag, ImportTag, and a SetTag.
3. `usedDeferredWords`: this is a set of strings that are for when a word (variable or function) is used but its usage needs to be deferred. For example:
`{{ range(0, deferred|int) }}` would result in `usedDeferredWords == ["range", "deferred", "int"]`. This set can help find words that will not be supported in second rendering pass or would require additional dependencies to evaluate (if it is a custom function, for example).

When handling `EagerTokens`, any `setDeferredWords` are marked as DeferredValues on the context. `usedDeferredWords` are marked as DeferredValues on the context, but only if those words already existed in the context (this prevents something like the "range" function from being put onto the context itself.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -29,6 +29,7 @@ import com.hubspot.jinjava.lib.fn.FunctionLibrary;
 import com.hubspot.jinjava.lib.fn.MacroFunction;
 import com.hubspot.jinjava.lib.tag.Tag;
 import com.hubspot.jinjava.lib.tag.TagLibrary;
+import com.hubspot.jinjava.lib.tag.eager.EagerToken;
 import com.hubspot.jinjava.tree.Node;
 import com.hubspot.jinjava.util.DeferredValueUtils;
 import com.hubspot.jinjava.util.ScopeMap;
@@ -78,6 +79,7 @@ public class Context extends ScopeMap<String, Object> {
   private final Set<String> resolvedFunctions = new HashSet<>();
 
   private Set<Node> deferredNodes = new HashSet<>();
+  private Set<EagerToken> eagerTokens = new HashSet<>();
 
   private final ExpTestLibrary expTestLibrary;
   private final FilterLibrary filterLibrary;
@@ -93,6 +95,7 @@ public class Context extends ScopeMap<String, Object> {
   private final Stack<String> renderStack = new Stack<>();
 
   private boolean validationMode = false;
+  private boolean protectedMode = false;
 
   public Context() {
     this(null, null, null);
@@ -170,6 +173,7 @@ public class Context extends ScopeMap<String, Object> {
     resolvedFunctions.clear();
     dependencies = HashMultimap.create();
     deferredNodes = new HashSet<>();
+    eagerTokens = new HashSet<>();
   }
 
   @Override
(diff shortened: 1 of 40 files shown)


src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
```diff
@@ -126,6 +151,24 @@ private static String rebuildTemplateForNodes(Set<Node> nodes) {
     return joiner.toString();
   }
 
+  private static String rebuildTemplateForEagerTagTokens(
```
Joeoh, line 154: A test for this could be good

Joeoh (commented):
Thanks for this, looking great!

boulter:
I wonder if, instead of fixed behavior such as eager or deferred, we introduce a render "phase" string to the interpreter. Then each tag can evaluate as much or as little as it wants in that phase.

src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -526,4 +551,13 @@ public void addDependencies(SetMultimap<String, String> dependencies) {
   public SetMultimap<String, String> getDependencies() {
     return this.dependencies;
   }
+
+  public boolean isProtectedMode() {
```
boulter, line 555: what does this do?

jasmith-hs: It's (going to be) used to signify that certain tags need to have different functionality. For example when executing a `SetTag`, it can normally just be executed and affect the variable on the context. However, if we are evaluating within an IfTag where the correct branch path is unknown due to a deferred value, then if a value is replaced or updated during this "protected" execution mode, then that has to be handled specially.

PR #530, opened 2020-11-10
add size limited pymaps and pylists

boulter:
User-provided templates should not be allowed to use unlimited resources. This is a failsafe that adds a limit the maximum size of lists and maps. The default is no (practical) limit.

This is similar to https://github.com/HubSpot/jinjava/pull/137

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
@@ -1,6 +1,10 @@
 package com.hubspot.jinjava.objects.collections;
 
 import com.hubspot.jinjava.interpret.CollectionTooBigException;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateError;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorType;
 import com.hubspot.jinjava.objects.PyWrapper;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -8,6 +12,7 @@ import java.util.List;
 
 public class SizeLimitingPyList extends PyList implements PyWrapper {
   private int maxSize;
+  private boolean hasWarned;
 
   private SizeLimitingPyList(List<Object> list) {
     super(list);
@@ -69,6 +74,21 @@ public class SizeLimitingPyList extends PyList implements PyWrapper {
   private void checkSize(int newSize) {
     if (newSize > maxSize) {
       throw new CollectionTooBigException(newSize, maxSize);
+    } else if (!hasWarned && newSize >= maxSize * 0.9) {
+      hasWarned = true;
+      JinjavaInterpreter
+        .getCurrent()
+        .addError(
+          new TemplateError(
+            ErrorType.WARNING,
+            ErrorReason.COLLECTION_TOO_BIG,
+            String.format("List is at 90%% of max size (%d of %d)", newSize, maxSize),
+            null,
+            -1,
+            -1,
+            new CollectionTooBigException(newSize, maxSize)
+          )
+        );
     }
   }
 }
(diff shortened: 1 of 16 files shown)


mattcoley (commented):
Throwing the `OutputTooBigException` in the `LengthLimitingStringBuilder` has been a bit of a mess trying to handle that runtime exception in multiple places. Would it make better sense to truncate the collection to keep it under the limit and add a warning? This would protect against the situation where data being pulled in gets continually larger until it trips the limit which could break templates rather than just truncate them.

boulter:
This throws an `IndexOutOfRangeException` which extends `InterpretException`, so may be already handled. `OutputTooBigException` extends `RuntimeException` directly. You could argue that `IndexOutOfRangeException` is not quite right and maybe I could add another exception which also extends `InterpretException`.

src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
```diff
@@ -0,0 +1,90 @@
...
+  public void add(int index, Object element) {
+    if (size() >= maxSize) {
+      throw createOutOfRangeException(size() + 1);
+    }
+    super.add(index, element);
+  }
+
+  @Override
+  public boolean addAll(int index, Collection<?> elements) {
+    if (size() + elements.size() >= maxSize) {
```
jasmith-hs, line 62: Integer overflow might be a concern. It should be safer to do `size() >= maxSize - elements.size()`

src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
```diff
@@ -0,0 +1,90 @@
...
+  public boolean add(Object element) {
+    if (size() >= maxSize) {
+      throw createOutOfRangeException(size() + 1);
+    }
+    return super.add(element);
+  }
+
+  @Override
+  public void add(int index, Object element) {
+    if (size() >= maxSize) {
```
jasmith-hs, line 54: To be consistent with `addAll`, could check what the size would be after adding the one element

jasmith-hs (commented):
I like this idea. If it's wrapped in an `ELException`, it would be handled [here](https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java#L146) by adding a warning-level template error

src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
boulter: I think we will run out of memory long before this if an array has 2.4B items in it!

boulter:
@mattcoley I decided to create a new `CollectionTooBigException` because `IndexOutOfRangeException` was not appropriate and this is a pretty severe error. I don't want to just ignore it and continue because it could end up creating thousands of error messages. These limits are really high, so I don't think this will come up very often, but it should save the renderer when it does.

jasmith-hs:
What if it's a warning when it's at like 90% capacity to help prevent the situation that Matt described

src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
```diff
@@ -1,13 +1,18 @@
...
+import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorType;
 import com.hubspot.jinjava.objects.PyWrapper;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
 public class SizeLimitingPyList extends PyList implements PyWrapper {
   private int maxSize;
+  private boolean hasWarned;
```
jasmith-hs, line 15: Nice

boulter:
> What if it's a warning when it's at like 90% capacity to help prevent the situation that Matt described

Done. Check it out.

jasmith-hs (approved):

PR #531, opened 2020-11-12

michaelpro1:
This change includes the following:

- Upgrades to filters with >1 params that have 100% code coverage
- There are some fixes to Jinjavadoc as they didn't match unit tests

@boulter This is following up on the `AbstractFilter` #523 , these are draft changes so let me know what you think of these changes and whether you'd prefer split PRs too

Jinjavadoc alignment with tests includes:

- DateTimeFormatFilter
- JoinFilter
- various `number` => `int` changes where filter actually required an integer - the conversion is done in AbstractFilter

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
@@ -6,6 +6,7 @@ import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.lib.fn.Functions;
 import com.hubspot.jinjava.objects.date.StrftimeFormatter;
+import java.util.Map;
 
 @JinjavaDoc(
   value = "Formats a date object",
@@ -17,19 +18,19 @@ import com.hubspot.jinjava.objects.date.StrftimeFormatter;
   ),
   params = {
     @JinjavaParam(
-      value = "format",
+      value = DateTimeFormatFilter.FORMAT_PARAM,
       defaultValue = StrftimeFormatter.DEFAULT_DATE_FORMAT,
       desc = "The format of the date determined by the directives added to this parameter"
     ),
     @JinjavaParam(
-      value = "timezone",
-      defaultValue = "utc",
+      value = DateTimeFormatFilter.TIMEZONE_PARAM,
+      defaultValue = "UTC",
       desc = "Time zone of output date"
     ),
     @JinjavaParam(
-      value = "locale",
+      value = DateTimeFormatFilter.LOCALE_PARAM,
       type = "string",
-      defaultValue = "us",
+      defaultValue = "en-US",
       desc = "The language code to use when formatting the datetime"
     )
   },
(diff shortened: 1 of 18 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
```diff
@@ -40,19 +41,29 @@
     )
   }
 )
-public class DateTimeFormatFilter implements Filter {
+public class DateTimeFormatFilter extends AbstractFilter implements Filter {
+  public static final String FORMAT_PARAM = "format";
```
boulter, line 45: I really appreciate the static constants!

src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
```diff
@@ -64,87 +57,21 @@ public String getName() {
...
-  public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
-    if (var instanceof String) {
-      int length = DEFAULT_TRUNCATE_LENGTH;
-      String ends = DEFAULT_END;
-
-      if (args.length > 0) {
-        try {
-          length = Integer.parseInt(Objects.toString(args[0]));
-        } catch (Exception e) {
-          ENGINE_LOG.warn(
```
boulter, line 118: wow, this removes a lot of code! Can you ensure that this is still covered in the unit tests?

boulter (commented):
I like it. It seems to clean up a lot of code.

src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
```diff
@@ -40,19 +41,29 @@
...
-    } else {
+  public Object filter(
+    Object var,
+    JinjavaInterpreter interpreter,
+    Map<String, Object> parsedArgs
+  ) {
+    String format = (String) parsedArgs.get(FORMAT_PARAM);
+    String timezone = (String) parsedArgs.get(TIMEZONE_PARAM);
+    String locale = (String) parsedArgs.get(LOCALE_PARAM);
+    if (format == null && timezone == null && locale == null) {
```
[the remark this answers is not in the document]
michaelpro1: I could probably fix up the Functions code, but currently it errors out on zone being null. Let me know if you are happy for me to change it

src/main/java/com/hubspot/jinjava/lib/filter/TruncateHtmlFilter.java
michaelpro1: Ah I see your point, it ignores parsing errors. Given that we are using `defaultValue` now did you want to keep `parse-and-ignore-error` functionality or rely on correct user input/sensible default value?

boulter: I think we should just confirm that it still validates the value but we don't need to log a warning. It's pretty useless.

michaelpro1: Take a look at the last commits, I've added a default on error. Additionally I've fixed up the number parsing as NumberUtils seems to default to 0 if not parsed.

Issue #271, opened 2019-01-09
Add equality between Enums and Strings.


mattcoley:
There isn't a way to declare an enum value in Jinjava which makes it difficult to check for different enum values. 
`{% if enum == 'ENUM_VALUE' %}` should ideally work since `{{ enum }}` will display a string value.

mattcoley:
In the HubSpot case, the enum would be something like a language enum that would print out the language tag `en`, `ge`, `jp`, etc.

mattcoley:
The hacky way to get around this is to do `tojson|fromjson` on an object to convert these enums to strings. We need a better long term solution.

PR #540, opened 2020-11-23

mattcoley:
Reverts HubSpot/jinjava#539

Looking into some diffs around datetimeformat filter

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/TruthyTypeConverter.java
+++ b/src/main/java/com/hubspot/jinjava/el/TruthyTypeConverter.java
@@ -93,6 +93,7 @@ public class TruthyTypeConverter extends TypeConverterImpl {
       return "";
     }
 
+    // super() behavior breaks equality between enums and strings so removed here.
     if (value instanceof String) {
       return (String) value;
     }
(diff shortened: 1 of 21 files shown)


mattcoley:
The main issue has to do with https://github.com/HubSpot/jinjava/issues/271. I fixed this by manually overriding the `coerceToString` method. This has the added effect of enums not roundtripping however, it adds equality between enums and strings which is more desirable. In Jinjava we lose all type information on conversion so we never coerce back to the enum.

boulter (approved):

michaelpro1:
out of curiousity whats an example of an enum used?

mattcoley:
There isn't a way to declare an enum or create one via Jinjava, however an often use case is to add an enum value to the context. Previously functions would typically accept generic Object values and call `toString()` to extract the String, however the TruthyTypeConverter extends a type converter from Juel that was calling `name()` which can be different depending on the enum. I am planning on merging this soon and rechecking our diff reports to make sure everything is OK.

michaelpro1:
Noticed that this PR has been sitting for a little while, did you need any help with it?

mattcoley:
Let me merge this in again and check our custom diff reports.

Issue #532, opened 2020-11-12
Eager Execution


jasmith-hs:
This issue is to serve as the status and main point of info for the Eager Execution project that I'm working on.
Check out https://github.com/HubSpot/jinjava/issues/519 for the initial draft PR that I opened for this project. (This PR will stay a draft)

## Purpose
To partially render as much of a Jinjava/HubL template as possible to allow for a lightweight Jinjava engine to be used from the cdn worker at serve-time to eliminate as much serve-time dependency on the cos-renderer.

When executing eagerly, if there exist any deferred values in the Jinjava code, then the template will need to be rendered a second time with the values not deferred. The second render should have the same output as if the template was only rendered once, normally, with all values resolved on the context. Only the tags and functions that directly depend on the deferred values will end up getting deferred, and they are marked this way with a new set of objects called the context’s `eagerTokens`.

## Background
TagNodes will become Deferred nodes if the `tag.interpretOutput()` throws a DeferredValueException. The node, which includes its children, will get put into the context’s `deferredNodes`. ExpressionNodes will become Deferred nodes if the expression within the tag throws a DeferredValueException when `interpreter.resolveELExpression()` is called.

Since the node and its children get deferred, it is unknown what kind of operations are therefore deferred. Additionally, there may be logic that could have been executed but got deferred because the parent node was deferred. Without eager execution, it is a requirement that an initial pass that results in deferred nodes, when rendered for a second pass, must be rendered with an engine with the same context.

Eager execution is therefore beneficial for 2 main reasons: The final rendering pass will be faster as there will be less unevaluated tags, functions, filters, and expressions to evaluate. And second, it allows for the final pass to be done by a different jinjava interpreter. (Limited access to functions, filters, etc; limited access to context; sandbox/separation from other services)

### Merged PR's
[(Branch) Execution Mode config option](https://github.com/jasmith-hs/jinjava/pull/2)
[Chunk Resolver](https://github.com/HubSpot/jinjava/pull/525)
[Introduce Eager Tokens](https://github.com/HubSpot/jinjava/pull/529)
[Tests and EagerTagDecorator skeleton](https://github.com/HubSpot/jinjava/pull/542)
[Run prepareContext method](https://github.com/HubSpot/jinjava/pull/543)
[Use strategy pattern for ExpressionNode](https://github.com/HubSpot/jinjava/pull/544)
[Full EagerTagDecorator Functionality (minus macro functions)](https://github.com/HubSpot/jinjava/pull/545)
[...]

PR #542, opened 2020-11-23
Add skeleton for test cases and EagerTagDecorator

jasmith-hs:
---
Moved from https://github.com/jasmith-hs/jinjava/pull/1

This PR mainly serves to introduce the various tests that I'm using to test eager execution with the different tags.
The `EagerTest` uses an `ExpectedTemplateInterpreter` which takes the name of a jinja file (`file-name.jinja`) and compares its interpreted output to that of `file-name.expected.jinja`. Some second pass tests are added too which use a suffix of `.expected.expected.jinja`

Additionally, this adds an `EagerTagFactory` which is used to register the tags which are getting decorated with the `EagerTagDecorator` for general functionality, as well as specific functionality for each tag. The way it's currently set up, the eager overrides for tags are registered by getting an instance of one from the `EagerTagFactory` and registering the tag on the context. Since they have the same name as the original tag that they're decorating, they replace the original tags in the `TagLibrary`.
The tests here are marked with `@Ignore` because the functionality to make them pass isn't added yet.

Most of these changes are the tests for which the logic is not included yet. cc @boulter @Joeoh @mattcoley


boulter:
Wow, that's a lot of new files to maintain. Can we run existing files through a processor or something so we don't have to maintain them all?

jasmith-hs:
> Wow, that's a lot of new files to maintain. Can we run existing files through a processor or something so we don't have to maintain them all?

I made the tests in this way because it's really easy to write new tests and very straightforward to understand what's happening within the tests. I don't reckon that the resource files will be difficult to maintain (although adding a new test would add 2 new files), but I'm curious about what you mean by processing the existing files.

boulter:
I assumed you didn't manually write 71 new test files, but instead modified existing ones. Perhaps you did write new ones. If they won't change very often and require updating, then that's fine.

jasmith-hs:
Yeah, they are not supposed to change. This PR adds just one new Test class, but most of the tests are run using the `.jinja` resource files that I wrote. Each of those files performs one test so they should remain static. Essentially, I took all the jinja template strings that could be defined within the java test file, and moved them into the resources folder to simplify the test class.

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class EagerTagFactory {
+  public static final Map<Class<? extends Tag>, Class<? extends EagerTagDecorator<? extends Tag>>> EAGER_TAG_OVERRIDES = ImmutableMap
+    .<Class<? extends Tag>, Class<? extends EagerTagDecorator<?>>>builder()
+    .build();
+
+  @SuppressWarnings("unchecked")
+  public static <T extends Tag> Optional<EagerTagDecorator<T>> getEagerTagDecorator(
+    Class<T> clazz
+  ) {
```
jasmith-hs, line 17: This will get called to register each tag in the `EagerExecutionMode.prepareContext()` (seen [here](https://github.com/HubSpot/jinjava/pull/543/files#diff-b251361da38e79a3eeee02262b427a58336d8666de114d39672f86b85e2fcae7R21))
@boulter This is the basic principle for how I am separating the eager execution code from the default execution code. The tags get wrapped with an `EagerTagDecorator` in this factory so that they can be registered to override the default tags

jasmith-hs:
This isn't enabled anywhere and is mostly adding the test cases, but I want to get more of the functionality I've written into PR form so I'm going to merge this in so that I can get more PR's opened up that depend on this.

PR #554, opened 2020-11-30
Implement EagerForTag

jasmith-hs:
---

This PR adds the functionality to eagerly execute a for tag. Some refactoring is done to the `ForTag` to allow for code to be reused in the `EagerForTag` as well.

If the expression to determine the loop variable(s) has a deferred value in it, then the `eagerInterpret()` method will be called. The loop variable(s) become deferred variables as their values are not yet known and the image of the for tag is preserved, and the children are executed in a protected mode to prevent modification to the context. This allows for the for tag to be re-evaluated during a subsequent rendering pass while retaining the proper end result.

In the scenario where the loop variable can be determined, but there are contents within the for loop that end up getting eagerly executed, then the default ForTag behaviour is run as the output from those children is safely preserved by the eager execution architecture.

---
Examples:
```
{% set foo = 'hello' %}
{% for item in deferred %}
{{ item }}
{% print foo %}
{% endfor %}
```
Since `{% print foo %}` doesn't depend on the deferred value, with eager execution it can be pre-rendered like:

```
{% for item in deferred %}
{{ item }}
hello
{% endfor %}
```
Of course, if it was instead `{% print item %}`, it wouldn't get evaluated, like `{{ item }}`.

Also with `resolved = ['piece', 'chunk', 'bite']`:

```
{% for item in resolved %}
A {{ item }} of {{ deferred }}.
{% endfor %}
```
could get pre-rendered to:

```
A piece of {{ deferred }}.
A chunk of {{ deferred }}.
A bite of {{ deferred }}.
```

cc @jboulter @Joeoh


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerForTag.java
```diff
@@ -0,0 +1,76 @@
...
+            joiner.toString(),
+            tagToken.getLineNumber(),
+            tagToken.getStartPosition(),
+            tagToken.getSymbols()
+          ),
+          chunkResolver.getDeferredWords(),
+          new HashSet<>(loopVars)
+        )
+      );
+    return (newlyDeferredFunctionImages + joiner.toString());
```
Joeoh, line 74: Maybe join this string just once

Joeoh (approved):
This is really great, thanks!

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerForTag.java
Joeoh: Also, should we be using the LengthLimitingStringBuilder here too? I'm thinking of something like {% for i in range(0,MAX_INT) %} with a large body where the output could become huge

jasmith-hs: That's a good call

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerForTag.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    String loopExpression = getTag().getLoopExpression(helperTokens, loopVars);
+    ChunkResolver chunkResolver = new ChunkResolver(
+      loopExpression,
+      tagToken,
+      interpreter
+    );
+
+    StringJoiner joiner = new StringJoiner(" ");
```
boulter, line 47: There's a guava joiner which I see used more commonly. One advantage is that you can declare it statically. I'm not sure if it's worth changing to that one.

jasmith-hs: I'll be swapping these usages out with [the length limiting joiner that I'm adding](https://github.com/HubSpot/jinjava/blob/3097aa7eabba4378d8331a15dd60a95b78722744/src/main/java/com/hubspot/jinjava/util/LengthLimitingStringJoiner.java) so I will see which makes more sense to use in that class.
Maybe I'll run a simple performance test to see which is faster

jasmith-hs: Looks like the java.util.StringJoiner is 2-3x faster than the guava joiner so I'll stick with it.
Time to join 10k and 1 million `a` characters with a `,` character:

| Number of elements |StringJoiner | Guava Joiner |
| --- | --- | --- |
| 10000 | 1 ms | 8 ms |
| 1000000 | 17 ms | 51 ms |

boulter: Interesting. Thanks for verifying. The JDK StringJoiner was introduced in Java 8, so maybe it's just historical that we use it. That said, the guava version has some nice features like `skipNulls`

PR #555, opened 2020-12-01
Limit eager length

jasmith-hs:
Based on the suggestion from @Joeoh here: https://github.com/HubSpot/jinjava/pull/554#discussion_r533217276
Having the `LengthLimitingStringBuilder` is useful to prevent code explosion when doing eager execution which could result in very large output. Using this will catch nodes that are creating too much output and defer them rather than continuing with their eager execution. This PR also introduces a StringJoiner wrapper called `LengthLimitingStringJoiner` which functions like the string builder, but allows for a delimiter to be specified.

This approach allows the overriding subclasses to have the flexibility to use the length limiters on some or all of their output. By default, the entire output is limited, as well as the tag image reconstruction, and the set tag pre-pending. If any are too large, then eager execution will quit and the node will be marked and handles as a deferred node via `handleDeferredNode()`. (This happens when the DeferredValueException is thrown but not caught within the decorator)

In the future, this may be substituted or supplemented with a check on the overall output size of the render from eager execution.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
@@ -11,6 +11,8 @@ import com.hubspot.jinjava.interpret.DeferredValueException;
 import com.hubspot.jinjava.interpret.DisabledException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter.InterpreterScopeClosable;
+import com.hubspot.jinjava.interpret.OutputTooBigException;
+import com.hubspot.jinjava.interpret.TemplateError;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.lib.fn.MacroFunction;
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
```diff
@@ -51,7 +55,17 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
...
       return tag.interpret(tagNode, interpreter);
     } catch (DeferredValueException | TemplateSyntaxException e) {
-      return wrapInAutoEscapeIfNeeded(eagerInterpret(tagNode, interpreter), interpreter);
+      try {
+        return wrapInAutoEscapeIfNeeded(
+          eagerInterpret(tagNode, interpreter),
+          interpreter
+        );
+      } catch (OutputTooBigException e1) {
+        interpreter.addError(TemplateError.fromOutputTooBigException(e));
```
boulter, line 64: Should you be using e or e1 here?

jasmith-hs: e1, thanks

Joeoh (approved):

PR #557, opened 2020-12-01
Implement EagerSetTag

jasmith-hs:
---

This PR first extracts some functionality from the SetTag into static methods to be reusable by the EagerSetTag.

There are a few main use cases for interpreting a set tag while in `EagerExecutionMode`:
1. It's a normal set tag that doesn't work with or around any deferred values.
This can be interpreted as normal. The new value is set onto the context.

2. It's trying to set some value to a deferred value (`{% set foo = deferred + 1 %}`).
The target value will become deferred, and the set tag will be reconstructed and included in the output to get run during a later pass. This reconstructed set tag will get registered as an `EagerToken` on the context.

3. It's trying to set a deferred value to something (`{% set deferred = 9 %}`).
The set tag will be reconstructed and included in the output. This reconstructed set tag will get registered as an `EagerToken` on the context.

4. It's trying to modify a value on the context while in protected mode.
The target value will become deferred, and the set tag will be reconstructed and included in the output to get run during a later pass. This reconstructed set tag will get registered as an `EagerToken` on the context.
*If the execution of the right-hand side of the equation causes any changes in the context, then additional set tags will need to get prepended to the output to preserve the initial context. *This part is true for anything running in "protected mode" that causes changes to the values on the context scope map.
---
There's also the case where the set tag is executed within an imported file, and that's handled by converting it to a DoTag that performs an update on the imported map. For instance:
foo.jinja:
```
{% set foobar = deferred %}
```
main.jinja:
```
{% import 'foo.jinja' as foo %}
{{ foo.foobar }}
```
This will become like:
```
{% set foo = {} %}
{% do foo.update({'foobar', deferred}) %}
{{ foo.foobar }}
```

cc @jboulter @Joeoh let me know if I missed something in this writeup or if anything is unclear.


Joeoh:
Nice! Am I right in saying that the main change here is handling the imports? The other behaviour existed from handling the DeferredValueExcecption -> reconstructImage of the set tag right?

jasmith-hs:
Yeah, it's pretty similar, one change is that the set tag gets partially reconstructed and partially evaluated. `{% set foo = deferred + bar %}` -> `{% set foo = deferred + 15 %}`. And case 4 makes it safe to be used inside a deferred IfTag or ForTag.

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
```diff
@@ -0,0 +1,133 @@
...
+  public EagerSetTag() {
+    super(new SetTag());
+  }
+
+  public EagerSetTag(SetTag setTag) {
+    super(setTag);
+  }
+
+  @Override
+  public String getEagerTagImage(TagToken tagToken, JinjavaInterpreter interpreter) {
```
boulter, line 27: A more general question here:

You probably understand this better than me at this point, but why don't we implement this by manipulating the tree directly?

The way most compilers work is by parsing the code up into a tree (which we have), doing n passes through the tree to manipulate nodes and perform evaluation when possible. Then they write out the code in one pass. Could we do that here instead of returning a string as the output?

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
```diff
@@ -0,0 +1,133 @@
...
+    StringBuilder prefixToPreserveState = new StringBuilder(
+      interpreter.getContext().isProtectedMode()
+        ? resolvedExpression.getPrefixToPreserveState()
+        : ""
+    );
+    String[] varTokens = variables.split(",");
+
+    if (
+      chunkResolver.getDeferredWords().isEmpty() &&
+      !interpreter.getContext().isProtectedMode()
```
boulter, line 69: did we decide on a different name for `isProtectedMode`? Every time I see it, I can't remember what it protects from.

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
jasmith-hs: The jinjava engine is an interpreter rather than a compiler so it parses one line at a time into its output. Not lines exactly, as it first parses the template into a tree, and then it interprets each node in that tree as it sees it (because unlike python, jinja doesn't have a strong line-based syntax). 

The way I have this implemented is so that nodes are still evaluated on the fly, but their evaluation is handled differently when a value is deferred. Rather than throwing an error there, a specific output can be generated such that running that output with resolved values will produce the same output as running the original with resolved values, but faster. This approach allows for the sequential node evaluation to be done as normal.

In a different rendering engine that's more c-based than python-based, it would probably be done by manipulating the tree

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
jasmith-hs: I have not yet come up with a better name, and I'm aware that it's not exactly clear what it does by the name.

I have defined "protected mode" as a mode where the surrounding context cannot be modified such that the state of the context before entering "protected mode" will be the same as after exiting it. Any changes will result in the values attempted to be modified becoming `DeferredValue`s.

This mode is commonly seen in IfTags and ForTags when the children are being rendered but their results need to be protected such that when the if branch/for loop is finally executed normally, the output will be correct.

Something like `isDeferredExecutionMode` might be more understandable as the code being evaluated is being partially deferred, knowing that it's final execution is being deferred

boulter: `isDeferredExecutionMode` sounds good to me.

PR #558, opened 2020-12-01
Implement EagerPrintTag and EagerDoTag

jasmith-hs:
---

Since the print tag and do tag are nearly identical (print tags include the expression output, while do tags omit the output), both of their eager implementations are included in this PR.

Print and Do tags are also similar to Set tags, but they are essentially simpler. Rather than taking the output of one or more expressions and storing it into a key on the context scope map, a single expression is simply evaluated, and in the case of a print tag, it is also output. So see https://github.com/HubSpot/jinjava/pull/557 for an overview of the SetTag's logic, as the logic in this PR is a simplified version of it.

These are the use cases:
1. It's a normal print/do tag that doesn't work with or around any deferred values.
This can be interpreted as normal. The expression is resolved and the value is output if we're dealing with a print tag.

2. It's trying to evaluate an expression using a deferred value (`{% print deferred + 1%}`).
The expression will be simplified if possible by using the [ChunkResolver](https://github.com/HubSpot/jinjava/pull/525). The print/do tag will be reconstructed/re-output and then it will get registered as an `EagerToken` on the context.

3. It's trying to modify a value on the context while in protected mode.
Just like with a set tag, the target value will become deferred, and the print/do tag will be reconstructed and included in the output to get run during a later pass. This reconstructed tag will get registered as an `EagerToken` on the context.
*If the execution of the right-hand side of the equation causes any changes in the context, then additional set tags will need to get prepended to the output to preserve the initial context. *This part is true for anything running in "protected mode" that causes changes to the values on the context scope map.

cc @jboulter @Joeoh let me know if you want to see more detailed examples here


src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerDoTagTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @After
+  public void teardown() {
+    JinjavaInterpreter.popCurrent();
+  }
+
+  @Test
+  public void itHandlesDeferredDo() {
+    context.put("foo", 2);
+    expectedNodeInterpreter.assertExpectedOutput("handles-deferred-do");
```
boulter, line 45: just as a personal preference, if inputs and expected files are short enough, I would just put them right in the code here. It's way easier to read when you don't have to flip back and forth  between files.

Someday we'll get [text blocks](https://openjdk.java.net/jeps/355).

jasmith-hs: Okay, I'll make a note to clean up the one-liners, and text blocks would be perfect here if we get them

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerDoTag.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  public String getEagerTagImage(TagToken tagToken, JinjavaInterpreter interpreter) {
+    String expr = tagToken.getHelpers();
+    if (StringUtils.isBlank(expr)) {
+      throw new TemplateSyntaxException(
+        interpreter,
+        tagToken.getImage(),
+        "Tag 'print' expects expression"
```
Joeoh, line 28: Should be "do"

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerDoTag.java
```diff
@@ -0,0 +1,74 @@
...
+      );
+    }
+    ChunkResolver chunkResolver = new ChunkResolver(expr, tagToken, interpreter);
+    EagerStringResult resolvedExpression = executeInChildContext(
+      eagerInterpreter -> chunkResolver.resolveChunks(),
+      interpreter,
+      true
+    );
+    StringJoiner joiner = new StringJoiner(" ");
```
Joeoh, line 37: Looks like this logic is the same for both tags, and maybe others? Should we split it out into a funciton?

jasmith-hs: Yes, that is a good call.

PR #559, opened 2020-12-03
Extract some logic into static methods for from/import tag classes

jasmith-hs:
These classes have `interpret()` methods that are very long, and some functionality within them is useful to share when doing eager execution. So this PR refactors those 2 methods to use several static methods. Since these changes are just refactoring, I am putting them into a separate PR from the EagerFromTag and EagerImportTag logic for simplicity.


jasmith-hs:
There are more additions than deletions because there are several new static methods so their definitions, and invocations increase the line count. However, the code is easier to understand as the methods are more cohesive given their more limited responsibilities now.

PR #561, opened 2020-12-07
Eager execution tests for ExtendsTag and IncludesTag

jasmith-hs:
---

This PR doesn't really add any functionality because the `ExtendsTag` and `IncludesTag` are handled with the generic logic in the `EagerTagDecorator`. Instead, this PR just adds tests for the logic to ensure that these tags work properly when doing eager execution and when values are deferred.


jasmith-hs:
This basically is adding more test coverage for the extends and includes tags

PR #566, opened 2020-12-09

mattcoley:
Reverts HubSpot/jinjava#565

I am going to pull this locally and find the bug that is causing issues with datetimeformat.


michaelpro1:
@mattcoley does excluding DateTimeFormatFilter from upgrades fix regression? In that case you can keep old functionality for that filter and merge it

Issue #496, opened 2020-08-26
Rendering is not happening when default value for an argument is single quote '''


ashrithasurabhi:
{% macro config_appGig(vlan1='1', vlan2 ='2', separator=''') %}
{{'switchport vlan ' ~ vlan1 ~ separator ~ vlan2}}
{% endmacro %}

{{config_appGig(100,200) }}

Expected: 

switchport vlan 100'200

Actual: No errors are thrown but same content is returned without resolving the macro

{% macro config_appGig(vlan1=1,vlan2 ='2',separator=''') %}
{{'switchport vlan ' ~ vlan1 ~ separator ~ vlan2}}
{% endmacro %}

{{config_appGig(100,200) }}

jasmith-hs:
You either need to escape the `'` character or surround it with double-quotes instead. As written, the syntax is invalid for Jinja/Jinjava 

```
{% macro config_appGig(vlan1='1', vlan2 ='2', separator='\'') %}
{{'switchport vlan ' ~ vlan1 ~ separator ~ vlan2}}
{% endmacro %}

{{config_appGig(100,200) }}
```
Should do it for you. The result of interpreting the above is: `switchport vlan 100'200`.

However, it does look like there is a bug when defining a macro function that uses commas within the default string values such as: `{% macro config_appGig(vlan1=1,vlan2 ='2',separator=',') %}`

PR #569, opened 2020-12-10
Allow commas in kwargs for macro tags

jasmith-hs:
Fixes the problem exposed in https://github.com/HubSpot/jinjava/issues/496 by using the chunk resolver to split the arguments in the macro tag.

The chunk resolver respects the hierarchy of quotes, parentheses, brackets, etc so it can determine when a comma is in the place to actually signify a new keyword argument, rather than something related to the value.

This requires an option to be added in the chunk resolver to not do the interpreting, and just output the raw strings that are split into a list by a comma.


jasmith-hs:
This is no longer how the chunkresolver works so it cannot be used in this way to fix the issue.

PR #570, opened 2020-12-11
Add filter to interpret a string early

liamrharwood:
This PR adds a new `RenderFilter` which can be used to interpret a template string early.

For instance, the following:
```
{{ "{% if null %}}Hello{% else %}World{% endif %}"|render }}
```
would output `World`. However, you could use the result of the filter in other filters, like so:
```
{{ "{% if null %}}Hello{% else %}World{% endif %}"|render|lower }}
```
which would output `world`.

This is useful in cases where a function or other construct results in an expression that must be interpreted before applying other filters / transformations.

I'm open to naming this something different if people don't like `render`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
@@ -1,0 +1,29 @@
+package com.hubspot.jinjava.lib.filter;
+
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import java.util.Objects;
+
+@JinjavaDoc(
+  value = "Renders a template string early to be used by other filters and functions",
+  input = @JinjavaParam(value = "s", desc = "String to render", required = true),
+  snippets = {
+    @JinjavaSnippet(
+      code = "{{ \"{% if my_val %}} Hello {% else %} world {% endif %}\"|render }}"
+    ),
+  }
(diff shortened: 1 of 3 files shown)


mattcoley (approved):
I don't think this is more dangerous than the regular nested interpretation stuff but wanted to get @jasmith-hs thoughts on this as well

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -0,0 +1,29 @@
...
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import java.util.Objects;
+
+@JinjavaDoc(
+  value = "Renders a template string early to be used by other filters and functions",
+  input = @JinjavaParam(value = "s", desc = "String to render", required = true),
+  snippets = {
+    @JinjavaSnippet(
+      code = "{{ \"{% if my_val %}} Hello {% else %} world {% endif %}\"|render }}"
```
jasmith-hs, line 14: There's an extra `}` after `{% if my_val %}`

src/test/java/com/hubspot/jinjava/lib/filter/RenderFilterTest.java
```diff
@@ -0,0 +1,23 @@
...
+  private RenderFilter filter;
+
+  @Before
+  public void setup() {
+    filter = new RenderFilter();
+  }
+
+  @Test
+  public void itWritesObjectAsString() {
+    String stringToRender = "{% if null %}}Hello{% else %}world{% endif %}";
```
jasmith-hs, line 19: There's an extra `}` here too

jasmith-hs (approved):
I like this idea, just one change to make in the snippet to eliminate confusion

PR #572, opened 2020-12-21
Variable function evaluator

GregDThomas:
This PR adds the ability to supply a "Variable Function" - a `java.util.Function<String, Object>` that will map a variable name to it's value.

This is useful when you do not know the variables that may exist in a template at run time. Rather than populate the `Map<String, Object>` with every possible variable that can be supported, simply set the variable function that will be called at runtime. The code in your `Function` can then determine the value when it is needed.

Note that this is slightly different to the LazyExpression concept that already exists - this is a `Supplier`, so requires the caller to know the list of variables to populate in advance.


GregDThomas:
Note; this is my first PR for this project; firstly thank for maintaining it, secondly if you want any improvements to be made in order to get this merged, please let me know and I'll see what I can do.

GregDThomas:
PPS. Should also say I added a Maven wrapper as well; simply invoke mvnw instead of mvn if you don't alreaedy have the correct version of Maven installed.

GregDThomas:
Hi - any thoughts of what would need to be changed to get this merged? Thanks :)

matthewwells:
+1 - This would be a really useful enhancement.  I have an exact use case for this dynamic processing of the variable names.

bhreinb:
+1 - for this PR...I have a use case for this PR in that I'd like to override the default value for a variable when It's not found...so hopefully it can be merged soon 🙏

lordnelson:
This is a useful feature to resolve many template variables supplied from multiple sources

akseiya:
+1 for this PR, for complex templates and costly variables (say we need to call APIs to have values) it's super useful to know at template resolution time which ones are needed and which ones are not

jasmith-hs (approved):
Thanks for adding this, it is a very nice option

GregDThomas:
Thank you for merging it, and your continued work on the project!

PR #575, opened 2020-12-23
Replicate handling of non-existent variables in ChunkResolver

jasmith-hs:
This PR addresses 2 bugs in the ChunkResolver:
- An even number of slashes followed by a quote character would be interpreted as an escaped quote when it shouldn't be escaped
- In Jinjava, when an unknown variable is used in an expression node, the result is an empty string:
  `Something {{ nonexistent }}!` -> `Something !` So match this functionality with the ChunkResolver, as it previously would just output the variable name

ChunkResolver is currently just used within Eager Execution so the scope of this bug is essentially nothing.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
+++ b/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
@@ -59,6 +59,8 @@ public class ChunkResolver {
     ']'
   );
 
+  private static final String VARIABLE_REGEX = "[A-Za-z_][\\w.]*";
+
   private final char[] value;
   private final int length;
   private final Token token;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
```diff
@@ -182,14 +184,23 @@ public String resolveChunks() {
...
     }
     miniChunkBuilder.append(resolveToken(tokenBuilder.toString()));
     chunks.add(resolveChunk(miniChunkBuilder.toString()));
     return chunks;
   }
 
+  private void setPrevChar(char c) {
+    if (c == '\\' && prevChar == '\\') {
+      // Backslashes cancel each other out for escaping when there's an even number.
+      prevChar = '\0';
```
jasmith-hs, line 198: Here it's just set to the null character as `prevChar` is only used to check escaping. So setting it to the null character simply ignores the slash for escaping purposes, allowing any number of slashes to be back-to-back.

PR #577, opened 2020-12-29
Better null handling in ChunkResolver

jasmith-hs:
Rather than returning an empty string `""`, a null value should be interpreted during the resolving process as null. There may be logic that treats `null` differently from `''`. After the string is resolved, however, a result that is exactly `null` should be returned as the empty Jinjava string (`''`) because `{{ null }}` should be interpreted to result in nothing.

This is tricky logic, and so I'm working to make the ChunkResolver handle edge cases like this exactly how they would normally be handled with expressions. This is part of making Eager Execution's output mirror that of the default execution.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
+++ b/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
@@ -105,7 +105,12 @@ public class ChunkResolver {
       .getThrowInterpreterErrors();
     try {
       interpreter.getContext().setThrowInterpreterErrors(true);
-      return String.join("", getChunk(null));
+      String expression = String.join("", getChunk(null));
+      if ("null".equals(expression)) {
+        // Resolved value of null as a string is ''.
+        return "''";
+      }
+      return expression;
     } finally {
       interpreter.getContext().setThrowInterpreterErrors(isThrowInterpreterErrorsStart);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
```diff
@@ -105,7 +107,12 @@ public String resolveChunks() {
       .getThrowInterpreterErrors();
     try {
       interpreter.getContext().setThrowInterpreterErrors(true);
-      return String.join("", getChunk(null));
+      String expression = String.join("", getChunk(null));
+      if (JINJAVA_NULL.equals(expression)) {
+        // Resolved value of null as a string is ''.
+        return JINJAVA_EMPTY_STRING;
```
jasmith-hs, line 113: This should be the only place where null needs to be swapped out for the empty string as this is the last step where the object must be resolved to a string. (That string could be presented in the final output of the render)

jasmith-hs: See https://private.hubteam.com/opengrok/xref/all/github.com-HubSpot-jinjava%23master/src/main/java/com/hubspot/jinjava/lib/expression/DefaultExpressionStrategy.java#22 where null is replaced with empty when converting to a string.

PR #589, opened 2021-01-25
Intial support for vsCodeTagSnippets

phalaphone:
Added new method to produce a string of tag snippets formatted for use in the hubspot vscode plugin.

Updated
(1/29) Output:
```
{% import ${1:path} ${2:import_name} %}

{% elif ${1:condition} %}

{% autoescape %}
{% endautoescape %}

{% from ${1:path} ${2:macro_name} %}

{% macro ${1:macro_name} ${2:argument_names} %}
{% endmacro %}

{% raw %}
{% endraw %}

{% else %}

{% if ${1:condition} %}
{% endif %}

{% print ${1:expr} %}

{% do %}

{% cycle ${1:string_to_print} %}

{% unless ${1:expr} %}
{% endunless %}

{% include ${1:path} %}

{% ifchanged %}
{% endifchanged %}

{% set ${1:var} = ${2:expr} %}

{% block ${1:block_name} %}
{% endblock %}

{% call %}
{% endcall %}

{% extends ${1:path} %}

{% for ${1:items_to_iterate} %}
{% endfor %}
```

### DIFF
--- a/src/main/java/com/hubspot/jinjava/doc/JinjavaDocFactory.java
+++ b/src/main/java/com/hubspot/jinjava/doc/JinjavaDocFactory.java
@@ -308,19 +308,31 @@ public class JinjavaDocFactory {
     snippet.append(tag.getName());
     int i = 1;
     for (JinjavaParam param : docAnnotation.input()) {
-      snippet.append(" ${" + i + ":" + param.value() + "}");
+      String inputValue = "${" + i + ":" + param.value() + "}";
+      if (param.value().equalsIgnoreCase("path")) {
+        inputValue = "'" + inputValue + "'";
+      } else if (param.value().equalsIgnoreCase("argument_names")) {
+        inputValue = "(" + inputValue + ")";
+      }
+      snippet.append(" " + inputValue);
       i++;
     }
 
     for (JinjavaParam param : docAnnotation.params()) {
-      snippet.append(" ${" + i + ":" + param.value() + "}");
+      String paramValue = "${" + i + ":" + param.value() + "}";
+      if (param.value().equalsIgnoreCase("path")) {
+        paramValue = "'" + paramValue + "'";
+      } else if (param.value().equalsIgnoreCase("argument_names")) {
+        paramValue = "(" + paramValue + ")";
+      }
+      snippet.append(" " + paramValue);
       i++;
     }
 
     snippet.append(" %}");
 
     if (tag.getClass().getAnnotation(JinjavaHasCodeBody.class) != null) {
-      snippet.append("\n${" + i + ":Code}");
+      snippet.append("\n$0");
     }
     if (tag.getEndTagName() != null) {
       snippet.append("\n{% " + tag.getEndTagName() + " %}");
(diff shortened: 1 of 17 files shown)


mattcoley (commented):
🆒  Looks like `if/elif` are missing the jinjavaparam in the doc, can you add it here?

mattcoley:
I think set tag is a bit custom here since I think you want `{% set ${1:var} = ${2:expr} %}`. Maybe it is worth adding a new annotation to specify an override for this custom behavior on the few tags that we need to adjust the snippet?

phalaphone:
I updated the output in at the top of the page. I did notice for a lot of them (i.e. if, else, macro, etc.) We have no jinjava annotation for body, would it be worthwhile to add that so we can represent where code would go in an if statement for example?

anthmatic:
> we can represent where code would go

@phalaphone do you mean something like this?

```
{% if ${1:condition} %}
  ${2:body}
{% endif %}
```
If so, I think that would be helpful. Each `${}` tag is a tabbable placeholder in VSCode

src/main/java/com/hubspot/jinjava/doc/JinjavaDocFactory.java
```diff
@@ -4,6 +4,7 @@
 import com.hubspot.jinjava.doc.annotations.JinjavaMetaValue;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.doc.annotations.JinjavaVSCodeSnippet;
```
anthmatic, line 7: I'm not sure if this should be VSCode specific. This syntax is technically borrowed from [Textmate](https://macromates.com/manual/en/snippets) and other editors use a similar format so I _think_ this could be used across other editors pretty easily

phalaphone:
@mattcoley @anthmatic 

Here is the current list:
```
{% import ${1:path} ${2:import_name} %}

{% elif ${1:condition} %}
${2:Code}

{% autoescape %}
{% endautoescape %}

{% from ${1:path} ${2:macro_name} %}

{% macro ${1:macro_name} ${2:argument_names} %}
${3:Code}
{% endmacro %}

{% raw %}
{% endraw %}

{% else %}
${1:Code}

{% if ${1:condition} %}
${2:Code}
{% endif %}

{% print ${1:expr} %}

{% do %}

{% cycle ${1:string_to_print} %}

{% unless ${1:expr} %}
${2:Code}
{% endunless %}

{% include ${1:path} %}

{% ifchanged %}
${1:Code}
{% endifchanged %}

{% set ${1:var} = ${2:expr} %}

{% block ${1:block_name} %}
{% endblock %}

{% call %}
{% endcall %}

{% extends ${1:path} %}

{% for ${1:items_to_iterate} %}
${2:Code}
{% endfor %}
```
Any changes before we get this out as a starting point?

anthmatic:
Thanks @phalaphone! I made some changes to some snippets below:

```jinja
{% import '${1:path}' ${2: as ${3:import_name}} %}

{% elif ${1:condition} %}
$0

{% autoescape %}
{% endautoescape %}

{% from '${1:path}' import ${2:macro_name} %}

{% macro ${1:macro_name}(${2:argument_names}) %}
$0
{% endmacro %}**

{% raw %}
{% endraw %}

{% else %}
$0

{% if ${1:condition} %}
$0
{% endif %}

{% print ${1:expr} %}

{% do ${1:expr} %}

{% cycle ${1:string_to_print} %}

{% unless ${1:expr} %}
$0
{% endunless %}

{% include '${1:path}' %}

{% ifchanged %}
$0
{% endifchanged %}

{% set ${1:var} = ${2:expr} %}

{% block ${1:block_name} %}
$0
{% endblock {$1} %}

{% call ${1:macro_name}(${2:argument_names}) %}
$0
{% endcall %}

{% extends '${1:path}' %}

{% for ${1:items} in ${2:list} %}
{{ ${1} }}$0
{% endfor %}```

The most notable change is that the code placeholder should just be `$0`, but there are a bunch of small changes to most of these. Let me know if you have any questions about any changes.

PR #590, opened 2021-01-27
Expose template errors during eager execution expression resolving

jasmith-hs:
Improves https://github.com/HubSpot/jinjava/issues/532
---

I found that fatal errors were not being added properly in eager execution, and this also was causing some discrepancies in the HTML output when there were fatal errors comparing eager execution mode with default execution mode. Before this PR, `{{ unknown(null) }}` would resolve to `unknown(null)` instead of ` ` (blank).

This PR changes it so that the final `resolveChunk()` call is made in the EagerExpressionStrategy*, which will have template errors that are thrown get put onto the interpreter, rather than being hidden. 

*Note that this cannot be done in the ChunkResolver because it is used to resolve expressions like `1,2,3` which throw syntax errors, as they are invalid expressions, but are used in places such as SetTags. We are hiding errors like these within the ChunkResolver with `interpreter.getContext().setThrowInterpreterErrors(true)` which will throw fatal errors rather than adding them to the error list, and ignore warning errors.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/expression/EagerExpressionStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/lib/expression/EagerExpressionStrategy.java
@@ -48,7 +48,11 @@ public class EagerExpressionStrategy implements ExpressionStrategy {
         : ""
     );
     if (chunkResolver.getDeferredWords().isEmpty()) {
-      String result = WhitespaceUtils.unquoteAndUnescape(resolvedExpression.getResult());
+      String fullyResolved = chunkResolver.resolveChunk(
+        resolvedExpression.getResult(),
+        ""
+      );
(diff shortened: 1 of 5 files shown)


boulter:
I hadn't thought of this. Will we lose any template warnings generated in the first pass? Those are only available in debug mode, so maybe that mode is always rendered on demand.

jasmith-hs:
If there is some desire to store them, the template warnings and errors can be stored as they are (aiming to be) the same from rendering with eager execution versus with default execution.

The current approach that we're using eager execution with at HubSpot does not preserve the rendering errors, so the only rendering errors that would still be around would be from the second pass only. This could be modified, but a better solution would probably be to render such debug requests on-demand to get a clear view of the warnings.

PR #591, opened 2021-01-27
[Eager Execution] Reserve null keyword in ChunkResolver

jasmith-hs:
Improves https://github.com/HubSpot/jinjava/issues/532
---

Since we don't ever want `null` to be put on the context as a deferred word, we can add it to the set of reserved words here. Additionally, to allow for some short-circuiting, we can add `true` and `false` here as well as we know these are words should be reserved from ever being deferred as well.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
+++ b/src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
@@ -265,7 +265,7 @@ public class ChunkResolver {
       String resolvedChunk;
       Object val = interpreter.resolveELExpression(chunk, token.getLineNumber());
       if (val == null) {
-        return JINJAVA_NULL;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ChunkResolver.java
```diff
@@ -265,7 +268,7 @@ private String resolveChunk(String chunk) {
       String resolvedChunk;
       Object val = interpreter.resolveELExpression(chunk, token.getLineNumber());
       if (val == null) {
-        return JINJAVA_NULL;
+        resolvedChunk = JINJAVA_NULL;
```
jasmith-hs, line 271: Fixes minor whitespace issue that was found from the test I added, where the space before a null value would be removed. Short-circuiting here prevented the space from being preserved, so this makes sure to preserve the space if it exists

PR #595, opened 2021-02-09
[Eager Execution] Don't prematurely resolve complex objects

jasmith-hs:
Fixes a diff that I found where an expression like `{{ foo.update({'bar': bar}) }}` was being resolved, except `bar` was a class that could not be serialised to a string while being able to retain the functionality. Basically, serializing and then deserializing `bar` would yield a much different object.

After consideration for how the chunk resolver aims to resolve as much as possible when there happens to be a deferred value, I modified it so that it will only do the premature evaluation of certain object classes: `String`, `Number`, `Boolean`, and `PyishSerializable`, as well as Collections and Maps of these. If there are other classes such as some arbitrary class `Person`, for example, then it will not be converted to a string. It will wait until the final evaluation call in the chunk resolver to actually resolve the value within an entire expression. With this, evaluating `{{ foo.update({'bar': bar}) }}` through the chunk resolver will result in `foo.bar` being equivalent to `bar`.

I also am making a change to DeferredValueUtils as with these changes, I exposed a minor bug that's fixed by just passing in the most recent `EagerToken` when deferring new values. This change prevents variables in the different context scopes, but with the same name both becoming deferred if one of them becomes deferred.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -333,7 +333,7 @@ public class Context extends ScopeMap<String, Object> {
 
   public void handleEagerToken(EagerToken eagerToken) {
     eagerTokens.add(eagerToken);
-    DeferredValueUtils.findAndMarkDeferredProperties(this);
+    DeferredValueUtils.findAndMarkDeferredProperties(this, eagerToken);
     if (getParent() != null) {
       Context parent = getParent();
       //Ignore global context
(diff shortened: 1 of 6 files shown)


Joeoh:
Have you given any thought to what to do if the same variable name is used in different scopes and both need to be deferred? I've seen this issue crop if eg. if someone uses the same module twice in a template

jasmith-hs:
If they both need to be deferred, then they'll both become deferred as the `handleEagerToken()` method is called. Could you provide an example of something that would cause the issue to which you're referring?

jasmith-hs:
> I also am making a change to DeferredValueUtils as with these changes, I exposed a minor bug that's fixed by just passing in the most recent EagerToken when deferring new values. This change prevents variables in the different context scopes, but with the same name both becoming deferred if one of them becomes deferred.

This should help address the issue that Joe mentioned he ran into. That issue has basically the same cause, which should be aided by this change in eager execution.

Also, the build got stuck in travis, but it is fine locally.

PR #599, opened 2021-02-17
Fix bug with whitespace controls not applying properly

mattcoley:
Whitespace controls operate on the assumption that textnodes cannot be adjacent. You can break up text nodes with a comment node `\n {# comment #} \n`. This should not affect the whitespace control but it does under the current code. In order to solve for this I merged adjacent text nodes. I did try a few other approaches over a couple of hours to try to get the same effect but could not solve for the case where there are more than 2 text nodes stacked next to each other.


jasmith-hs (approved):
This seems like a good solution. We essentially want the adjacent text tokens to be treated as the same token when handling whitespace, so merging them makes sense.

PR #604, opened 2021-02-26
[Eager Execution] Don't remove the import resource path from the context in EagerImportTag

jasmith-hs:
The `Context. IMPORT_RESOURCE_PATH_KEY` is used when doing relative path resolving within macro functions imported from other files so we don't want to remove it from the imported context.

The simple case is fixing this for Eager Execution where the macro is evaluated during the first pass. I just had to remove the lines that deleted that key from the context.

The more difficult case is when a macro function needs to be evaluated later, so we will need to somehow store the import resource path for that specific macro function in the eager execution output. This is done by including the value in the alias map that's output via set/do tags. Then when the macro tag is built later, if it uses an alias (ex: `var.something(1, 2)`), then we'll load the resource path for that macro function's context from the `var` alias map.


jasmith-hs:
I think I need to explore a case where a macro function is imported but does not have an alias

src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
```diff
@@ -124,24 +159,16 @@ public String reconstructImage() {
...
-      importResourceKey =
-        (String) (
-          (Map<String, Object>) ((DeferredValue) aliasMap).getOriginalValue()
-        ).get(Context.IMPORT_RESOURCE_PATH_KEY);
-      if (StringUtils.isEmpty(importResourceKey)) {
-        return "";
-      }
       PyMap deferredAliasMap = new PyMap(new HashMap<>());
-      deferredAliasMap.put(Context.IMPORT_RESOURCE_PATH_KEY, importResourceKey);
+      deferredAliasMap.put(Context.IMPORT_RESOURCE_PATH_KEY, importResourcePath);
```
Joeoh, line 171: Could this cause problems if theres multiple imports?

jasmith-hs: It shouldn't be an issue. The `deferredAliasMap` represents the whole object that is imported with an alias.
```
{% import 'file.html' as var %}
```
The `deferredAliasMap` will be put at `context.var`. A different import will have a separate map put onto the context with whatever key it's named with.

If there are multiple layers of imports, you would end up with something like:
```
context = {
  'var': {
    'import_resource_path': 'file.html',
    'nested': {
      'import_resource_path': 'nested.html',
      ... // other variables from nested.html
    },
    ... // other variables from file.html
  },
  ... // other variables on the context
}

PR #605, opened 2021-03-01
[Eager Execution] When importing and removing reference cycles, filter on equality

jasmith-hs:
Instead of naively removing those keys that match the current import alias, first check to make sure they are equal. We are trying to prevent reference cycles, so we can check if the references actually equal each other before removing the key from the child bindings.

Eager execution may result in an import alias deferring itself, so that's why it's necessary to prevent the reference cycle.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerImportTagTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerImportTagTest.java
@@ -436,6 +436,15 @@ public class EagerImportTagTest extends ImportTagTest {
     assertThat(interpreter.render(result).trim()).isEqualTo("12345 cbaabaaba");
   }
 
+  @Test
+  public void itImportsDoublyNamed() {
+    setupResourceLocator();
+    String result = interpreter.render(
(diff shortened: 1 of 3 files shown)


src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerImportTagTest.java
```diff
@@ -436,6 +436,15 @@ public void itDefersQuadLayer() {
     assertThat(interpreter.render(result).trim()).isEqualTo("12345 cbaabaaba");
   }
 
+  @Test
+  public void itImportsDoublyNamed() {
+    setupResourceLocator();
+    String result = interpreter.render(
+      "{% import 'variables.jinja' as foo %}{{ foo.foo['foo'].bar }}"
```
jasmith-hs, line 443: Before this change, `foo.foo` would be undefined, as we were removing the `foo` key from the child bindings to prevent a reference loop. In this case they're not referencing the same thing so it's fine to put the imported `foo` as a key within the `foo` alias.
(I know, it's a bit hard to explain which foo is which, but they need to have the same name for this bug)

PR #606, opened 2021-03-03
[Eager Execution] Improve whitespace handling in Chunk Resolver

jasmith-hs:
Jinja/Jinjava allow for some flexibility with whitespace and as the goal with Eager Execution is to have parity with default Jinjava functionality, we need to handle this flexible whitespace.

For example:
```
range (0, 1)
```
This needs to be interpreted as the range function, we do not want to split up the tokens with the whitespace between `range` and `(`. I've worked out that we should not be splitting tokens on whitespace, unless the whitespace precedes a letter or digit.
For example:
```
foo or bar
```
We can split using this whitespace so that the `foo` and `bar` can be resolved separately.


jasmith-hs:
Looks like I missed a case such as `{{ post . slug }}`

PR #608, opened 2021-03-04
Add function and filter to convert string to date

liamrharwood:
We currently have the `strtotime` function and filter that can be used to convert a string object into a `DateTime` object. However, we don't have a good way to do this for `Date` objects that don't include a time.

For instance, there's no good way to convert the string `3/4/21` to a different format, since it doesn't include the time.

This PR adds the new `strtodate` function/filter that can convert dates without time.

For instance, you can do something like this:
```
{{ "3/4/21"|strtodate('d/M/yy')|datetimeformat('%B %e, %Y') }}
```
To output the following:
`March 4, 2021`

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/StringToDateFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/StringToDateFilter.java
@@ -1,0 +1,36 @@
+package com.hubspot.jinjava.lib.filter;
+
+import com.hubspot.jinjava.interpret.InvalidInputException;
+import com.hubspot.jinjava.interpret.InvalidReason;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateSyntaxException;
+import com.hubspot.jinjava.lib.fn.Functions;
+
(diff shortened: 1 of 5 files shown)


boulter:
Should we call it `strptime`? That seems like the closest equivalent in python and what someone would use in Jinja.

https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

liamrharwood:
I'm not against it, but we already have `strtotime` so I thought `strtodate` would be consistent.

mattcoley:
Why did we name it `strtotime` in the first place?

mattcoley:
https://github.com/HubSpot/jinjava/pull/258 👎

src/main/java/com/hubspot/jinjava/lib/filter/StringToDateFilter.java
```diff
@@ -0,0 +1,51 @@
...
+        getName(),
+        "requires 1 argument (date format string)"
+      );
+    }
+
+    if (var == null) {
+      return null;
+    }
+
+    if (!(var instanceof String)) {
```
mattcoley, line 40: We could write the object to String here

mattcoley (approved):

PR #610, opened 2021-03-04
[Eager Execution] Support legacy dictionary creation

jasmith-hs:
Since Jinjava doesn't support variables names in dictionaries, we shouldn't support it in Eager Execution.

At the same time, it doesn't seem entirely right to make changes to this repo which aims to mirror Jinja that make it more different from Jinja without providing an alternative.

@boulter I have this idea to use a new config parameter `legacyFunctionality` to control whether Jinjava uses its legacy functionality or if it uses breaking functionality that closer mirrors Jinja functionality. This config would allow https://github.com/HubSpot/jinjava/pull/403 to be used if that flag is set to false. HubSpot would be using `legacyFunctionality=true`, but others may care more about having functionality that's like Jinja than supporting legacy functionality. There are other places where we've been unable to make changes to Jinjava as they would break existing templates, but a flag like this would give us freedom to fix such issues.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/LegacyOverrides.java
+++ b/src/main/java/com/hubspot/jinjava/LegacyOverrides.java
@@ -1,0 +1,55 @@
+package com.hubspot.jinjava;
+
+/**
+ * This class allows Jinjava to be configured to override legacy behaviour.
+ * LegacyOverrides.NONE signifies that none of the legacy functionality will be overridden.
+ */
+public class LegacyOverrides {
+  public static final LegacyOverrides NONE = new LegacyOverrides.Builder().build();
+  private final boolean evaluateMapKeys;
+  private final boolean iterateOverMapKeys;
+
+  private LegacyOverrides(Builder builder) {
+    evaluateMapKeys = builder.evaluateMapKeys;
+    iterateOverMapKeys = builder.iterateOverMapKeys;
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public boolean isEvaluateMapKeys() {
+    return evaluateMapKeys;
+  }
+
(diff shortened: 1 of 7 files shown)


boulter (changes requested):
I guess I would make the config more specific to this behavior rather than generic "legacy". What someone wants some of the legacy functionality, but their code breaks with others?

jasmith-hs:
So maybe an immutable that has flags for different legacy content that get overridden?
```
public class LegacyOverride {
  boolean legacyDictionaryKeys = true;
  ...
}
.
.
.
JinjavaConfig.withLegacyOverride(legacyOverrideBuilder.build());
```

boulter:
Jinjava doesn't use Immutables, but yeah, you could use the builder pattern to make a class like this. There may already be some other config options that could be classified as legacy behavior.

src/main/java/com/hubspot/jinjava/LegacyOverrides.java
```diff
@@ -0,0 +1,55 @@
+package com.hubspot.jinjava;
+
+/**
+ * This class allows Jinjava to be configured to override legacy behaviour.
+ * LegacyOverrides.NONE signifies that none of the legacy functionality will be overridden.
+ */
+public class LegacyOverrides {
+  public static final LegacyOverrides NONE = new LegacyOverrides.Builder().build();
+  private final boolean evaluateMapKeys;
+  private final boolean iterateOverMapKeys;
```
jasmith-hs, line 10: `evaluateMapKeys` and `iterateOverMapKeys` seem to be the most fitting so far as they are config options that allow for either preserving legacy functionality, or matching Jinja2 functionality

boulter (approved):

PR #612, opened 2021-03-08
[Eager Execution] Treat non-standard number representations as strings

jasmith-hs:
A phone number such as `+12223334444` would resolve as `12223334444` before this PR. This happened because when we converted that into a pyish object, we didn't surround it with quotes because it was interpretable as a double.

This PR makes it so that we write it a plus sign-prefixed string is written as a string, rather than a number. This means it is surrounded with quotes.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/PyishSerializer.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/PyishSerializer.java
@@ -28,7 +28,10 @@ public class PyishSerializer extends JsonSerializer<Object> {
     }
     try {
       Double.parseDouble(string);
-      if (string.length() > 1 && string.charAt(0) == '0' && string.indexOf('.') != 1) {
+      if (
+        string.length() > 1 &&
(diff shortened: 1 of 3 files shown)


boulter (commented):
It's really weird to have to think about phone number syntax here. Are there any other tokens that you think will have to be handled specially?

jasmith-hs:
It's when there's a non-standard representation of a number. Something that can be parsed as a number, but aren't the typical representation of it. I also had to do some handling with doubles that didn't start with a 0

src/main/java/com/hubspot/jinjava/objects/serialization/PyishSerializer.java
```diff
@@ -28,7 +28,10 @@ public void serialize(
     }
     try {
       Double.parseDouble(string);
-      if (string.length() > 1 && string.charAt(0) == '0' && string.indexOf('.') != 1) {
+      if (
+        string.length() > 1 &&
+        ((string.charAt(0) == '0' && string.indexOf('.') != 1) || string.charAt(0) == '+')
+      ) {
```
jasmith-hs, line 34: I think I might do a whitelisting approach instead using a regular expression

PR #616, opened 2021-03-11
Revert "[Eager Execution] Don't remove the import resource path from the context in EagerImportTag"

mattcoley:
Reverts HubSpot/jinjava#604

I added the test case that works before this commit.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
@@ -1,18 +1,13 @@
 package com.hubspot.jinjava.lib.fn.eager;
 
-import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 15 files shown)


jasmith-hs:
Ah, I was only looking at the import path stack for some logic, and not the `fromStack` so macros imported with `from`  had an isolated context. This caused them to not have macro functions from that same file registered...

I'm going to alter the approach I used to be safer by introducing a new variable that's only used within eager execution for this so that I don't cause side-effects to normal macro function behaviour. The logic I was adding was supposed to run a second rendering pass after a macro function was deferred in eager execution. This causes the function to get redefined, but on a single output source (Rather than being in a separate file) so there needed to be special logic to ensure that when it was created for a second time, it still had the same `import_resource_path`.

mattcoley:
@liamrharwood knows a lot about the different path stacks so worth tagging him on review.

PR #619, opened 2021-03-12
[Eager Execution] Add temporary override for preserveRawTags

jasmith-hs:
There may be scenarios where even if it's running in eager execution mode, the raw tags should be unwrapped. For instance, if there's a custom tag or function renders the expression nodes in a special manner causing the text inside the raw tags to not get output in the final result, but end up elsewhere, such as in the resulting context.

This adds the ability to use a try with resources to ensure that raw tags are unwrapped during such instances:
```
@Override
public String doInterpret(
  ...
  try (
      TemporaryValueClosable<Boolean> c = preserveInterpreter
        .getContext()
        .withUnwrapRawOverride()
    ) {
        customLogic(node);
  }
}
```
Since it's similar to the `withPartialMacroEvaluation()` auto-closable, I made a generic static class that both with-methods return.


src/test/resources/eager/handles-complex-raw.jinja
```diff
@@ -0,0 +1,10 @@
+{%- macro raw_exp(var) -%}
+{{ '{% raw %}}} ' ~ var ~ ' {{{% endraw %}' }}
+{%- endmacro -%}
+{%- macro raw_print(var) -%}
+{% print '}} ' ~ var ~ ' {{' %}
+{%- endmacro -%}
```
jasmith-hs, line 6: This is a difficult test to solve for so, but I want there to be parity between default and eager execution, so this should be handled. It's handled by re-rendering the result of a macro tag after it's output if raw tags are being preserved.

PR #626, opened 2021-03-29
[Eager Execution] Stricter number-writing logic

jasmith-hs:
Only write a value as a number if when converted to a number and then back to a string, it maintains the same value. The general theme of the diffs that popped up because of the logic here is that if there are multiple ways to represent the same value, then there will be a discrepancy if we try to write it as a number. So for these string values that cause a discrepancy, just write them as a string. We can check by converting back to a string and seeing if there is any change.


jboulter:
smart!

PR #632, opened 2021-04-06
[Eager Execution] Pass eager child context updates on to the parent

jasmith-hs:
When execution within a child context, if there are changes made to the context, then those changes should happen to the parent context as that is what would happen in default execution. This could happen in a custom function that when run in an expression node, adds or changes some value on the context.

This can be accomplished by copying over the child's session bindings. Changes to `__macros__` are ignored because it's a hidden context variable.


jasmith-hs:
In order to properly handle the context updates when a deferred value is encountered and we're evaluating within a deferred execution mode (such as when eagerly evaluating If-Else branches), we combine the session mappings with the prefix to preserve state. If we're running in a deferred execution mode, then we won't put the new changes on the context, and will instead convert them to a `set` tag. Otherwise, we can apply the new changes to the context as would happen with Default Jinjava execution.

PR #641, opened 2021-04-12
Support `not` operator without `is` keyword

### LINKED ISSUE #640: Missing `not in` operator
The following snippet is a valid Jinja expression but Jinjava fails to render since `not in` is not currently supported:

```
{% if '' not in [''] -%}
{% endif %}
```

```
2021-04-12 12:59:39 INFO stderr - 0 0 - 0 0 - com.hubspot.jinjava.interpret.FatalTemplateErrorsException: Error parsing ' '' not in [''] ': syntax error at position 5, encountered 'not', expected '}'
2021-04-12 12:59:39 INFO stderr - 0 0 - 0 0 - 	at com.hubspot.jinjava.Jinjava.render(Jinjava.java:191)
```

buremba:
Fixes #640

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -448,6 +448,12 @@ public class ExtendedParser extends Parser {
               v = createAstMethod(filterProperty, createAstParameters(filterParams)); // function("filter:" + filterName, new AstParameters(filterParams));
             } while ("|".equals(getToken().getImage()));
           } else if (
+            "not".equals(getToken().getImage()) &&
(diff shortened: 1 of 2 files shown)


jasmith-hs (changes requested):
It seems like the solution should be more centered around the `in` operator as if we are just allowing the `not`. Because if we support `{{ foo not in bar }}`, we should also support `{{ foo in bar }}`

In jinja, it seems like `in` can optionally be preceded by an `is`, but this is not the case with other expression tests. `{{ foo divisibleby 4 }}` is not valid syntax. So maybe add 2 branches here that address `in` and `not in`

jasmith-hs:
This was fixed in https://github.com/HubSpot/jinjava/pull/707

PR #658, opened 2021-04-29
add rangeLimit to JinjavaConfig (1000 by default)

ylacaute:
Changes: 
 - we need a JinjavaInterpreter instance to execute range function because we now retrieve the rangeLimit from the configuration instead of an hardcoded value
 - default range limit is still 1000 to avoid any breaking change
 - add 2 units test and update an existing one
 - mvn prettier:write ok

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
@@ -1,6 +1,8 @@
 package com.hubspot.jinjava.lib.fn;
 
+import static com.hubspot.jinjava.interpret.JinjavaInterpreter.getCurrent;
 import static com.hubspot.jinjava.util.Logging.ENGINE_LOG;
+import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.Lists;
 import com.hubspot.jinjava.JinjavaConfig;
@@ -34,7 +36,7 @@ public class Functions {
   public static final String STRING_TO_TIME_FUNCTION = "stringToTime";
   public static final String STRING_TO_DATE_FUNCTION = "stringToDate";
 
-  public static final int RANGE_LIMIT = 1000;
+  public static final int DEFAULT_RANGE_LIMIT = 1000;
 
   @JinjavaDoc(
     value = "Only usable within blocks, will render the contents of the parent block by calling super.",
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -1,6 +1,8 @@
 package com.hubspot.jinjava.lib.fn;
 
+import static com.hubspot.jinjava.interpret.JinjavaInterpreter.getCurrent;
```
boulter, line 3: any reason for the static import here? It's only used once.

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -386,15 +388,21 @@ public static int movePointerToJustBeforeLastWord(int pointer, String s) {
     "With two parameters, the range will start at the first value and increment by 1 up to (but not including) the second value. " +
     "The third parameter specifies the step increment. All values can be negative. Impossible ranges will return an empty list. " +
     "Ranges can generate a maximum of " +
-    RANGE_LIMIT +
-    " values.",
+    DEFAULT_RANGE_LIMIT +
+    " values by default, but this integer value is configurable.",
```
boulter, line 392: it is configurable, but not by the user which is who this documentation is targeted to.
```suggestion
    " values.",
```

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -386,15 +388,21 @@ public static int movePointerToJustBeforeLastWord(int pointer, String s) {
...
     params = {
       @JinjavaParam(value = "start", type = "number", defaultValue = "0"),
       @JinjavaParam(value = "end", type = "number"),
       @JinjavaParam(value = "step", type = "number", defaultValue = "1")
     }
   )
   public static List<Integer> range(Object arg1, Object... args) {
+    int rangeLimit = requireNonNull(
+        JinjavaInterpreter.getCurrent(),
```
boulter, line 401: I think it's only possible for this to be null in tests if you forget to push an interpreter onto the stack.

boulter (changes requested):
A few comments, otherwise looks good. Thanks!

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
ylacaute: Indeed, I remove it

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
ylacaute: I remove it, I just don't like when my IDE is screaming at me when something can be null.

boulter (approved):
LGTM. Thanks!

PR #664, opened 2021-05-12
[Eager Execution] Bring back context preserving

jasmith-hs:
With the `checkForContextChanges` flag, we can add back stringified context preserving without much overhead. This allows us to preserve changes to lists and maps if they are modified during `context.isDeferredExecutionMode()` by turning them into deferred values and pre-pending a `set` tag that assigns it to its original value.

Adding this allows us to unignore one of the tests, working on that allowed me to find a bug that the test now surfaced where expressions in `if` and `elif` tokens would be executed more than once (up to 3 times). This could cause issues if there were non-idempotent actions in those token expressions, and could cause slowness depending on how expensive those actions are. To fix this, I added logic to determine which branch the `DeferredParsingException` happened on and use the result within that expression as the new image for that token.

I also realised that the eager ast node parsing lacked the short-circuiting that is present in `AND` and `OR` binary operations, so I added a `DeferredELContext` which will defer any attempted evaluation after the left value in a binary expression if the left value is deferred. (Because we don't know if we'll need to short-circuit the operation so we need to defer the evaluation of what's on the right-hand side). Looking at in in action via `EagerAstBinaryTest.java`, helps explain why that's necessary.


jasmith-hs:
I tested this change on a very heavy render, and it causes a negligible slowdown in rendering time

PR #665, opened 2021-05-13
[Eager Execution] Prevent import aliases from leaking outside of their scope

jasmith-hs:
This fixes a bug where when using eager execution, import aliases could leak outside of the scope they were declared in due to calling `parent.getContext().get()` rather than `parent.getContext().getSessionBindings().get()`. If the parent had a variable present in its parent's context, then we would override that value rather than creating a new HashMap.

This can be demonstrated with an example with 3 files that will be imported:
1. `print-a.jinja`
```
{% macro print() %}
A
{% endmacro %}
```
2. `intermediate-b.jinja`
```
{% import 'print-b.jinja' %}
```
3. `print-b.jinja`
```
{% macro print() %}
B
{% endmacro %}
```
Rendering:
```
{% import 'print-a.jinja' as printer %}
{% import 'intermediate-b.jinja' as inter %}
{{ printer.print() }}
```
Should result in `A`, but the bug caused `printer` to be overridden as the import from `print-b.jinja`, and the turned the result to `B`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
@@ -140,11 +140,15 @@ public class MacroFunction extends AbstractCallableMethod {
     LengthLimitingStringBuilder result = new LengthLimitingStringBuilder(
       interpreter.getConfig().getMaxOutputSize()
     );
+    int initialEagerTokenCount = interpreter.getContext().getEagerTokens().size();
 
     for (Node node : content) {
       result.append(node.render(interpreter));
     }
-    if (interpreter.getConfig().getExecutionMode().isPreserveRawTags()) {
+    if (
+      interpreter.getConfig().getExecutionMode().isPreserveRawTags() &&
+      initialEagerTokenCount == interpreter.getContext().getEagerTokens().size()
+    ) {
       try (
         TemporaryValueClosable<Boolean> c = interpreter
           .getContext()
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
```diff
@@ -140,11 +140,15 @@ public String getEvaluationResult(
...
+    int initialEagerTokenCount = interpreter.getContext().getEagerTokens().size();
 
     for (Node node : content) {
       result.append(node.render(interpreter));
     }
-    if (interpreter.getConfig().getExecutionMode().isPreserveRawTags()) {
+    if (
+      interpreter.getConfig().getExecutionMode().isPreserveRawTags() &&
+      initialEagerTokenCount == interpreter.getContext().getEagerTokens().size()
+    ) {
```
jasmith-hs, line 151: Unrelated, but saw in my testing, don't need to unwrap the raw tags from a macro function if it has deferred tokens within it.

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerImportTag.java
```diff
@@ -108,20 +108,20 @@ private static String getDoTagToPreserve(
...
-        .getContext()
-        .put(currentImportAlias, DeferredValue.instance(currentAliasMap));
+      currentAliasMap = DeferredValue.instance(currentAliasMap);
+      interpreter.getContext().put(currentImportAlias, currentAliasMap);
     }
     for (Map.Entry<String, Object> entry : (
-      (Map<String, Object>) (
-        (DeferredValue) interpreter.getContext().get(currentImportAlias)
-      ).getOriginalValue()
+      (Map<String, Object>) ((DeferredValue) currentAliasMap).getOriginalValue()
```
jasmith-hs, line 124: Just refactoring this by storing a variable to make it a bit cleaner

PR #667, opened 2021-05-19
Track correct path for multiple level of Extends inheritance

bkrainer:
We're not tracking the currentPath correctly when dealing with multiple levels of inheritance via `extends` tags, which is leading to incorrect relative path resolution.

Prior to this PR, the test example: https://github.com/HubSpot/jinjava/pull/667/files#diff-2970b2d908f3ae9b6918c9b1bd9c7b678a92a8a054cb2c9bded5817bd4a41a9eR1

would resolve `./hello.html` to `/tags/extendstag/relative/hello.html` because it was incorrectly using `tags/extendstag/relative/nested-relative-extends-2.jinja` as the current path.


jasmith-hs (approved):
LGTM

PR #678, opened 2021-06-03
Use interpreter config timezone as default instead of UTC

mattcoley:
The interpreter config allows you to set a timezone, however we are not using that config value in our datetime functions leading to unexpected datetimeformat output.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
@@ -141,7 +141,7 @@ public class Functions {
       );
     }
     ZoneId zoneOffset = getDefaultZoneId();
-    if (var.length > 0) {
+    if (var.length > 0 && var[0] != null) {
       String timezone = var[0];
       try {
         zoneOffset = ZoneId.of(timezone);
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -326,11 +326,20 @@ public static PyishDate stringToDate(String dateString, String dateFormat) {
...
       String convertedFormat = StrftimeFormatter.toJavaDateTimeFormat(dateFormat);
-      return new PyishDate(
+      ZonedDateTime.of(
         LocalDate
           .parse(dateString, DateTimeFormatter.ofPattern(convertedFormat))
-          .atTime(0, 0)
-          .toInstant(ZoneOffset.UTC)
+          .atTime(0, 0),
+        getDefaultZoneId()
+      );
```
jasmith-hs, line 334: I think you didn't mean to include this block.

jasmith-hs (approved):

src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
mattcoley: whoops 😄

PR #679, opened 2021-06-03
Ensure lazy expressions that supply null don't cause template errors

jmaroeder:
In the current code, if a LazyExpression supplies `null`, then `JinjavaInterpreterResolver#wrap` generates a NullPointerException which becomes a TemplateError due to calling `getClass()` on `null` post-resolution. This prevents that from happening.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java
@@ -294,6 +294,9 @@ public class JinjavaInterpreterResolver extends SimpleResolver {
 
     if (value instanceof LazyExpression) {
       value = ((LazyExpression) value).get();
+      if (value == null) {
+        return null;
+      }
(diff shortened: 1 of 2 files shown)


jasmith-hs:
Could this happen here too https://github.com/HubSpot/jinjava/blob/15d8154f6267dafc7fc114f96c6ee31d5f526e14/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java#L207-L220

jmaroeder:
> Could this happen here too
> 
> https://github.com/HubSpot/jinjava/blob/15d8154f6267dafc7fc114f96c6ee31d5f526e14/src/main/java/com/hubspot/jinjava/el/JinjavaInterpreterResolver.java#L207-L220

@jasmith-hs I think so, and also below it. I'll do a quick search to see if it's possible anywhere else LazyExpressions are get()'d and update this PR

jasmith-hs (approved):

PR #688, opened 2021-06-17
Add new MapELResolver with type coercion to support accessing enum keys

liamrharwood:
related to #271 

This PR adds the new `TypeConvertingMapELResolver` in order to allow accessing map entries when their key does not have equality with its string representation.

This is most often seen with enums. For instance, imagine a Java map with key type `MyEnum`:
```
public enum MyEnum { FOO, BAR; } 
```
Outputting the map via `{{ map }}` could produce the following:
```
{FOO=value1, BAR=value2}
```
However, `{{ map.FOO }}` would not resolve to anything as the string `"FOO"` does not have equality with `MyEnum.FOO`.

To fix this, I've used the `CollectionMembershipOperator` and `TruthyTypeConverter` to determine if a key is valid and convert it to the correct class. This means that this should work for all types that have this issue, not just enums (for instance, numeric types).

After this PR, `{{ map.FOO }}` will output `value1`, which is the expected result.


src/main/java/com/hubspot/jinjava/el/TypeConvertingMapELResolver.java
```diff
@@ -0,0 +1,36 @@
...
+  public Object getValue(ELContext context, Object base, Object property) {
+    Object value = super.getValue(context, base, property);
+
+    if (value != null) {
+      return value;
+    }
+
+    if (
+      base instanceof Map &&
+      (Boolean) CollectionMembershipOperator.OP.apply(TYPE_CONVERTER, property, base)
```
mattcoley, line 25: What does this do separate from the `TYPE_CONVERTER.convert` call later on?

src/main/java/com/hubspot/jinjava/el/TypeConvertingMapELResolver.java
```diff
@@ -0,0 +1,36 @@
...
+    ) {
+      Class<?> keyClass = ((Map) base).keySet().iterator().next().getClass();
```
mattcoley, line 27: I think we need to check if the map is not empty to avoid NPE if we get rid of the CollectionMembershipOperator line above.

mattcoley (changes requested):

src/main/java/com/hubspot/jinjava/el/TypeConvertingMapELResolver.java
liamrharwood: It basically makes sure we will be able to convert to the correct type. 
Otherwise it will add an `ELException` to the interpreter errors: `Cannot coerce 'blah' of class java.lang.String to class {enum}`.

mattcoley: Ah so the empty map will be covered there.

mattcoley: It would be worth just adding a test case with an empty map to cover that then LGTM

mattcoley (approved):

src/main/java/com/hubspot/jinjava/el/TruthyTypeConverter.java
```diff
@@ -103,7 +104,20 @@ protected String coerceToString(Object value) {
         return enumSet.iterator().next();
       }
     }
-    return super.coerceToEnum(value, type);
+
+    try {
+      return super.coerceToEnum(value, type);
+    } catch (ELException e) {
+      if (value instanceof String) {
```
liamrharwood, line 111: @mattcoley Since jinjava actually uses the `toString` representation when printing enums, `Enum.valueOf` in the regular `coerceToEnum` will not always work. For that reason, I added extra logic here to check the `toString` values of the enums as well.
You can see an example in the `itGetsDictValuesWithEnumKeysUsingToString` test case below.

bkrainer (approved):

mattcoley (approved):

PR #689, opened 2021-06-17
[Eager Execution] Skip unnecessary downstream deferring

jasmith-hs:
If there is a named parameter used in something like a macro function, then we don't want to mark the key as a deferred value.
```
{% macro func(foo) %}
{{ foo }}
{% endmacro %}
{% set foo = 'val' %}
{{ func(foo=deferred) }}
{{ foo }}
```
Since foo's usage in the named parameter is not the same as the variable, we shouldn't mark it as something to be deferred. It should keep its value of `val`. Additionally, we make sure that when registering an `EagerToken`, we only send `usedDeferredWords` that have not been deferred already so that we don't end up marking the `foo` used in the macro function as a DeferredValue throughout the whole context tree. Doing so would cause the second `{{ foo }}` to not be evaluated, when it clearly can be evaluated to `val`.


src/test/java/com/hubspot/jinjava/EagerTest.java
```diff
@@ -478,7 +478,7 @@ public void itEvaluatesNonEagerSet() {
           .flatMap(eagerToken -> eagerToken.getUsedDeferredWords().stream())
           .collect(Collectors.toSet())
       )
-      .contains("deferred");
+      .isEmpty();
```
jasmith-hs, line 481: These test changes are because now we only add words to `usedDeferredWords` if they aren't already `DeferredValues`. Here, `deferred` is already a deferred value so it won't be present in `usedDeferredWords`.

PR #692, opened 2021-06-22
[Eager Execution] Execute deferred branches of if tag in separate contexts

jasmith-hs:
As the name suggests, when speculatively evaluating if tag branches, use separate contexts for each branch to maximize the values that can be resolved.
```
{% set foo = 1 %}
{% if deferred %}
{% set foo = 2 %}
{% else %}
{{ foo }}
{% endif %}
```
This change allows us to know that `foo=1` when evaluating `{{ foo }}`, rather than have it already be deferred. Since we set deferred values in a waterfall-like fashion, we can temporarily override a deferred value by setting the value on the lowest child context (it will still be deferred on the parent contexts).

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
@@ -133,6 +133,21 @@ public class EagerSetTag extends EagerStateChangingTag<SetTag> {
         )
       );
     }
+    if (
+      eagerExecutionResult.getResult().isFullyResolved() &&
+      interpreter.getContext().isDeferredExecutionMode()
+    ) {
+      try {
+        getTag()
+          .executeSet(
+            tagToken,
+            interpreter,
+            varTokens,
+            eagerExecutionResult.getResult().toList(),
+            true
+          );
+      } catch (DeferredValueException ignored) {}
+    }
     return wrapInAutoEscapeIfNeeded(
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTag.java
```diff
@@ -133,6 +133,21 @@ public String getEagerTagImage(TagToken tagToken, JinjavaInterpreter interpreter
         )
       );
     }
+    if (
+      eagerExecutionResult.getResult().isFullyResolved() &&
+      interpreter.getContext().isDeferredExecutionMode()
+    ) {
+      try {
+        getTag()
+          .executeSet(
```
jasmith-hs, line 142: Try to execute the set after waterfall-deferring if it is fully resolvable so that we can use the value during this context

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
```diff
@@ -267,7 +267,13 @@ public static EagerExecutionResult executeInChildContext(
         .entrySet()
         .stream()
         .filter(
-          entry -> !entry.getValue().equals(interpreter.getContext().get(entry.getKey()))
+          entry ->
+            entry.getValue() != null &&
```
jasmith-hs, line 271: This fixes a NPE

Issue #693, opened 2021-06-23
set tag doesn't work correctly, endset is not known for javajinja


dimmonn:
the below setup doesn't work while it's expected to

    {% for item in items %}
    {% if item['some_prop_1'] and item['some_prop_1'] != "" %}
    {% set someAttribute %}{{item['some_prop_1']}}{% endset %}
    {% else %}

TemplateError{severity=FATAL, reason=SYNTAX_ERROR, item=OTHER, message='Error parsing '[{{item['some_prop_1']}} ]': syntax error at position 36, encountered '}', expected ':'', fieldName='}} ]', lineno=12, startPosition=36, scopeDepth=2, category=UNKNOWN, categoryErrors=null}

mattcoley:
You should be using the set tag like `{% set someAttribute = item['some_prop_1'] %}`

dimmonn:
Right but jinja supports the other way too that works with endset tagIn my case i can't just change because i have many templates like that already writtenSo would be good to be compliant with jinjaOn Jun 23, 2021 15:38, Matt Coley ***@***.***> wrote:
You should be using the set tag like {% set someAttribute = item['some_prop_1'] %}

jasmith-hs:
This functionality has been added in now

Issue #422, opened 2020-04-09
Support block assignments


buremba:
Jinja2 supports for the following syntax out of the box:
```
{% set variable %} 
  value of the variable
{% endset %}
```

It would be nice to have support for it in jinjava as well. Please see [the docs here](https://jinja.palletsprojects.com/en/2.11.x/templates/#block-assignments).

bhreinb:
+1 - for this construct too as it's something I would use too.

jasmith-hs:
This functionality has been added in now

PR #698, opened 2021-06-25
Support Assignment Blocks with Set tags

jasmith-hs:
For https://github.com/HubSpot/jinjava/issues/693 and https://github.com/HubSpot/jinjava/issues/422
Jinja 2.8 adds support for block set assignments (https://jinja.palletsprojects.com/en/3.0.x/templates/#block-assignments)

Add support for set tags with the notation:
```
{% set foo %}
A value of {{ 1 + 1 }}
{% endset %}
```
`foo = 'A value of 2'`

Additionally, Jinja 2.10 adds support for filters in the assignments:
```
{% set foo | upper %}
hello
{% endset %}
```
`foo = 'HELLO'`.

---
The `FlexibleTag` interface adds a way to determine Node-based eligibility for requiring an `EndTag` rather than Tag-based eligibility through the `hasEndTag(TagToken tagToken)` method. Since this method takes a `TagToken` as the parameter, whether or not the tag will have a corresponding endtag depends on what is in the TagToken. For `SetTag`, it depends on whether there is an `=` in the helpers.

I'm opening this PR separately from the logic to handle this change with Eager Execution because those changes are much larger.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
@@ -148,17 +148,20 @@ public class SetTag implements Tag, FlexibleTag {
         Collections.singletonList(sb.toString()),
         false
       );
-      Object finalVal = interpreter.resolveELExpression(
-        tagNode.getHelpers().trim(),
-        tagNode.getMaster().getLineNumber()
-      );
-      executeSet(
-        (TagToken) tagNode.getMaster(),
-        interpreter,
-        varAsArray,
-        Collections.singletonList(finalVal),
-        false
-      );
+      if (filterPos >= 0) {
+        // apply and save the filtered result
+        Object finalVal = interpreter.resolveELExpression(
+          tagNode.getHelpers().trim(),
+          tagNode.getMaster().getLineNumber()
+        );
+        executeSet(
+          (TagToken) tagNode.getMaster(),
+          interpreter,
+          varAsArray,
+          Collections.singletonList(finalVal),
+          false
+        );
+      }
     } catch (DeferredValueException e) {
       DeferredValueUtils.deferVariables(varAsArray, interpreter.getContext());
       throw e;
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
```diff
@@ -66,7 +68,7 @@
   }
 )
 @JinjavaTextMateSnippet(code = "{% set ${1:var} = ${2:expr} %}")
```
jasmith-hs, line 70: Snippets should probably be adjusted

src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerSetTagTest.java
```diff
@@ -122,4 +122,16 @@ public void itThrowsAndDefersVarWhenValContainsDeferred() {
...
+  @Ignore
+  public void itSetsBlock() {
+    // not implemented yet
+  }
+
+  @Test
+  @Ignore
+  public void itSetsBlockWithFilter() {
+    // not implemented yet
+  }
```
jasmith-hs, line 136: It is complex to properly support this with eager execution so it's in a separate PR https://github.com/HubSpot/jinjava/pull/699

src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
```diff
@@ -126,6 +129,43 @@ public String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
...
+        tagNode.getHelpers().trim(),
+        tagNode.getMaster().getLineNumber()
+      );
+      executeSet(
+        (TagToken) tagNode.getMaster(),
+        interpreter,
+        varAsArray,
+        Collections.singletonList(finalVal),
+        false
+      );
```
jmaroeder, line 161: Why is `executeSet` executed twice? It looks like this would allow for variable variable names, but I wasn't aware that was a supported/necessary feature of jinjava. Is there another reason?

jmaroeder (approved):
Looks good, one question for my understanding!

src/main/java/com/hubspot/jinjava/lib/tag/SetTag.java
jasmith-hs: Good question, I should probably clarify this in the code, but Jinja supports the usage of filters within set blocks like:
```
{% set foo | upper %}
Some text
{% endset %}
```
`foo == 'SOME TEXT'`.
It is easy to do this by first saving `'Some text'` to foo, and then evaluating `foo | upper` and setting the result of that expression to `foo`

Issue #511, opened 2020-09-24
Global autoescape not propagated to other contexts


thusoy:
I wanted to turn on global autoescape for a `Jinjava` instance and thought this should have worked:

```java
Jinjava jinjava = new Jinjava();
jinjava.getGlobalContext().setAutoEscape(true);
assertEquals("&lt;b&gt;", jinjava.render("{{ unsafe }}", Collections.singletonMap("unsafe", "<b>")));
```
However it seems the `autoescape` property isn't transferred into the `Context` actually used for the rendering, as nothing is escaped. I can work around by this wrapping the templates we render in an `{% autoescape %}` block, but not as elegant as having the property work out of the box.

Joeoh:
Hi @thusoy , try using renderForHtml method instead

Joeoh:
You need to configure the interpreter with this Jinjava config, so try something like this: 

```
         renderContext.setAutoEscape(true);
        
        JinjavaInterpreter interpreter = new JinjavaInterpreter(
          jinjava,
          renderContext,
          jinjava.getGlobalConfig()
        );
        String output = interpreter.render(templateSource);
```

thusoy:
Okay, I got that working, but this seems to be an awful lot of code for what seems to be a simple config change and what jinja does by default?

```java
    @Test
    public void testRenderAutoescape() {
        var jinjava = new Jinjava();
        var context = jinjava.getGlobalContextCopy();
        context.putAll(Map.of("unsafe", "<b>"));
        context.setAutoEscape(true);
        var config = JinjavaConfig.newBuilder()
                .build();
        var interpreter = new JinjavaInterpreter(
                jinjava,
                context,
                config
        );
        var template = "{{ unsafe }}";
        var rendered = interpreter.render(template);
        var fatalErrors = interpreter.getErrors().stream()
                .filter((error) -> error.getSeverity() == TemplateError.ErrorType.FATAL)
                .collect(Collectors.toList());

        if (!fatalErrors.isEmpty()) {
            throw new FatalTemplateErrorsException(template, fatalErrors);
        }
        assertEquals("&lt;b&gt;", rendered);
    }
```

PR #700, opened 2021-06-25
Propogate Global AutoEscape

frankliu197:
This pull request is to solve the following issue:
https://github.com/HubSpot/jinjava/issues/511

The pull request is to allow the code: ```jinjava.getGlobalContext().setAutoEscape(true);``` to turn on global autoscaping.

e.g.
```
@Test
    public void AutoEscapeTest() {
        Jinjava jinjava = new Jinjava();
        jinjava.getGlobalContext().setAutoEscape(true);
        assertThat("&lt;b&gt;").isEqualTo(jinjava.render("{{ unsafe }}", Collections.singletonMap("unsafe", "<b>")));
    }
```
@Joeoh for previously dealing with the issue linked to this pull request

### DIFF
--- a/src/main/java/com/hubspot/jinjava/Jinjava.java
+++ b/src/main/java/com/hubspot/jinjava/Jinjava.java
@@ -326,6 +326,7 @@ public class Jinjava {
     globalContext.getAllFilters().forEach(context::registerFilter);
     globalContext.getAllFunctions().forEach(context::registerFunction);
     globalContext.getAllTags().forEach(context::registerTag);
+    context.setAutoEscape(globalContext.isAutoEscape());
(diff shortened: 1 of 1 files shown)


macleodbroad-wf:
This looks good to me.

Joeoh (approved):
Thank you!

macleodbroad-wf:
I now think this is problematic for macros that include html tags

jasmith-hs:
Problematic in what way? In that they are not escaped?

macleodbroad-wf:
If a macro returns html tags, they are escaped as well.
I am comparing the behaviour of this library to Jinja2 with the `jinja2.ext.autoescape` extension.

Issue #705, opened 2021-06-30
1st argument must be iterable


Gab0o777:
Given issue #640, the following Jinja expression fails to render
```java
{% if VAR1 is defined and VAR1 not in VAR2 %}
```
I'm trying to migrate a lot of templates like this to a syntax Jinjava can render but when trying the following, I still get an error about the second variable not being iterable, since it's a string.

```java
{% if VAR1 is defined and VAR1 is not in VAR2 %}
```

```
2021-06-30 12:53:06.835  INFO -- [p2044211046-117] c.m.p.beans.TemplateValidationBean       : [TemplateError{severity=FATAL, reason=INVALID_ARGUMENT, item=PROPERTY, message='Invalid argument in 'in': 1st argument with value 'SDSQXCG-128G-GN6MA' must be iterable', fieldName='in', lineno=1, startPosition=29, scopeDepth=1, category=UNKNOWN, categoryErrors=null}]
```

is there another way to check if a string contains another?

PR #707, opened 2021-07-13
Support "not in"

jasmith-hs:
Fixes https://github.com/HubSpot/jinjava/issues/640 and https://github.com/HubSpot/jinjava/issues/705

We currently have support for the `in` operator, which handles collection membership in strings, lists, and maps, but we don't support the negation of this operation yet in jinjava. The negation (`not in`) is supported in Jinja, and the closest we have to this is by using `is not in`. This, however does not support strings as they are not iterable. The bug in 705 is caused by a faulty check that prevents the logic from working on strings.

The approach I went with overrides the `cmp` method, adding logic to the default case where if the current symbol is `not` and the next one is `in`, it uses a new CollectionNonMembershipOperator that returns the opposite of CollectionMembershipOperator.

"not in" is unique in Jinja https://jinja.palletsprojects.com/en/3.0.x/templates/#logic which is why I am handling it with a new operator, rather than doing a generic implementation with `not` before any comparison operator.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -461,6 +461,13 @@ public class ExtendedParser extends Parser {
           ) {
             consumeToken(); // 'is'
             v = buildAstMethodForIdentifier(v, "evaluate");
+          } else if ("in".equals(getToken().getImage())) {
+            v = buildAstMethodForIdentifier(v, "evaluate");
+          } else if (
+            "not".equals(getToken().getImage()) && "in".equals(lookahead(0).getImage())
+          ) {
+            consumeToken(); // 'not'
+            v = buildAstMethodForIdentifier(v, "evaluateNegated");
           }
 
           return v;
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -107,7 +107,7 @@ public class ExtendedParser extends Parser {
     );
     putExtensionHandler(
       CollectionNonMembershipOperator.TOKEN,
-      CollectionNonMembershipOperator.getHandler(false)
+      CollectionNonMembershipOperator.HANDLER
     );
 
     putExtensionHandler(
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -104,6 +105,10 @@ public ExtendedParser(Builder context, String input) {
       CollectionMembershipOperator.TOKEN,
       CollectionMembershipOperator.HANDLER
     );
+    putExtensionHandler(
+      CollectionNonMembershipOperator.TOKEN,
+      CollectionNonMembershipOperator.getHandler(false)
```
jmaroeder, line 110: Suggest `CollectionNonMembershipOperator.HANDLER` instead to match the other putExtensionHandler calls

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -344,6 +349,51 @@ protected AstNode literal() throws ScanException, ParseException {
...
+        case GT:
+          consumeToken();
+          v = createAstBinary(v, add(true), AstBinary.GT);
+          break;
+        case EXTENSION:
+          if (getExtensionHandler(getToken()).getExtensionPoint() == ExtensionPoint.CMP) {
+            v = getExtensionHandler(consumeToken()).createAstNode(v, add(true));
+            break;
+          }
+        default:
```
jmaroeder, line 381: This has a smell to me - I haven't spent any time in this parser code, but I don't like duplicating this much code to make the override work. Is it possible to create a two-word token ("not in")?

Another kernel of thought in how we approach this might be to rewrite `foo not in bar` to `not (foo in bar)`, but that has a different smell. Just sharing it here in case that gets you thinking, though

jmaroeder (changes requested):

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
jasmith-hs: It does have a bit of a smell, but the original Parser code doesn't provide more options for extending here. Because of the call to `add()` there isn't a way to use `super` calls and add on to the original code. Seems we've had to do this duplication in the past, for example with the `value()` method. I don't think there's a way that we can have a two-word token unfortunately because of the way that tokens are scanned.

While I like the idea of simply `not (foo in bar)`, to get parity here with Jinja, I want to add functionality in Jinjava for:
> The is and in operators support negation using an infix notation, too: `foo is not bar` and `foo not in bar`

jmaroeder (approved):

PR #719, opened 2021-07-26
Limit infinite evaluation from recursive extends tags

jasmith-hs:
Prevent infinite levels of extends tags being called when there is an extended child that has a child that extends itself. The basic stack way of detecting extends tag cycles did not catch this case. So this PR adds in a set that remembers the paths that have been extended when processing them and if any ever get encountered more than once, it will break and add an interpreter error.

The test case that I added is a simple demonstration of what caused infinite execution. This PR fixes that case.

cc @kacperadach @hs-lsong @bkrainer


jmaroeder (approved):
Looks good to me

PR #745, opened 2021-09-09
[Eager Execution] Reconstruct ignored extends output

jasmith-hs:
There is an issue that happens when a child page extends another page and that child page has deferred tokens that are outside any of the `{% block %}` that the child is overriding. Because all the output, save for that within a block, from a child normally gets discarded, if there are any reconstructed tokens that have been deferred, they will be lost.

In order to preserve the functionality of those reconstructed tokens, we must keep them in the output for evaluation at a later time, but we don't want the result of them to be included in any of the output. To accomplish this, we can wrap them in a block set tag whose name is something we don't care about. All this will do is interpret the code within the set block, and then save it to an arbitrary name, thus preventing any of it from being included in the final output, but still modifying the final context as necessary.


src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -385,15 +385,83 @@ public static String buildSetTag(
...
     JinjavaInterpreter interpreter
   ) {
+    Map<Library, Set<String>> disabled = interpreter.getConfig().getDisabled();
+    if (
+      disabled != null &&
+      disabled.containsKey(Library.TAG) &&
+      disabled.get(Library.TAG).contains(DoTag.TAG_NAME)
+    ) {
+      throw new DisabledException("do tag disabled");
+    }
```
jasmith-hs, line 460: Just adding this here for consistency

PR #747, opened 2021-09-16
Create interface for object truth values

mattcoley:
Allows an object to determine its truth value for resolving expressions like `{% if object %}`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
+++ b/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
@@ -29,6 +29,11 @@ public final class ObjectTruthValue {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
```diff
@@ -29,6 +29,11 @@ public static boolean evaluate(Object object) {
       return false;
     }
 
+    if (object instanceof HasObjectTruthValue) {
+      HasObjectTruthValue value = (HasObjectTruthValue) object;
+      return value.getObjectTruthValue();
```
jasmith-hs, line 34:
```suggestion
      return ((HasObjectTruthValue) object).getObjectTruthValue();
```

jasmith-hs (approved):

PR #750, opened 2021-09-23
Catch concurrent modification in for loop

mattcoley:
Catches `ConcurrentModificationException` that can happen in the for loop and adds an FATAL template error.


jasmith-hs (approved):
I think this is better than reworking how for loops are processed. At least for HubSpot

PR #751, opened 2021-09-23
[Eager Execution] Fix deferred import selection

jasmith-hs:
This PR handles the case where an import within deferred execution mode didn't work because we might not reconstruct the map.
This is handled by making sure that `{% set alias = {} %}` is called before the corresponding `{% do alias.update({...}) %}` calls are made to update the import alias.


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerImportTag.java
```diff
@@ -152,14 +157,6 @@ private static String getDoTagToPreserve(
...
       .getSessionBindings()
       .get(currentImportAlias);
-    if ((!(currentAliasMap instanceof DeferredValue))) {
-      // Make sure that the map is deferred.
-      if (!(currentAliasMap instanceof Map)) {
-        currentAliasMap = new PyMap(new HashMap<>());
-      }
-      currentAliasMap = DeferredValue.instance(currentAliasMap);
-      interpreter.getContext().put(currentImportAlias, currentAliasMap);
-    }
```
jasmith-hs, line 162: This is no longer needed because we are always deferring it beforehand with lines 104-108

PR #754, opened 2021-09-24
Throw a template error when attempting to divide by zero

jmaroeder:
Rather than throwing an uncaught ArithmeticException, this makes Jinjava throw a template error which can be handled like all other Jinjava template errors.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
@@ -74,41 +74,51 @@ public class DivideFilter implements Filter {
     } else {
       return object;
     }
-    if (object instanceof Integer) {
-      return (Integer) object / num.intValue();
-    }
-    if (object instanceof Float) {
-      return (Float) object / num.floatValue();
-    }
-    if (object instanceof Long) {
-      return (Long) object / num.longValue();
-    }
-    if (object instanceof Short) {
-      return (Short) object / num.shortValue();
-    }
-    if (object instanceof Double) {
-      return (Double) object / num.doubleValue();
-    }
-    if (object instanceof BigDecimal) {
-      return ((BigDecimal) object).divide(BigDecimal.valueOf(num.doubleValue()));
-    }
-    if (object instanceof BigInteger) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/DivideFilter.java
```diff
@@ -74,41 +74,51 @@ public Object filter(Object object, JinjavaInterpreter interpreter, String... ar
...
-      try {
-        return Double.valueOf((String) object) / num.doubleValue();
-      } catch (NumberFormatException e) {
-        throw new InvalidInputException(
-          interpreter,
-          this,
-          InvalidReason.NUMBER_FORMAT,
-          object.toString()
-        );
+    try {
```
jasmith-hs, line 77: Can we check that `num` is not zero instead?

jmaroeder: I'm open to doing this instead of the current approach, but I'm curious about your reasoning - are you trying to be more performant, or more exact, or something else?

My usual philosophy is to rely upon the existing exceptions as much as possible rather than duplicating logic that exists in the standard library

jasmith-hs: I was extrapolating the idea of using null checks rather than catching NullPointerException into checking for 0 rather than catching ArithmeticException, but it seems that catching ArithmeticException is an acceptable way to handle the divide by zero according to https://wiki.sei.cmu.edu/confluence/display/java/ERR08-J.+Do+not+catch+NullPointerException+or+any+of+its+ancestors

jmaroeder: Ah, yes, I can follow that logic! Yeah, I wouldn't catch NullPointerException or (heaven forbid) Exception, but with the specificity of ArithmeticException I would lean toward using it

boulter: Exceptions have significant overhead, so if you can write a simple check that avoids throwing and catching them, you'll save yourself a little performance overhead. That said, it likely doesn't matter unless you're throwing hundreds or thousands of exceptions.

PR #756, opened 2021-09-29
Unit test for escaped characters

mattcoley:


Joeoh:
Is it worth testing this when auto escape is enabled too?

mattcoley:
The test fails, this is to show how an escaped Unicode character in a string can break parsing.

PR #758, opened 2021-09-30
handle null argument in range function

boulter:
Passing a null argument to the range function would throw a NullPointerException instead of being handled as an interpret error. That's fixed now.

I also logged template errors for bad values in the truncate and truncateHtml functions instead of using the ENGINE_LOG which no one looks at.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
@@ -489,19 +489,19 @@ public class Functions {
 
     switch (args.length) {
       case 0:
-        if (NumberUtils.isNumber(arg1.toString())) {
+        if (NumberUtils.isCreatable(arg1.toString())) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/Functions.java
```diff
@@ -461,21 +479,29 @@ public static int movePointerToJustBeforeLastWord(int pointer, String s) {
...
+        JinjavaInterpreter.getCurrent(),
+        "range",
+        "Invalid null passed to range function"
+      );
+    }
+
     switch (args.length) {
       case 0:
-        if (NumberUtils.isNumber(arg1.toString())) {
+        if (NumberUtils.isCreatable(arg1.toString())) {
```
boulter, line 492: deprecated function

jasmith-hs (approved):
Similarly, the `ENGINE_LOG` is used in the TruncateHtmlFilter. It would be nice to have that add an error too.

PR #760, opened 2021-10-04
Track Current Processed Node In The Context

bhreinb:
Hi there,

I'd like to provide more detail when an exception is fired as a result of processing a macro block. The use case is something like the following:

I have a main template file that has the following snippet:
```
     {% import 'com/<companyName>/macros/translate.jinja' as macro %}

     {{macro.translateWord("hello", "de")}}  
     {{macro.translateWord("world", "es")}}
```

the macro has something like the following:
```
{%- macro translateWord(word, languageIso) %}
       {% set supportedIsoLanguages = ["en", "de", "fr"] %}
       {{ assert:contains(languageIso, supportedIsoLanguages) }}
       {{ util:translateString(word, languageIso) }}
{% endmacro -%}
```

So the first item renders as expected...in the second case it fails because it is not a supported language because of the `assert:contains`...however the error I’d like to report back is something like the following:

```
Expression `{{macro.translateWord("world", "es")}}` found at line 4 
failed because of expression `{{ assert:contains(languageIso, supportedLanguages) }}` at line 2.
```

I can get the current line via the interpreter (that would give me line 2 of the macro function) but I’d like to include more detail namely the original expression that dispatched to the macro function if at all possible. This is particularly useful if `macro.translateWord` is initiated often for example.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -735,7 +735,7 @@ public class Context extends ScopeMap<String, Object> {
   }
 
   public Node getCurrentNode() {
-    return this.currentNode;
+    return currentNode;
   }
 
   public void setCurrentNode(final Node currentNode) {
--- a/src/main/java/com/hubspot/jinjava/interpret/Context.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/Context.java
@@ -111,6 +111,7 @@ public class Context extends ScopeMap<String, Object> {
   private boolean unwrapRawOverride = false;
   private DynamicVariableResolver dynamicVariableResolver = null;
   private final Set<String> metaContextVariables; // These variable names aren't tracked in eager execution
+  private Node currentProcessedNode;
 
   public Context() {
     this(null, null, null, true);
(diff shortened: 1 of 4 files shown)


src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -111,6 +111,7 @@ public Context setValidationMode(boolean validationMode) {
   private boolean unwrapRawOverride = false;
   private DynamicVariableResolver dynamicVariableResolver = null;
   private final Set<String> metaContextVariables; // These variable names aren't tracked in eager execution
+  private Node currentProcessedNode;
```
jasmith-hs, line 114: I think `currentNode` is a bit simpler

src/main/java/com/hubspot/jinjava/interpret/Context.java
```diff
@@ -732,4 +733,12 @@ public void close() {
       resetValueConsumer.accept(previousValue);
     }
   }
+
+  public Node getCurrentProcessedNode() {
+    return this.currentProcessedNode;
```
jasmith-hs, line 738: The `this.` here is redundant.

jasmith-hs (commented):
Could you add a description to this PR that describes basically what you're wanting this for. (You are using jinjava and want to better track errors that result from macro function usage)

I don't think there are really any downsides to adding this. It is a very lightweight operation that provides more visibility about the interpreting process.

Do you have any opinions @boulter

Issue #743, opened 2021-09-07
[Feature Request]: Add support for base 64 encode/decode filters


derekcavaliero:
We've run into certain situations where having the ability to encode/decode a base 64 string inside HubL context would have been helpful.

Some community members suggested trying `{{ value|b64decode }}` and `{{ value|b64encode }}` as seen on this Stack Overflow post: https://stackoverflow.com/questions/39047657/base64-decode-string-in-jinja/39047772 - but it doesn't appear to be supported.

PR #763, opened 2021-10-05
Add Base 64 encode and decode filters

jasmith-hs:
Adds 2 new filters:

- `b64encode`
- `b64decode`

Each of them default to using UTF-8, but you can specify the charset as one of:
```
US_ASCII
ISO_8859_1
UTF_8
UTF_16BE
UTF_16LE
UTF_16
```
(If running on a machine that supports more charsets, you can use one of those too).
These filters are both supported in ansible

For https://github.com/HubSpot/jinjava/issues/743

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/Base64DecodeFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/Base64DecodeFilter.java
@@ -1,0 +1,43 @@
+package com.hubspot.jinjava.lib.filter;
+
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import java.nio.charset.Charset;
+import java.util.Base64;
+
+@JinjavaDoc(
+  value = "Decode a base 64 input into a string.",
+  input = @JinjavaParam(
+    value = "input",
+    type = "string",
+    desc = "The base 64 input to decode.",
+    required = true
+  ),
+  params = {
+    @JinjavaParam(
+      value = "encoding",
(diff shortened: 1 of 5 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/Base64DecodeFilter.java
```diff
@@ -0,0 +1,52 @@
...
+  params = {
+    @JinjavaParam(
+      value = "encoding",
+      type = "string",
+      desc = "The string encoding charset to use.",
+      defaultValue = "UTF-8"
+    )
+  },
+  snippets = {
+    @JinjavaSnippet(code = "{% set my_number = -53 %}\n" + "{{ my_number|abs }}")
```
jmaroeder, line 30: I think you copied this over from `AbsFilter` - should probably update it to use `b64decode`

src/main/java/com/hubspot/jinjava/lib/filter/Base64EncodeFilter.java
```diff
@@ -0,0 +1,91 @@
...
+  params = {
+    @JinjavaParam(
+      value = "encoding",
+      type = "string",
+      desc = "The string encoding charset to use.",
+      defaultValue = "UTF-8"
+    )
+  },
+  snippets = {
+    @JinjavaSnippet(code = "{% set my_number = -53 %}\n" + "{{ my_number|abs }}")
```
jmaroeder, line 36: Need to update this to use `b64encode`

jmaroeder (changes requested):
Cool! Just a couple missing snippets that I see

src/main/java/com/hubspot/jinjava/lib/filter/Base64DecodeFilter.java
jasmith-hs: Indeed, thanks for catching this!

Joeoh (approved):
Thanks for adding!

Issue #772, opened 2021-11-08
parse shell code error: ${#array[@]}


fangjinuo:
${#array[@]}  => $

because the {# is  the comment start flag in jinja2

jasmith-hs:
Many templates don't use whitespace after these tokens so this would, unfortunately, be a breaking change.

fangjinuo:
Not all template use whitespace, some template will use it , why not add a global configuration (e.g.   withWhitespaceAfterStartToken = false) ?

jasmith-hs:
If that is something that would work for your case, I can add it in as a [LegacyOverride](https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/LegacyOverrides.java)

fangjinuo:
add a global configuration in LegacyOverride

PR #773, opened 2021-11-08
* judge whitespace after {%, {#, {{

fangjinuo:
* judge whitespace after {%, {#, {{

fix the bug: #772


src/main/java/com/hubspot/jinjava/LegacyOverrides.java
```diff
@@ -9,11 +9,13 @@
   private final boolean evaluateMapKeys;
   private final boolean iterateOverMapKeys;
   private final boolean usePyishObjectMapper;
+  private final boolean useWhitespaceAfterStartToken;
```
jasmith-hs, line 12: I think a better name might be `whitespaceRequiredWithinTokens` as it makes it sound more like it changes how input whitespace is used rather than how whitespace is output

fangjinuo:
rename it,  useWhitespaceAfterStartToken -> whitespaceRequiredWithinTokens

PR #775, opened 2021-11-10
[Eager Execution] Handle modified variables in macro context scope

jasmith-hs:
We've seen some problems handling recursion for deferred macro functions that reference and modify previously unset variables such as:
```
{% macro renderNavigation(menuTree) %}
  {% set level = level + 1 %}
  <ul class="submenu level-{{ level }}">
    {% for node in menuTree.children %}
      {{ link(node) }}
    {% endfor %}
  </ul>
{% endmacro %} 
```
We need to make sure when deferring this macro that we don't end up reconstructing `{% set level = 1 %}` because that would not work correctly when the function used recursion. So we need to make sure that `level` gets deferred within the macro function and that it reconstructs with `{% set level = level + 1 %}`, and we can do that by evaluating it twice because level will be deferred at the start of the second evaluation. This makes it so that the `level + 1` gets deferred as-is.

Also, we need to make it so that future usage of the `level` variable are not deferred. Because the `{% set %}` tag only puts the value onto the lowest scope of the context, it does not get set outside of the macro function so we don't want to defer it any farther than the macro function's scope. We do this by stopping deferring "set" variables once we're outside a macro function's scope.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
@@ -142,6 +142,9 @@ public class EagerMacroFunction extends AbstractCallableMethod {
       return "";
     } else {
       try {
+        interpreter.getContext().put(Context.CURRENT_MACRO_FUNCTION_KEY, macroFunction);
+
+        int numEagerTokensStart = interpreter.getContext().getEagerTokens().size();
         String evaluation = (String) evaluate(
           macroFunction
             .getArguments()
@@ -149,6 +152,16 @@ public class EagerMacroFunction extends AbstractCallableMethod {
             .map(arg -> DeferredMacroValueImpl.instance())
             .toArray()
         );
+        if (interpreter.getContext().getEagerTokens().size() > numEagerTokensStart) {
+          evaluation =
+            (String) evaluate(
+              macroFunction
+                .getArguments()
+                .stream()
+                .map(arg -> DeferredMacroValueImpl.instance())
+                .toArray()
+            );
+        }
         result = (getStartTag(interpreter) + evaluation + getEndTag(interpreter));
       } catch (DeferredValueException e) {
         // In case something not eager-supported encountered a deferred value
--- a/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
@@ -142,8 +142,6 @@ public class EagerMacroFunction extends AbstractCallableMethod {
       return "";
     } else {
       try {
-        interpreter.getContext().put(Context.CURRENT_MACRO_FUNCTION_KEY, macroFunction);
-
         int numEagerTokensStart = interpreter.getContext().getEagerTokens().size();
         String evaluation = (String) evaluate(
           macroFunction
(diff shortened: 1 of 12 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerForTag.java
```diff
@@ -86,9 +86,10 @@ public String eagerInterpret(
           interpreter,
           true
         );
+
+      // Run for loop again now that the necessary values have been deferred
+      eagerExecutionResult = runLoopOnce(tagNode, interpreter);
```
jasmith-hs, line 91: Moved this because we should only have to run the for loop again if there are some speculative bindings. It's just an optimization, and it's unrelated to the reset of the changes here. (The logic for re-running a deferred for loop is very similar to how we need to rerun the deferred macro so that any values that got deferred the first time start off being deferred. This is because a for loop is much like a recursive macro function)

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerToken.java
```diff
@@ -15,12 +15,14 @@
...
 
   private final String importResourcePath;
+  private final String currentMacroFunction;
 
   public EagerToken(Token token, Set<String> usedDeferredWords) {
     this.token = token;
     this.usedDeferredWords = usedDeferredWords;
     this.setDeferredWords = Collections.emptySet();
     importResourcePath = acquireImportResourcePath();
+    currentMacroFunction = acquireCurrentMacroFunction();
```
jasmith-hs, line 25: We track the current macro function so that we know when to stop deferring values. We only want to defer up to the current macro functions context, and not and of the higher-level parents so that it doesn't become deferred in the code that runs outside the macro function's scope.

src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
```diff
@@ -89,13 +89,20 @@ public static void deferVariables(String[] varTokens, Map<String, Object> contex
...
+          .getCurrentMacroFunction()
+          .equals(context.getMacroStack().peek().orElse(null))
+      ) {
+        deferredProps.addAll(
+          getPropertiesUsedInDeferredNodes(
+            context,
+            rebuildTemplateForEagerTagTokens(eagerToken, true),
+            false
+          )
+        );
```
jasmith-hs, line 104: We only run this check for the variables that are deferred from being `set` because those are the ones that are attempted to be `put` onto the context. Therefore, those are the ones whose scope we need to limit. If trying to modify a variable that exists outside the macro function's context through something like `append()` within the macro function, we would still want to defer that as those changes propagate outside of the macro function's scope

src/main/java/com/hubspot/jinjava/lib/fn/eager/EagerMacroFunction.java
```diff
@@ -142,13 +142,24 @@ public String reconstructImage() {
...
+        int numEagerTokensStart = interpreter.getContext().getEagerTokens().size();
         String evaluation = (String) evaluate(
           macroFunction
             .getArguments()
             .stream()
             .map(arg -> DeferredMacroValueImpl.instance())
             .toArray()
         );
+        if (interpreter.getContext().getEagerTokens().size() > numEagerTokensStart) {
+          evaluation =
```
jasmith-hs, line 154: Run the evaluation again now that all the variables that we attempted to modify have been deferred. This makes it so that recursion will work properly and those modified variables will be properly deferred within the reconstructed image.

jasmith-hs (commented):
Some notes on the changes here

PR #776, opened 2021-11-12
[Eager Execution] Handle modification in double imported file

jasmith-hs:
When importing a file multiple times that has to defer the value of some of its exported variables, you could run into a certain case that occurs because we weren't reconstructing a set tag for deferring child bindings.
The simple test case that I created for this has an import file `deferred-modification.jinja`:
```
{% if deferred %}
{% set foo = foo ~'a' %}
{% endif %}
{% set foo = foo ~ 'b' %}
```
When importing as `bar`, this will end up "resolving" like:
```
{% set bar = {} %}{% if deferred %}
{% set foo = 'a' %}{% do bar.update({"foo": foo}) %}
{% endif %}
{% set foo = foo ~ 'b' %}{% do bar.update({"foo": foo}) %}
```
The problem ends up in that last line because we have `foo ~ 'b'`. This is fine if you are only importing this file once because  `foo` will still be `null` at this point, which is what we expect. However, if you import this twice like:
```
{% import 'deferred-modification.jinja' as bar1 %}
{% import 'deferred-modification.jinja' as bar2 %}
```
You will have some code that has a couple notable things:
```
...
{% set foo = foo ~ 'b' %}{% do bar1.update({"foo": foo}) %}
...
{% set foo = foo ~ 'b' %}{% do bar2.update({"foo": foo}) %}
...
```
And you can see that the second time we call `foo ~ 'b'`, foo will already have a value so you would end up with `bar1.foo == 'b'` and `bar2.foo == 'bb'`.
We can solve this by reconstructing set tags for these values (like `foo`) that get deferred while executing the import tag.
So we prepend the output with `{% set foo = null %}` so we end up with something more like:
```
{% set foo = null %}
...
{% set foo = foo ~ 'b' %}{% do bar1.update({"foo": foo}) %}
...
{% set foo = null %}
...
{% set foo = foo ~ 'b' %}{% do bar2.update({"foo": foo}) %}
...
```
This resets those values when beginning to read the reconstructed code from the import tags each time so that we get the expected result of `bar1.foo == 'b'` and `bar2.foo == 'b'`

### DIFF
--- a/src/test/resources/eager/handles-double-import-modification.expected.expected.jinja
+++ b/src/test/resources/eager/handles-double-import-modification.expected.expected.jinja
@@ -4,5 +4,5 @@
 
 
 ---
-{'foo': 'b', 'import_resource_path': 'deferred-modification.jinja'}
-{'foo': 'b', 'import_resource_path': 'deferred-modification.jinja'}
+b
+b
--- a/src/test/resources/eager/handles-double-import-modification.expected.expected.jinja
+++ b/src/test/resources/eager/handles-double-import-modification.expected.expected.jinja
@@ -1,0 +1,8 @@
+---
+
+
+
+
+---
+{'foo': 'b', 'import_resource_path': 'deferred-modification.jinja'}
+{'foo': 'b', 'import_resource_path': 'deferred-modification.jinja'}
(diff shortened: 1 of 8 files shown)


src/test/resources/eager/handles-double-import-modification.expected.expected.jinja
```diff
@@ -0,0 +1,8 @@
+---
+
+
+
+
```
jasmith-hs, line 5: Ah, these spaces shouldn't actually be here because we shouldn't be rendering anything from the import tags once we reach the second pass

jasmith-hs: I'll address that in a separate PR

PR #778, opened 2021-11-17
Fix NPEs in round filter

mattcoley:
Just check for null.


boulter:
ha. Mine has a test! :) https://github.com/HubSpot/jinjava/pull/779

mattcoley:
Ahh let me revert

PR #779, opened 2021-11-17
Handle null rounding method

boulter:
passing a null value as the rounding method would result in an NPE.


src/main/java/com/hubspot/jinjava/lib/filter/RoundFilter.java
```diff
@@ -77,7 +77,7 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
     }
 
     String method = "common";
-    if (args.length > 1) {
+    if (args.length > 1 && args[1] != null) {
```
mattcoley, line 80: Should we check arg[0] as well?

boulter: done

PR #781, opened 2021-11-17
Fix NPE in select filter test

mattcoley:
I think I beat @boulter to this one.


boulter:
no fair, I was in a meeting!

PR #782, opened 2021-11-23
Add NPE check in the parser

mattcoley:
Seeing NPEs on this line.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -256,7 +256,7 @@ public class ExtendedParser extends Parser {
     }
 
     Scanner.Token nextToken = getToken();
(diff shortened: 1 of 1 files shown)


boulter (approved):

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -255,7 +255,8 @@ protected AstDict dict() throws ScanException, ParseException {
       }
     }
 
-    if (!getToken().getImage().equals("}")) {
+    Scanner.Token nextToken = getToken();
+    if (nextToken == null || nextToken.getImage() == null || !nextToken.equals("}")) {
```
jasmith-hs, line 259:
```suggestion
    if (nextToken == null || !"}".equals(nextToken.getImage())) {
```
Typo, you're checking `nextToken.equals` rather than `nextToken.getImage().equals`

PR #788, opened 2021-11-30
Fix NPE with null timezone passed to today function

boulter:
errors looked like this

```
java.lang.NullPointerException: zoneId
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at java.base/java.time.ZoneId.of(ZoneId.java:400)
	at java.base/java.time.ZoneId.of(ZoneId.java:356)
	at com.hubspot.jinjava.lib.fn.Functions.today(Functions.java:133)
```

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
@@ -26,7 +26,9 @@ public class UnixTimestampFunctionTest {
     assertThat(Functions.unixtimestamp(d)).isEqualTo(epochMilliseconds);
     assertThat(
         Math.abs(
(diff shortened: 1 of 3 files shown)


src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
```diff
@@ -26,7 +26,9 @@ public void itGetsUnixTimestamps() {
     assertThat(Functions.unixtimestamp(d)).isEqualTo(epochMilliseconds);
     assertThat(
         Math.abs(
-          Functions.unixtimestamp(null) - ZonedDateTime.now().toEpochSecond() * 1000
+          Functions.unixtimestamp((Object) null) -
```
boulter, line 29: needed a cast

jasmith-hs (approved):

PR #790, opened 2021-12-01
Fixing reject filter evaluation when it has argument(s)

akatona84:


akatona84:
@jboulter / @boulter, @jasmith-hs , @mattcoley could u guys review pls?

jasmith-hs (approved):
LGTM

PR #794, opened 2021-12-02
[Eager Execution] Include the size of a map or list when determining the hashcode

jasmith-hs:
Hash map gets its hashcode by adding together the hashcode of its entries. An entry gets its hashcode by xor-ing the hashcode of the key and value. If the key and value are the same, then the hashcode of the entry will be 0. Therefore maps with only entries with equal key-value will have a hashcode of 0. We should be able to distinguish these at least somewhat by just including the size of the map as well as we are just trying to use the hashcode to tell if the map has changed.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
@@ -206,7 +206,9 @@ public class EagerReconstructionUtils {
       ) {
         return o; // Can't run hashcode as it will have infinite recursion and cause a stack overflow
       }
-      return o.hashCode();
+      return (
+        o.hashCode() + (o instanceof PyList ? ((PyList) o).size() : ((PyMap) o).size())
+      );
     }
     return o;
   }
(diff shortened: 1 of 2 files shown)


jmaroeder:
It seems like this new implementation would be prone to similar issues as the old one - e.g., {'1': '1'} and {'2': '2'} would still have the same hashcode

jasmith-hs:
Yeah, it's true that it can still happen with a case such as that. We could instead add the `keySet()` hashcode for hash maps. A case where a map is modified from `{x:x}` to `{y:y}` seems so rare as it requires a put for the new key and then a removal of the old key.

The situation where this occurred was already pretty strange. They used a map to determine if a for loop was executed once or more.
```
{% set num = {} %}
{% for i in collection %}
{% do num.update({'1':'1'}) %}
...
{% endfor %}
{% if num != {'1':'1'} %}
...
```

jmaroeder:
> A case where a map is modified from {x:x} to {y:y} seems so rare as it requires a put for the new key and then a removal of the old key.

I agree that it seems rare, but I expect hash maps consisting solely of entries where keys == values (and thus the hashcode is `0`) are _also_ rare... I would lean toward adding the keySet() hashcode, as that would cover both cases with (hopefully) minimal overhead

jmaroeder (approved):
🚢

PR #795, opened 2021-12-02
Do not rebuild collections when shuffling using CONSTANT_ZERO RandomNumberStrategy

boulter:
Even when using the CONSTANT_ZERO `RandomNumberStrategy`, values could change for collections since the collection is rebuilt.

This fixes that and also adds test coverage for the random filter.


src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyMap.java
```diff
@@ -38,6 +38,9 @@ public Object put(String s, Object o) {
 
   @Override
   public void putAll(Map<? extends String, ?> m) {
+    if (m == null) {
```
boulter, line 41: also fixed an NPE in here

PR #796, opened 2021-12-03
[Eager Execution] Don't update map with self

jasmith-hs:
This PR avoids the situation that could result in a StackOverflowError being thrown when evaluating reconstructed code if a file is imported with an alias that matches one of the variables declared in that file.
We avoid reconstructing something like `{% do env.update({'env': env}) %}`


jmaroeder (approved):
Looks good to me!

PR #805, opened 2021-12-15
[Eager Execution] Improve AST preserving

jasmith-hs:
Makes use of recursion in order to better preserve down to the identifier. For example,
if you have `foo.bar.append(deferred)`, then `foo.bar` should be preserved so that that
list object will be appended rather than `[].append(deferred)`, which has no effect.
See the tests added in EagerAstMethodTest.java for more examples.

Originally, I was just going to have EagerAstDot, EagerAstBracket, and EagerAstMethod (and EagerAstIdentifier) implement a new `PartiallyResolvable` interface, but I realised most of the Eager ast nodes could have a modifiable object (such as a list) returned and we would need to preserve the identifier for that list so that it can be properly referenced for the update during a second pass. 

By adding a method `getPartiallyResolved()` that uses `boolean:preserveIdentifier`, I refactored all of the AST nodes to use this method to reconstruct the `DeferredParsingException` that they will throw when the normal evaluation encounters some deferred value.

Within the `EvalResultHolder.reconstructNode()` method, we will return the evaluation result if it is primitive (String, Boolean, Integer, etc.) because they cannot be modified. We'll also return it if `preserveIdentifier == false` because in that case we can simply resolve it as much as possible (don't need to preserve the identifier if we aren't running a method on it). That is why EagerAstMethod uses `preserveIdentifier = true` when it calls the `getPartiallyResolved()` method. It is the only recursive usage of this method that turns that boolean flag on to let us know that we may be modifying something within the `property.prefix` of the method.

This may look like a ton of new code is added, but I think it is really a simplification. The refactoring ends up adding a few new methods to the `EvalResultHolder` interface so there are a few lines of boilerplate in each class that add up. But the logic in each of the Eager ast nodes is much more simple and generic.


src/main/java/com/hubspot/jinjava/el/ext/eager/EagerExtendedParser.java
```diff
@@ -58,6 +64,21 @@ public EagerExtendedParser(Builder context, String input) {
...
   }
 
+  @Override
+  protected AstEval eval(boolean required, boolean deferred)
+    throws ScanException, ParseException {
+    AstEval v = null;
+    Symbol startEval = deferred ? START_EVAL_DEFERRED : START_EVAL_DYNAMIC;
+    if (getToken().getSymbol() == startEval) {
+      consumeToken();
+      v = new AstEval(new EagerAstRoot(expr(true)), deferred);
```
jasmith-hs, line 74: This is added so that we have a way to just call `clearEvalResult()` after all the AST nodes have been processed, rather than as a `finally` clause at the end of each node.

(By the way, we need to clear the evaluation result because a Cache is used for the AstNodes, if they have the same string image as one that was used in a previous ELExpression, then it will be reused so we need to make sure that all the eval results for each Eager ast node is cleared after the current EL expression is evaluated.

Issue #809, opened 2022-01-03
[Enchancement] Migrate to Jakarta EL 8


dmaidaniuk:
Jinjava uses a very old EL 2.2 implementation that belongs to Java EE 6. Actually, JUEL implementation from Odysseus Software is dead now. [The last release](https://github.com/beckchr/juel/tags) was in 2014... I'm a bit confused with the fact that Jinjava sticks to this ancient EL implementation instead of using the current RI for EL 4 (or most fresh [Jakarta EL 8](https://projects.eclipse.org/projects/ee4j.el)).
We have the issue in our Eclipse Krazo project due to this very old EL implementation: https://github.com/eclipse-ee4j/krazo/issues/285

dmaidaniuk:
BTW, here is an official site of JUEL: http://juel.sourceforge.net/

PR #810, opened 2022-01-03
Feature/jakarta el 8

dmaidaniuk:
PR for issue #809


src/main/java/com/hubspot/jinjava/el/tree/impl/Scanner.java
```diff
@@ -0,0 +1,453 @@
...
+                }
+                return fixed(Symbol.LT);
+            case '>':
+                if (c2 == '=') {
+                    return fixed(Symbol.GE);
+                }
+                return fixed(Symbol.GT);
+            case '"':
+            case '\'': return nextString();
+            default: throw new IllegalArgumentException("Value " + c1 + " is not supported.");
```
jasmith-hs, line 386: This here causes a very large amount of the unit tests to fail. As this PR is currently, over half of the unit tests are failing.

jasmith-hs (changes requested):
I don't think we'd have a problem migrating off of the Odysseus implementation if it all works properly.
By the way, which repository did you get these new JUEL classes from, and is that repository still being maintained?

src/main/java/com/hubspot/jinjava/el/tree/impl/Scanner.java
jasmith-hs: Sorry, just realised you had marked this as a draft PR 3 days ago, I had originally looked over it when it was first opened.

dmaidaniuk: > This here causes a very large amount of the unit tests to fail.

Fixed))

dmaidaniuk:
> I don't think we'd have a problem migrating off of the Odysseus implementation if it all works properly. By the way, which repository did you get these new JUEL classes from, and is that repository still being maintained?

Well, I just built-in and upgrade Odysseus implementation according to EL 4 specification.

PR #811, opened 2022-01-04
[Eager Execution] Handle variables that get modified within macro function

jasmith-hs:
There were two problems occurring. First was that in macro function execution that threw a `DeferredParsingException`, we were not preserving the identifiers for the arguments passed in. So if the macro function modified one of the passed in arguments, we weren't modifying the actual reference object.
```
{% macro do_append(list, val) %}
{% do list.append(val) %}
{% endmacro %}

{% set list = [] %}
{% do do_append(list, deferred) %}
```
should keep `{% do do_append(list, deferred) %}`, but originally it was reconstructed as `{% do do_append([], deferred) %}` which has different functionality.

Secondly, we weren't reconstructing variables modified within macro functions if their name was the same as one of the macro function arguments. (Specifically, if it was named the same thing in the macro function context as it was outside the context). This was addressed by filtering out used deferred values from being deferred while evaluating the macro function if they matched an argument to the macro function.

```
1 {% macro do_append(list, val) %}
2 {% do list.append(val) %}
3 {% endmacro %}
4
5 {% set list = [] %}
6 {% do do_append(list, deferred) %}
```
In this example, we want `list` to get deferred on the main context on line 6 rather than line 2 so that the logic for checking the deferred words will see that `list` is not yet deferred, and needs to be... so it will prepend a `set` tag like `{% set list = [] %}` so that `list` is not just `null` during a second rendering pass.


src/main/java/com/hubspot/jinjava/util/DeferredValueUtils.java
```diff
@@ -93,7 +95,7 @@ public static void deferVariables(String[] varTokens, Map<String, Object> contex
         eagerToken.getCurrentMacroFunction() == null ||
         eagerToken
           .getCurrentMacroFunction()
-          .equals(context.getMacroStack().peek().orElse(null))
+          .equals(context.getParent().getMacroStack().peek().orElse(null))
```
jasmith-hs, line 98: What I'm trying to do here is find the context which the macro function is being called from and not defer certain values on that context level and higher (because within the macro function it is it's own context scope with variable names from its arguments which are isolated to it)

Calls the parent's macro stack because for the eager execution `executeInChildContext` logic, we enter a non-stacking scope which ends up having the parent context think that it has the macro function on the stack, even though it is its child context that the macro function started from.

PR #813, opened 2022-01-18
Make invalid timezone raise InvalidArgumentException

jmaroeder:
Changes the exception raised when an invalid timezone is passed to `datetimeformat` and `today` to `InvalidArgumentException`, which provides less misleading information than the `InvalidDateFormatException` being raised in the existing code.


jasmith-hs (approved):
Good call

PR #820, opened 2022-02-04
Handle nested import file paths

jasmith-hs:
There is a bug in Jinjava that affects the `import_resource_path` when there is a file that gets imported that has a macro, but that file also imports another file. The `import_resource_path` of the second file is put onto the context for the first file.

I created a test but it works like:
`double-import-macro.jinja`
```
{% import 'import-macro.jinja' -%}
{% macro print_path_macro2(var) -%}
{{ var|print_path }}
{{ print_path_macro(var) }}
{%- endmacro %}
```
`import-macro.jinja`
```
{% macro print_path_macro(var) %}
{{ var|print_path }}
{{ var }}
{% endmacro %}
```
Calling:
```
{% import 'double-import-macro.jinja' %}
{{ print_path_macro2('foo') }}
```
Should print
```
double-import-macro.jinja
import-macro.jinja
foo
```
But the bug had it outputting
```
import-macro.jinja
import-macro.jinja
foo
```
This PR fixes it for default and eager execution and also fixes the hashcode for macro functions to work better in eager execution. (Includes the `import_resource_path` of the macro in the hashcode.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
@@ -25,7 +25,9 @@ import com.hubspot.jinjava.util.HelperStringTokenizer;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 
 /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/ImportTag.java
```diff
@@ -164,6 +168,17 @@ public static void integrateChild(
...
 
+  public static Map<String, Object> getChildBindingsWithoutImportResourcePath(
+    Map<String, Object> childBindings
+  ) {
+    // Don't remove them from childBindings, because it is needed in a macro function's localContextScope
+    return childBindings
+      .entrySet()
+      .stream()
+      .filter(entry -> !entry.getKey().equals(Context.IMPORT_RESOURCE_PATH_KEY))
+      .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
```
jasmith-hs, line 179: This throws a NPE if there are any entries with `null` values

PR #828, opened 2022-02-23
Improve performance of selectattr and rejectattr

mattcoley:
Reusing the `Variable` object improves performance 50% for large collections.


boulter:
Wow, how did you find this and what makes constructing a new `Variable` so slow?

mattcoley:
It is not necessarily slow but unnecessary. Property resolution is also slow but I haven't been able to optimize that yet. Creating the variable constructs some additional objects like Lists so if you are selectattr'ing over a list of 10k objects you may be constructing an additional 40k objects that are not necessarily to create.

jessbrandi:
Awesome 👍

PR #829, opened 2022-02-24
[Eager Execution] Fix race condition in deferred AstNode evaluation

jasmith-hs:
There is a race condition that can happen in an AstNode that is evaluated with Eager Execution. If multiple threads share the same `ExpressionFactoryImpl`, then if a ValueExpression is pulled from the `TreeStore` cache before its evaluation (and subsequent `clearEvalResult()` call and it is used in another thread then it is possible that the stored `evalResult` from the first thread could be used in the second thread when encountering a DeferredValueException.

This is proven with the unit test `itIsThreadSafe()` that fails before the change I made by setting the cache size to 0 for eager execution's ExpressionFactory.

Here, is where it was possible to use the value stored from the other thread https://github.com/HubSpot/jinjava/blob/e144dcb387700ad98e95ad6f0b404f79d8204353/src/main/java/com/hubspot/jinjava/el/ext/eager/EvalResultHolder.java#L68-L71.
It could also happen here, but that's not easily reproducible in a unit test https://github.com/HubSpot/jinjava/blob/e144dcb387700ad98e95ad6f0b404f79d8204353/src/main/java/com/hubspot/jinjava/el/ext/eager/EvalResultHolder.java#L30-L31

### DIFF
--- a/src/main/java/com/hubspot/jinjava/Jinjava.java
+++ b/src/main/java/com/hubspot/jinjava/Jinjava.java
@@ -97,7 +97,7 @@ public class Jinjava {
       TreeBuilder.class.getName(),
       EagerExtendedSyntaxBuilder.class.getName()
     );
-    //    eagerExpConfig.setProperty(ExpressionFactoryImpl.PROP_CACHE_SIZE, "0");
+    eagerExpConfig.setProperty(ExpressionFactoryImpl.PROP_CACHE_SIZE, "0");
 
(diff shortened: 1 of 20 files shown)


Joeoh (approved):
Great find 👏

boulter:
wow, nice find.

hs-lsong:
:cool: That could be the cause of the mysterious content appears in another content.

PR #832, opened 2022-02-25
IncludeTag now respects ignore missing

aabeling:
Hi,
I did not find any information about which version of jinja2 should be supported with jinjava.
If at least version 2.2 should be supported then the include tag should also support "ignore missing":
https://jinja.palletsprojects.com/en/2.11.x/templates/#include


aabeling:
By the way: did you think about using GitHub workflows to automate the tests?

aabeling:
See issue #833

PR #841, opened 2022-04-21
Treat bad truthy value as false.

hs-lsong:
We get NPE for bad truthy value. Let's treat them as false.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DefaultFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DefaultFilter.java
@@ -75,7 +75,8 @@ public class DefaultFilter extends AbstractFilter implements AdvancedFilter {
       );
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/DefaultFilter.java
```diff
@@ -75,7 +75,8 @@ public Object filter(
       );
     }
 
-    boolean truthy = (boolean) parsedArgs.get(TRUTHY_PARAM);
+    Object truthyObj = parsedArgs.get(TRUTHY_PARAM);
+    boolean truthy = truthyObj != null && (boolean) truthyObj;
```
jasmith-hs, line 79: This may be simpler. What do you think?
```suggestion
    boolean truthy = Boolean.TRUE.equals(parsedArgs.get(TRUTHY_PARAM));
```

jasmith-hs (approved):

PR #850, opened 2022-05-06
[Eager Execution] Multiple loop variable reconstruction

jasmith-hs:
Fixes a problem where when there were multiple deferred for loops and deferred if tags, it would not reconstruct the output correctly. It would either reconstruct in the wrong place, or reconstruct with the wrong value.
See my comment for what the output of this new test used to be before this change.

Also, I fixed how a namespace gets reconstructed by having it extend `PyishSerializable` so that I could override it's `toPyishString()` implementation so that it will be like `namespace(%s)`

### DIFF
--- a/src/test/resources/eager/reconstructs-with-multiple-loops.expected.jinja
+++ b/src/test/resources/eager/reconstructs-with-multiple-loops.expected.jinja
@@ -1,0 +1,14 @@
+{% set my_list = [] %}{% for i in deferred %}
+  {% if deferred %}
+    {% for j in deferred %}
+      {% if deferred %}
+        {% do my_list.append(1) %}
+      {% else %}
+        {% do my_list.append(2) %}
+      {% endif %}
+    {% endfor %}
+  {% else %}
+    {% do my_list.append(3) %}
+  {% endif %}
+{% endfor %}
+{{ my_list }}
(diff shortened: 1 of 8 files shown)


src/test/resources/eager/reconstructs-with-multiple-loops.expected.jinja
```diff
@@ -0,0 +1,23 @@
...
+      {% do beta.append(1) %}
+      {% if deferred %}
+        {% do alpha.append(3) %}
+      {% else %}
+        {% do alpha.append(4) %}
+      {% endif %}
+    {% endfor %}
+  {% endif %}
+{% endfor %}
+{{ alpha ~ beta }}
```
jasmith-hs, line 23: Before this change, the output would look like:
```
[][]
[0][0]
{% set beta = [0, 1] %}{% for i in deferred %}
  {% if deferred %}
    {% set alpha = [0] %}{% for j in deferred %}
      {% if deferred %}
        {% do alpha.append(1) %}
      {% else %}
        {% do alpha.append(2) %}
      {% endif %}
    {% endfor %}
  {% else %}
    {% set alpha = [0] %}{% for j in deferred %}
      {% do beta.append(1) %}
      {% if deferred %}
        {% do alpha.append(3) %}
      {% else %}
        {% do alpha.append(4) %}
      {% endif %}
    {% endfor %}
  {% endif %}
{% endfor %}
{{ [0] ~ beta }}
```
Clearly this is wrong for 2 reasons: `alpha` does not get deferred (and its value is reset each for loop to `[0]`), and `beta` starts off with the wrong value

PR #855, opened 2022-05-12
Strftime formatter

svarmahubspot:
Fix the out of bound array index exception in StrftimeFormatter in case of > 255 ascii char in data format


boulter (approved):
great, I like the tests.

PR #857, opened 2022-05-16
Fix number truthiness by casting to individual types

jasmith-hs:
Truthiness for non-integer numbers functions differently in Jinjava than it does in Jinja. The following should print `yes`, but since it currently only takes the `Number.intvalue()`, it would return false. This PR fixes the issue by checking which type of Number the object is.
```
{% if 0.5 %}
yes
{% else %}
no
{% endif %}
```

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
+++ b/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
@@ -39,19 +39,7 @@ public final class ObjectTruthValue {
     }
 
     if (object instanceof Number) {
-      if (object instanceof Integer) {
-        return (Integer) object != 0;
-      }
-      if (object instanceof Double) {
-        return (Double) object != 0;
-      }
-      if (object instanceof Long) {
-        return (Long) object != 0;
-      }
-      if (object instanceof Short) {
-        return (Short) object != 0;
-      }
-      return (Float) object != 0;
+      return ((Number) object).doubleValue() != 0;
     }
 
     if (object instanceof String) {
--- a/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
+++ b/src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
@@ -39,7 +39,19 @@ public final class ObjectTruthValue {
     }
 
     if (object instanceof Number) {
-      return ((Number) object).intValue() != 0;
+      if (object instanceof Integer) {
+        return (Integer) object != 0;
+      }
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
```diff
@@ -39,7 +39,19 @@ public static boolean evaluate(Object object) {
...
+      if (object instanceof Double) {
+        return (Double) object != 0;
+      }
+      if (object instanceof Long) {
+        return (Long) object != 0;
+      }
+      if (object instanceof Short) {
+        return (Short) object != 0;
+      }
+      return (Float) object != 0;
```
jasmith-hs, line 54: Alternatively, I could just do:
```
return ((Number)object).doubleValue() != 0;
```

jmaroeder: Unfortunately I don't think that will quite work - Python doesn't treat `0.0` as truthy

src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
```diff
@@ -39,7 +39,19 @@ public static boolean evaluate(Object object) {
     }
 
     if (object instanceof Number) {
```
boulter, line 41: There are more [subclasses of `Number`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Number.html). While you don't have to handle them all, perhaps a sane default would be to use `Number.floatValue()`.

src/test/java/com/hubspot/jinjava/util/ObjectTruthValueTest.java
```diff
@@ -15,6 +14,46 @@ public void itEvaluatesObjectsWithObjectTruthValue() {
       .isFalse();
   }
 
+  @Test
+  public void itEvaluatesIntegers() {
```
boulter, line 18: might be worth a utility function here that takes a and b and does the assertions so you're not duplicating the code for every type.

src/main/java/com/hubspot/jinjava/util/ObjectTruthValue.java
jasmith-hs: Yeah, I wrote it like that thinking it would be faster, but I tested it. This shorter way is faster for longs, about the same time for floats and ints so I'm going to do the simpler approach

boulter (approved):

PR #859, opened 2022-05-18
[Eager Execution] Remove discarded eager tokens when not partially evaluating macro functions

jasmith-hs:
This helps to better track the total number of eager tokens, which when reaching a certain threshold will start killing the eager execution processing. With partial macro function evaluation turned off, however many tokens were deferred during the macro function's evaluation are replaced by a single token, which is the deferral of the macro function's call itself. This will improve the tracking of eager tokens that are actually used, and will prevent there from being "dead" eager tokens tracked on the context.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
+++ b/src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
@@ -9,6 +9,7 @@ import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter.InterpreterScopeClosable;
 import com.hubspot.jinjava.tree.Node;
 import com.hubspot.jinjava.util.LengthLimitingStringBuilder;
+import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/fn/MacroFunction.java
```diff
@@ -81,6 +82,9 @@ public Object doEvaluate(
           !interpreter.getContext().getEagerTokens().isEmpty()
         )
       ) {
+        interpreter
+          .getContext()
+          .removeEagerTokens(new HashSet<>(interpreter.getContext().getEagerTokens()));
```
hs-lsong, line 87: Does this remove the eager tokens added elsewhere?

jasmith-hs: We will enter a new scope when we evaluate the macro function. When entering a new scope, the eager tokens don't get carried over to the new child so it is a fresh set. Therefore, we are only removing the tokens that were added within this scope. The tokens that are added elsewhere outside of this macro function's scope will not get removed

PR #864, opened 2022-05-23
[Eager Execution] Improve ForTag with deferred tokens

jasmith-hs:
This PR addresses an issue where because a for loop uses a separate scope, if there are modifications done inside of a for loop that has some deferred tokens, then the for loop must be deferred rather than exploded. For example, we can't explode this:
```
{% set foo = 1 %}
{% for i in range(2) %}
{% set foo = deferred %}
{{ foo }}
{% endfor %}
{{ foo }}
```
to:
```
{% set foo = 1 %}
{% set foo = deferred %}
{{ foo }}
{% set foo = deferred %}
{{ foo }}
{{ foo }}
```
because the `{{ foo }}` that comes after the for loop should have a value of `1`. This is because a set tag inside of a for loop will not modify variables outside of the for loop. So we'd need to defer the for loop so that the correct scope can remain.

As part of this, I'm also making a modification to what we store inside of the `EagerToken`. Rather than storing the name of the current macro function, we will store one of the Context's call stacks (specifically the macro stack). We are using the call stack to be able to track the scope in which the eager token was created because adding a new scope will create a new call stack. So we will be able to check if the scope of the eager token is the same as the context's current scope. If they are the same scope, then we will defer normally, but if they're on different scopes, we won't defer any of the "set" variables. As they would be set on a different scope level.

In order to track this information and revert when running a for loop, we need to add a new boolean option to the `executeInChildContext()` method that will allow us to watch for context changes, but not run in deferred execution mode. This way, if nothing ends up getting deferred, then we can simply allow any changes to the context to happen, as they would've for a regular for loop execution. Rather than just adding another boolean flag, I converted all the flags into an immutable object so it is more visible what the config options are for that method call. Previously, `checkForContextChanges` would always force deferred execution mode to be on, but now, we can check for context changes without needing deferred execution mode to be on, which is what we'll use to run the for loop initially.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
@@ -10,6 +10,7 @@ import com.hubspot.jinjava.tree.parse.TagToken;
 import com.hubspot.jinjava.util.EagerExpressionResolver.EagerExpressionResult;
 import com.hubspot.jinjava.util.EagerExpressionResolver.EagerExpressionResult.ResolutionState;
 import com.hubspot.jinjava.util.EagerReconstructionUtils;
+import com.hubspot.jinjava.util.EagerReconstructionUtils.EagerChildContextConfig;
 import com.hubspot.jinjava.util.LengthLimitingStringJoiner;
 import java.util.Collections;
 import java.util.Optional;
@@ -46,9 +47,7 @@ public class EagerBlockSetTagStrategy extends EagerSetTagStrategy {
         );
       },
       interpreter,
-      true,
-      false,
-      false
+      EagerChildContextConfig.newBuilder().withTakeNewValue(true).build()
     );
   }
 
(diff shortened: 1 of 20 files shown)


hs-lsong:
> This is because a set tag inside of a for loop will not modify variables outside of the for loop.

Hmm, just learned this.

jasmith-hs:
Yeah, Jinja ends up being pretty restricte in that sense: [docs](https://jinja.palletsprojects.com/en/3.0.x/templates/#:~:text=Please%20note%20that%20assignments%20in%20loops%20will%20be%20cleared%20at%20the%20end%20of%20the%20iteration%20and%20cannot%20outlive%20the%20loop%20scope.)

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
```diff
@@ -46,9 +47,7 @@ protected EagerExecutionResult getEagerExecutionResult(
         );
       },
       interpreter,
-      true,
-      false,
-      false
+      EagerChildContextConfig.newBuilder().withTakeNewValue(true).build()
```
boulter, line 50: this is much easier to understand then a list of booleans!

boulter (approved):

Issue #868, opened 2022-05-31
Nested macro functions with same name functions differently in Jinjava vs Jinja


jasmith-hs:
The output of:
```
{% macro foo() -%}
{% macro foo() -%}
1
{%- endmacro %}
{{ foo() }} 2
{%- endmacro %}
{{ foo() }}
{{ foo() }}
```
Should be:
```
1 2

1 2
```
However, in Jinjava, the output is:
```
1 2
1
```

This is because macro functions defined inside of a macro function will override existing macro functions, rather than defining macro functions that exist solely within that scope.

This problem extends beyond just overriding macro functions with the same name. Macro functions defined inside of another macro function should not be usable outside of their defined scope. For example:
```
{% macro foo() -%}
{% macro bar() -%}
1
{%- endmacro %}
{{ bar() }} 2
{%- endmacro %}
{{ foo() }}
{{ bar() }}
```
The final `bar()` should not be calling anything, but the result in Jinjava is:
```
1 2
1
```

hs-lsong:
I can't get this. why the second `{{ foo() }}` only outputs `1` instead of `1 2`? What would be output of a third `{{ foo() }}` ?

hs-lsong:
:+1:

PR #869, opened 2022-05-31
Fix macro function scoping inside of another macro function

jasmith-hs:
Addresses https://github.com/HubSpot/jinjava/issues/868
Rather than copying the global macros over when evaluating a macro function, we'll instead create a new HashMap copy of it so that if any new macro functions are added, they will be localised to the current scope, rather than the parent's scope.


jmaroeder (approved):
The test looks like it pretty well covers the issues mentioned in #868 - LGTM!

hs-lsong (approved):

PR #875, opened 2022-06-14
Write null map values within PyishObjectMapper

jasmith-hs:
When I handled null keys within maps, I unnecessarily added configuration to disable writing null values https://github.com/HubSpot/jinjava/pull/835

In the case of an operation like `{% do my_map.update({'foo': null}) %}`, the update map would get serialized as `{}`, which is not correct.


jmaroeder (approved):
Good catch!

PR #876, opened 2022-06-14
[Eager Execution] Handle for loop resetting when a DeferredValueException is thrown

jasmith-hs:
With https://github.com/HubSpot/jinjava/pull/872, I fixed how values are reverted in for loops, but I didn't handle the case when a DeferredValueException is thrown by calling `getTag().interpretUnchecked(tagNode, interpreter)` inside EagerForTag. This can happen if the `loop` variable gets deferred as a DeferredValueException will get thrown here: https://github.com/HubSpot/jinjava/blob/3a35687a4ffab46cd3955b01b5d399676aefd28f/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java#L265-L277

The test case does a good job of demonstrating this bug.
```
{% set my_list = [] %}
{% for i in range(2) %}
{% for j in deferred %}
{% if loop.first %}
{% do my_list.append(1) %}
{% endif %}
{% endfor %}
{% endfor %}
{{ my_list }}
```
Before this fix, the output would look like:
```
{% for i in [0, 1] %}
{% for j in deferred %}
{% if loop.first %}
{% do my_list.append(1) %}
{% endif %}
{% endfor %}
{% endfor %}
{{ my_list }}
```
We would lose the initial value of `my_list` as it would fail to get reconstructed. This is because it would become a DeferredValue. Normally, if a variable becomes a DeferredVariable, it will get reconstructed within the logic of `EagerReconstructionUtils.executeInChildContext()`, but since the aforementioned DeferredValueException was not getting caught, it was breaking out of the `executeInChildContext()` logic and we weren't ever flagging the `my_list` variable as one that had been deferred within the supplied function's evaluation.
By adding a new `ResolutionState.NONE` and catching the exception here: https://github.com/HubSpot/jinjava/blob/014a49d435eebc5b215feccea4403ce17b14d602/src/main/java/com/hubspot/jinjava/util/EagerExpressionResolver.java#L386-L388
We are able to revert the `my_list` value to what it was originally, and properly run the for loop in `deferredExecutionMode`

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
@@ -31,24 +31,25 @@ public class EagerBlockSetTagStrategy extends EagerSetTagStrategy {
     String expression,
     JinjavaInterpreter interpreter
   ) {
-    int numEagerTokens = interpreter.getContext().getEagerTokens().size();
-    return EagerReconstructionUtils.executeInChildContext(
-      eagerInterpreter -> {
-        StringBuilder sb = new StringBuilder();
-        for (Node child : tagNode.getChildren()) {
-          sb.append(child.render(eagerInterpreter).getValue());
-        }
-        return EagerExpressionResult.fromString(
-          sb.toString(),
-          numEagerTokens ==
-            eagerInterpreter.getContext().getParent().getEagerTokens().size()
-            ? ResolutionState.FULL
-            : ResolutionState.PARTIAL
-        );
-      },
+    EagerExecutionResult result = EagerReconstructionUtils.executeInChildContext(
+      eagerInterpreter ->
+        EagerExpressionResult.fromSupplier(
+          () -> {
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerBlockSetTagStrategy.java
```diff
@@ -31,24 +31,25 @@ protected EagerExecutionResult getEagerExecutionResult(
...
-        StringBuilder sb = new StringBuilder();
-        for (Node child : tagNode.getChildren()) {
-          sb.append(child.render(eagerInterpreter).getValue());
-        }
-        return EagerExpressionResult.fromString(
-          sb.toString(),
-          numEagerTokens ==
-            eagerInterpreter.getContext().getParent().getEagerTokens().size()
-            ? ResolutionState.FULL
-            : ResolutionState.PARTIAL
```
jasmith-hs, line 46: The change is this file is just refactoring to not have to duplicate this ResolutionState selection logic outside of EagerExpressionResolver.java

PR #882, opened 2022-06-22
Add Null Checks to Interpreter for Error counts and Split Filter

nathanielpautzke:
Adding a filter on jinjava interpreter for counting errors to only allow non null objects. Unable to reproduce how null values are getting into the LinkedList. Also, adding null check to Split filter to default a null value to using whitespace character as the default separator does already. Null as the separator caused a Null Pointer exception in the google splitter library. 

@jasmith-hs

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
@@ -717,21 +717,27 @@ public class JinjavaInterpreter implements PyishSerializable {
     }
 
     // Limit the number of error.
-    if (errors.size() < MAX_ERROR_SIZE) {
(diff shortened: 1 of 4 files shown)


boulter (changes requested):
We run jinjava only in a single thread so I don't think this is a threading problem. Adding synchronization has a performance penalty for code that shouldn't be multi-threaded anyway. What was the stack trace from the error? From there you can see which line the NPE was on and just add some null checking.

jasmith-hs:
We determined it's impossible for this to happen without there being a concurrency issue, but I don't know where one could occur

jasmith-hs:
A check for a null value there could work to prevent this sporadic, non-reproducible bug but it doesn't explain how a null value got there in the first place

PR #883, opened 2022-06-22
Add random_int function to JinJava

Houzl:
Current we use `{{ range(1,40000) | random }}`  to generate a random integer. But it’s `{{ range(1,40000) }} ` only returns the first 1000 numbers because of the  RAM or CPU concerns. 

add new `random_int()` function to generate a positive random int. It could also useful to CMS developers.


Houzl:
Instead of do it in JinJava, we want to do it in our own repo.
https://hubspot.slack.com/archives/CGAKVNT6G/p1655927186540989?thread_ts=1655909754.448779&cid=CGAKVNT6G

PR #894, opened 2022-07-20
Handle deferred variable which references same address as other variable

jasmith-hs:
Since variables on the context can be modified and there can multiple keys pointing to the same object, when a variable gets deferred from attempted modification, we must defer all other keys which reference that object. We don't know which function calls will result in a modification, so we defer all other keys for anything that isn't "setting", because we know that setting results in a new object being created.

To do this, I add a new class which extends LazyExpression called `LazyReference` and a deferred version of it called `LazyDeferredReference`. These allow us to track these referential defers and reconstruct variables so that they point to the same object when running a second rendering pass.
```
{% set foo = [] -%}
{%- set bar = foo -%}
{% do foo.append(1) %}
{% do foo.append(deferred) %}
{% do foo.append(2) %}
{{ foo ~ 'and' ~ bar }}
```
Even though `foo` is being modified and ends up getting deferred from the `{% do foo.append(deferred) %}` call, we need to also defer `bar` because that append call will modify it as well. We need to therefore look through the context and find any variables that reference the same address as `foo` does (in this case) and defer them as well when deferring non-set variables.


boulter (approved):

boulter:
This is why there are no mutable variables in jinja. the `append` method is leaked method from the java class that users found which certainly complicates things.

jasmith-hs:
If objects weren't mutable in Jinjava it would make my life much easier

jmaroeder (approved):

PR #895, opened 2022-07-20
Bind custom operators between `mul` and `unary`

### LINKED ISSUE #893: Unary minus binds too loosely
Consider the following expression:

```
{{ -1 is integer }}
```

I would expect this to evaluate to `true`, since `-1` is an `integer`. However, this actually throws an error.

```
Exception in thread "main" com.hubspot.jinjava.interpret.FatalTemplateErrorsException: Cannot negate 'class java.lang.Boolean'
	at com.hubspot.jinjava.Jinjava.render(Jinjava.java:192)
	at Scratch.main(scratch.java:6)
```

This implies that the unary minus is binding looser than the `is` operator. That is, Jinjava interprets that expression the same as `-(1 is integer)` when it should interpret it as `(-1) is integer`.

tkindy:
Closes #893 

Currently, our custom operators like `|` and `is` bind tighter than unary operators. This means `-1 is integer` incorrectly evaluates like `-(1 is integer)`, leading to an error when Jinjava tries to make `true` negative. This is because these operators are defined within `value` which is called by `unary`, and therefore have precedence greater than the unary operators.

This PR pulls our custom operators up from the `unary` level of the JUEL grammar to the `mul` level. This means they bind tighter than `mul` but less tightly than `unary`. From my testing, this aligns with Jinja's precedence rules.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -452,7 +452,46 @@ public class ExtendedParser extends Parser {
             break;
           }
         default:
-          return v;
+          if (
+            "|".equals(getToken().getImage()) && lookahead(0).getSymbol() == IDENTIFIER
+          ) {
+            do {
+              consumeToken(); // '|'
+              String filterName = consumeToken().getImage();
+              List<AstNode> filterParams = Lists.newArrayList(v, interpreter());
+
+              // optional filter args
+              if (getToken().getSymbol() == Symbol.LPAREN) {
+                AstParameters astParameters = params();
+                for (int i = 0; i < astParameters.getCardinality(); i++) {
+                  filterParams.add(astParameters.getChild(i));
+                }
+              }
+
+              AstProperty filterProperty = createAstDot(
+                identifier(FILTER_PREFIX + filterName),
+                "filter",
+                true
+              );
+              v = createAstMethod(filterProperty, createAstParameters(filterParams)); // function("filter:" + filterName, new AstParameters(filterParams));
+            } while ("|".equals(getToken().getImage()));
+          } else if (
+            "is".equals(getToken().getImage()) &&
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -427,53 +427,30 @@ protected AstRangeBracket createAstRangeBracket(
   }
 
   @Override
-  protected AstNode value() throws ScanException, ParseException {
```
tkindy, line 430: This diff is really weird, but basically I moved the parsing of our custom operators from the `default` of `value` to the `default` of `mul`.

jasmith-hs (changes requested):
assertThat(interpreter.render("{{ (-5 * -4 | abs) }}")).isEqualTo("-20");

I think the filter operation should be at the same level as `unary` rather than the same level as mul.
Might make sense to add a new `filter(boolean required)` method and have `mul()` point to that instead of `unary()` and have the filter method call unary and do the `|` and `is` logic

tkindy:
> `assertThat(interpreter.render("{{ (-5 * -4 | abs) }}")).isEqualTo("-20");`

Thanks for the extra test case! I confirmed Jinja indeed returns -20 here.

image

> I think the filter operation should be at the same level as `unary` rather than the same level as mul.

I think you mean filters should bind tighter than `mul` but less tightly than `unary`; that is, that we should introduce a new level of precedence. They currently bind at the same level as `unary`, and that's the problem.

Your implementation suggestion sounds good to me, I'll try that out.

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
tkindy: With my recent changes, the diff makes more sense now.

jasmith-hs (approved):
LGTM

PR #902, opened 2022-07-28
Reintroduce operator precedence fix with legacy override

### LINKED ISSUE #893: Unary minus binds too loosely

tkindy:
Closes #893 

This PR is basically https://github.com/HubSpot/jinjava/pull/895 but with the changes disabled by default. They can be enabled by setting a new flag on `LegacyOverrides` as part of the `JinjavaConfig`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
@@ -434,7 +434,7 @@ public class ExtendedParser extends Parser {
     if (v == null) {
       return null;
     }
-    return foo(v);
+    return parseOperators(v);
   }
 
   protected AstRightValue createAstNested(AstNode node) {
@@ -520,14 +520,14 @@ public class ExtendedParser extends Parser {
           if (shouldUseNaturalOperatorPrecedence()) {
             return v;
           }
-          return foo(v);
+          return parseOperators(v);
       }
     }
   }
 
-  private AstNode foo(AstNode v0) throws ScanException, ParseException {
+  private AstNode parseOperators(AstNode left) throws ScanException, ParseException {
     if ("|".equals(getToken().getImage()) && lookahead(0).getSymbol() == IDENTIFIER) {
-      AstNode v = v0;
+      AstNode v = left;
 
       do {
         consumeToken(); // '|'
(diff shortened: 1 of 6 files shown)


tkindy:
Actually, I want to move some of the tests around first.

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -578,4 +628,20 @@ public AstNode createAstNode(AstNode... children) {
...
+        .getConfig()
+        .getLegacyOverrides()
+        .isUseNaturalOperatorPrecedence()
+    );
+  }
+
+  @FunctionalInterface
+  private interface ParseLevel {
+    AstNode apply(boolean required) throws ScanException, ParseException;
+  }
```
jasmith-hs, line 646: I like this!

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
```diff
@@ -578,4 +628,20 @@ public AstNode createAstNode(AstNode... children) {
...
+
+  private static boolean shouldUseNaturalOperatorPrecedence() {
+    return (
+      JinjavaInterpreter.getCurrent() != null &&
+      JinjavaInterpreter
+        .getCurrent()
+        .getConfig()
+        .getLegacyOverrides()
+        .isUseNaturalOperatorPrecedence()
+    );
```
jasmith-hs, line 640: Theres
`JinjavaInterpreter.getCurrentMaybe().map(interpreter -> interpreter.getConfig().getLegacyOverrides().isUseNaturalOperatorPrecedence()).orElse(false)` too if you're big on writing expressions like that

jasmith-hs (approved):

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
tkindy: Ooh I very much am, thanks! 😄 I don't know why I didn't think to just to an `Optional.ofNullable` myself.

src/main/java/com/hubspot/jinjava/el/ext/ExtendedParser.java
tkindy: Thanks, I'm glad! 😄

PR #903, opened 2022-07-28
Reintroduce whitespace control parsing fix with legacy override

### LINKED ISSUE #892: Unary minus operator at beginning of expression is ignored
Adding parenthesis around an expression with a unary sub-expression, changes the output incorrectly.

`{{ (-10 | abs + 4)}}  is -6`
`{{ -10 | abs + 4 }}  is 14`

In the former example, the unary operation is being applied last, whereas in the latter example, the unary operation is being applied before the `abs` filter (as i would expect)

tkindy:
Closes #892 
Depends on #902 

This PR is basically #896 but with the changes disabled by default. They can be enabled by setting a new flag on `LegacyOverrides` as part of the `JinjavaConfig`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
+++ b/src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
@@ -4,7 +4,7 @@ public class StrictWhitespaceControlParser implements WhitespaceControlParser {
 
   @Override
   public boolean hasLeftTrim(String unwrapped) {
-    return unwrapped.startsWith("-");
+    return unwrapped.charAt(0) == '-';
   }
 
   @Override
@@ -14,7 +14,7 @@ public class StrictWhitespaceControlParser implements WhitespaceControlParser {
 
   @Override
   public boolean hasRightTrim(String unwrapped) {
-    return unwrapped.endsWith("-");
+    return unwrapped.charAt(unwrapped.length() - 1) == '-';
   }
 
   @Override
--- a/src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
+++ b/src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
@@ -1,0 +1,24 @@
+package com.hubspot.jinjava.tree.parse;
+
+public class StrictWhitespaceControlParser implements WhitespaceControlParser {
+
+  @Override
+  public boolean hasLeftTrim(String unwrapped) {
+    return unwrapped.startsWith("-");
+  }
+
+  @Override
+  public String stripLeft(String unwrapped) {
+    return unwrapped.substring(1);
+  }
+
+  @Override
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/tree/parse/Token.java
```diff
@@ -83,6 +84,37 @@ public void setRightTrimAfterEnd(boolean rightTrimAfterEnd) {
...
+   * @return the content stripped of any whitespace control characters.
+   */
+  protected final String handleTrim(String unwrapped) {
+    boolean parseWhitespaceControlStrictly =
+      JinjavaInterpreter.getCurrent() != null &&
+      JinjavaInterpreter
+        .getCurrent()
+        .getConfig()
+        .getLegacyOverrides()
+        .isParseWhitespaceControlStrictly();
```
jasmith-hs, line 100: Same as https://github.com/HubSpot/jinjava/pull/902#discussion_r932519485 if you like that idea

src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
```diff
@@ -0,0 +1,24 @@
+package com.hubspot.jinjava.tree.parse;
+
+public class StrictWhitespaceControlParser implements WhitespaceControlParser {
+
+  @Override
+  public boolean hasLeftTrim(String unwrapped) {
+    return unwrapped.startsWith("-");
```
jasmith-hs, line 7: Might as well have `unwrapped.charAt(0) == '-'`, which is faster

src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  @Override
+  public String stripLeft(String unwrapped) {
+    return unwrapped.substring(1);
+  }
+
+  @Override
+  public boolean hasRightTrim(String unwrapped) {
+    return unwrapped.endsWith("-");
```
jasmith-hs, line 17: And this could be `unwrapped.charAt(unwrapped.length() - 1) == '-'`

jasmith-hs (approved):

src/main/java/com/hubspot/jinjava/tree/parse/Token.java
tkindy: Definitely, thanks!

src/main/java/com/hubspot/jinjava/tree/parse/StrictWhitespaceControlParser.java
tkindy: Good call, given that this gets called a bunch throughout a whole template.

tkindy: I also added a unit test just to make sure empty expressions don't reach this point without any characters in them (so there's no need to check that the string isn't empty before indexing into it).

Issue #865, opened 2022-05-23
[Eager Execution] Add short-circuiting for EagerAstChoice


jasmith-hs:
Add short-circuiting for EagerAstChoice.
Similar to https://github.com/HubSpot/jinjava/pull/737 which was done to add short-circuiting for EagerAstBinary, it should also be done for EagerAstChoice.

For example:
```
deferred ? foo_list.add(foo) : foo_list.add(1 + 1)
```
should become:
```
deferred ? foo_list.add('foo val') : foo_list.add(2)
```
and not:
```
deferred ? true : true
```

PR #904, opened 2022-07-29
Short circuit EagerAstChoice on ChoiceIdentifier

nathanielpautzke:
When trying to resolve a choice tag, if the identifier is deferred the yes and no options will be evaluated when they should not be. This will short circuit the expression to not evaluate the yes or no expressions. 

Expression: 
```
deferred ? foo_list.add(foo) : foo_list.add(1 + 1)
```
Before fix: 
```
deferred ? true : true
```
Now fix will be evaluated to: 
```
deferred ? foo_list.add('foo val') : foo_list.add(2)
```
Foo will be resolved to 'foo val' variable value while 1 + 1 will resolve leaving the identifier as deferred. If identifier is not deferred and is true the expression will resolve if the yes node is not deferred. Similar to if the identifier is false and the no node is valid it will resolve as well. 

Relates to issue: [Issue 865](https://github.com/HubSpot/jinjava/issues/865)

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/eager/EagerAstChoice.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/eager/EagerAstChoice.java
@@ -1,5 +1,6 @@
 package com.hubspot.jinjava.el.ext.eager;
 
+import com.hubspot.jinjava.el.NoInvokeELContext;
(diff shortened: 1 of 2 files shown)


jasmith-hs (approved):
LGTM

PR #908, opened 2022-08-30
Return long value when under min int value from IntFilter

AlpriElse:
Returning long values when over `Integer.MAX_VALUE` was added in this pr:
https://github.com/HubSpot/jinjava/pull/315

This pr also makes sure this also happens when we underflow.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/IntFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/IntFilter.java
@@ -74,7 +74,7 @@ public class IntFilter implements Filter {
   }
 
   private Object convertResult(Long result) {
-    if (result > Integer.MAX_VALUE) {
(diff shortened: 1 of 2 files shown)


mattcoley (approved):

mattcoley:
I am confused how this will solve the ticket. Weren't we getting underflow for large values?

AlpriElse:
> I am confused how this will solve the ticket. Weren't we getting overflow for large values?

There is actually two problems that was in the JIRA.

1. The `IntFilter` underflowing eg. `-42595200000|int` -> 354472960
2. The `DateTimeFormatFilter` not properly converting `-42595200000L` also presumably due to underflow because we are interpreting the value as an Integer.

This PR resolves the first issue and unblocks the internal engineer by enabling them to use `-42595200000|int|datetimeformat` while I continue to investigate the second problem here.

PR #910, opened 2022-09-09
Upgrade jackson and serailize with single quote char

jasmith-hs:
Speed up serialization by not needing to do a regex replace on double quotes. Instead in jackson 2.10, they introduced the ability to specify the quote character. If we set the quote character to be a single quote rather than a double quote, then we don't have to use regex to replace the double quotes later.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
@@ -50,14 +50,27 @@ public class PyishObjectMapper {
   public static String getAsPyishStringOrThrow(Object val)
     throws JsonProcessingException {
     String string = PYISH_OBJECT_WRITER.writeValueAsString(val);
-    Optional<Long> maxStringLength = JinjavaInterpreter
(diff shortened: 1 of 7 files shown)


src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
```diff
@@ -54,14 +57,10 @@ public static String getAsPyishStringOrThrow(Object val)
     if (maxStringLength.map(max -> string.length() > max).orElse(false)) {
       throw new OutputTooBigException(maxStringLength.get(), string.length());
     }
-    String result = string
-      .replace("'", "\\'")
-      // Replace double-quotes with single quote as they are preferred in Jinja
-      .replaceAll("(?<!\\\\)(\\\\\\\\)*(?:\")", "$1'");
-    if (!result.contains("{{")) {
-      return String.join("} ", result.split("}(?=})"));
+    if (string.contains("}}") && !string.contains("{{")) {
```
jboulter, line 60: there are config values for these

jboulter (commented):
scary, but exciting!

PR #915, opened 2022-09-23
Migration to Jakarta EL 4 compatible version of JUEL

dmaidaniuk:
PR for issue https://github.com/HubSpot/jinjava/issues/809

Please note, that here used a fork of JUEL library upgraded to Jakarta EL 4:

https://github.com/UkrSoftTech/juel/tree/jakarta-el-4

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,14 +29,14 @@
         <version>1.14.2</version>
       </dependency>
       <dependency>
-        <groupId>de.odysseus.juel</groupId>
-        <artifactId>juel-api</artifactId>
-        <version>2.2.7</version>
+        <groupId>jakarta.el</groupId>
+        <artifactId>jakarta.el-api</artifactId>
+        <version>4.0.0</version>
       </dependency>
       <dependency>
         <groupId>de.odysseus.juel</groupId>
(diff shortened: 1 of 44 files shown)


dmaidaniuk:
The original repo of JUEL https://github.com/beckchr/juel is not maintained since 2018. So, either the forked repo should be deployed to the Sonatype repository with a changed group id, or sources of JUEL should be included in the Jinjava library.

dmaidaniuk:
@jasmith-hs please take a look.

jasmith-hs:
We will likely lean more towards including the sources into the Jinjava library like in https://github.com/HubSpot/jinjava/pull/810 
Since the original repo isn't maintained. It will be simpler to fix any bugs this way/upgrade to EL 5 if we desire that.

We could also maintain our own fork of JUEL cc @mattcoley @boulter

dmaidaniuk:
@jasmith-hs I can transfer ownership of https://github.com/UkrSoftTech/juel/ repository to HubSpot organization. Also, in that fork I made the migration to EL 5 on another branch https://github.com/UkrSoftTech/juel/tree/jakarta-el-5.

PR #916, opened 2022-09-29
Throw warning when intersection Set elements have mismatched types

AlpriElse:
### Problem
While the `type()` function helps give developers visibility of the types of jinjava variables, sometimes developers infer the types of objects by printing them in an expression eg. `{{ some variable }}`.

This can cause some confusion when variables are evaluated to Strings.

For example, within HubL, `{{ content.topic_list }}` is evaluated to a list of strings, but are actually a list of objects.

### Solution
Throw a Warning to inform a developer that the two Sets being intersected have elements of different type to improve DX.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
@@ -78,7 +78,7 @@ public abstract class AbstractSetFilter implements AdvancedFilter {
         ),
         "list",
         interpreter.getLineNumber(),
-        -1,
+        interpreter.getPosition(),
         null
       )
     );
--- a/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
@@ -2,6 +2,7 @@ package com.hubspot.jinjava.lib.filter;
 
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
+import com.hubspot.jinjava.lib.fn.TypeFunction;
 import com.hubspot.jinjava.util.ForLoop;
 import com.hubspot.jinjava.util.ObjectIterator;
 import java.util.LinkedHashSet;
@@ -28,5 +29,12 @@ public abstract class AbstractSetFilter implements AdvancedFilter {
       result.add(loop.next());
     }
     return result;
+  }
+
+  protected String getTypeOfSetElements(Set<Object> set) {
+    if (set.isEmpty()) {
+      return "null";
+    }
+    return TypeFunction.type(set.iterator().next());
   }
 }
--- a/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
@@ -1,6 +1,7 @@
 package com.hubspot.jinjava.lib.filter;
 
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateError;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.lib.fn.TypeFunction;
 import com.hubspot.jinjava.util.ForLoop;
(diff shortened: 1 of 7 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/IntersectFilter.java
```diff
@@ -35,9 +38,41 @@ public Object filter(
...
+        )
+      );
+    }
+
+    return new ArrayList<>(Sets.intersection(varSet, argSet));
+  }
+
+  private String getTypeOfSetElements(Set<Object> set) {
+    if (set.isEmpty()) {
+      return "null";
```
mattcoley, line 73: Should we make this optional? You should be able to intersect a set with an empty set, however it appears the above logic will add a warning for that.

src/test/java/com/hubspot/jinjava/lib/filter/IntersectFilterTest.java
```diff
@@ -31,4 +34,35 @@ public void itReturnsEmptyOnNullParameters() {
...
+
+    String renderedOutput = interpreter.render("{{ [1, 2, 3]|intersect([1, 2, 3]) }}");
+    assertThat(renderedOutput).isEqualTo("[1, 2, 3]");
+
+    List<TemplateError> errors = interpreter.getErrors();
+    assertThat(errors).isEmpty();
+  }
+
+  @Test
+  public void itThrowsWarningOnMismatchTypes() {
```
mattcoley, line 50: Can we add a unit test for the empty set case to make sure it doesn't add a warning.

src/main/java/com/hubspot/jinjava/lib/filter/IntersectFilter.java
```diff
@@ -35,9 +38,41 @@ public Object filter(
...
   ) {
-    return new ArrayList<>(
-      Sets.intersection(objectToSet(var), objectToSet(parseArgs(interpreter, args)))
-    );
+    Object argObj = parseArgs(interpreter, args);
+
+    Set<Object> varSet = objectToSet(var);
+    Set<Object> argSet = objectToSet(argObj);
+
+    boolean areMismatchedElementTypes = !getTypeOfSetElements(varSet)
```
mattcoley, line 46: Can we do this in the parent `AbstractSetFilter` class?

src/main/java/com/hubspot/jinjava/lib/filter/IntersectFilter.java
```diff
@@ -35,9 +38,41 @@ public Object filter(
...
+      .equals(getTypeOfSetElements(argSet));
+
+    if (areMismatchedElementTypes) {
+      interpreter.addError(
+        new TemplateError(
+          TemplateError.ErrorType.WARNING,
+          TemplateError.ErrorReason.OTHER,
+          TemplateError.ErrorItem.FILTER,
+          String.format(
+            "Mismatched types. `value` elements are of type `%s` and `list` elements are of type `%s`. This may lead to unexpected behavior.",
```
mattcoley, line 56: I am slightly confused by this wording. Maybe something like 
`Mismatched Types: input set has elements of type '%s' but arg set has elements of type '%s'. Use `|map` filter to convert sets to the same type for filter to work correctly.`

mattcoley (changes requested):

src/main/java/com/hubspot/jinjava/lib/filter/IntersectFilter.java
AlpriElse: Kept this existing logic, but reworked the conditional check

AlpriElse:
Moved the warning attachment to `AbstractSetFilter` and implemented throwing the warning in all the implementing classes.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
```diff
@@ -29,4 +32,62 @@ protected Set<Object> objectToSet(Object var) {
...
+  ) {
+    Set<Object> varSet = objectToSet(var);
+    Set<Object> argSet = objectToSet(parseArgs(interpreter, args));
+
+    attachMismatchedTypesWarning(interpreter, varSet, argSet);
+
+    return filter(varSet, argSet);
+  }
+
+  public abstract Object filter(Set<Object> varSet, Set<Object> argSet);
```
AlpriElse, line 51: Now that the concrete implementations of `AbstractSetFilter` have the `filter` method is wrapped here, all of them will have the mismatched type warning.

src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
```diff
@@ -29,4 +32,62 @@ protected Set<Object> objectToSet(Object var) {
...
+        interpreter.getLineNumber(),
+        -1,
+        null
+      )
+    );
+  }
+
+  private String getTypeOfSetElements(Set<Object> set) {
+    if (set.isEmpty()) {
+      return "null";
```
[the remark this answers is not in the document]
AlpriElse: This is borrowed from logic in `TypeFunction` which returns `"null"` as a type when the passed object is null:
https://github.com/HubSpot/jinjava/blob/5cb246370d8f2bdb304fdcfa22cfa57dfab2c14f/src/main/java/com/hubspot/jinjava/lib/fn/TypeFunction.java#L42-L45

src/main/java/com/hubspot/jinjava/lib/filter/AbstractSetFilter.java
```diff
@@ -29,4 +32,59 @@ protected Set<Object> objectToSet(Object var) {
...
+        TemplateError.ErrorReason.OTHER,
+        TemplateError.ErrorItem.FILTER,
+        String.format(
+          "Mismatched Types: input set has elements of type '%s' but arg set has elements of type '%s'. Use |map filter to convert sets to the same type for filter to work correctly.",
+          getTypeOfSetElements(varSet),
+          getTypeOfSetElements(argSet)
+        ),
+        "list",
+        interpreter.getLineNumber(),
+        -1,
```
mattcoley, line 81: You might as well pass in interpreter.getPosition()

mattcoley (approved):
One more comment and then LGTM

PR #917, opened 2022-09-30
Delete unused FormattedDate

tkindy:
I'm looking to do some work around the `datetimeformat` filter and noticed no instances of the `FormattedDate` class were ever constructed. This PR removes formatting logic related to it and the class itself.


hs-lsong:
FormattedDate is a public class. Have you check if it is used by other people?

tkindy:
Ah, good call @hs-lsong, it looks like there are usages of it in other HubSpot code. I'll close this out.

PR #925, opened 2022-10-15
Whitelist -> Safelist in jsoup

rhuddleston:
Whitelist is officially depricated in jsoup 1.15.1+ now and this syntax works on the version of jsoup currently included

This also allow upgrading jsoup to fix a CVE https://github.com/HubSpot/jinjava/pull/914


jasmith-hs:
Thank you

jasmith-hs:
@jowiho , I'll handle this with https://github.com/HubSpot/jinjava/pull/927

PR #928, opened 2022-10-20
Adding EagerCycleTag to Context when Deferred 

nathanielpautzke:
This change adds the cycle tag to the context when in deferred execution mode. Before this change, when eager execution is enable the cycle tag would drop off and never get added to the context. During eager execution, cycle tag will reconstruct and able to be rendered on second pass.


src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerCycleTagTest.java
```diff
@@ -39,4 +44,10 @@ public void eagerSetup() {
   public void teardown() {
     JinjavaInterpreter.popCurrent();
   }
+
+  @Test
+  public void itHandlesDeferredCycle() {
+    String template = "{% for item in deferred %}{% cycle {{foo}},{{bar}} %}{% endfor %}";
+    assertThat(interpreter.render(template)).isEqualTo(template);
+  }
```
jasmith-hs, line 52: This test passes before your change is introduced so it doesn't test the bug

src/test/java/com/hubspot/jinjava/lib/tag/eager/EagerCycleTagTest.java
```diff
@@ -1,13 +1,22 @@
 package com.hubspot.jinjava.lib.tag.eager;
 
+import static org.assertj.core.api.Java6Assertions.assertThat;
```
jasmith-hs, line 3: Use `org.assertj.core.api.Assertions.assertThat` instead

jasmith-hs (approved):
Just the one point about the `assertThat`, but otherwise LGTM

PR #931, opened 2022-10-21
Refactor StrftimeFormatter conversion logic

tkindy:
Depends on #930 

This PR refactors the logic for converting a `strftime`-compatible pattern into a `DateTimeFormatter`.

Previously, this process built a pattern string fit for `DateTimeFormatter.ofPattern`. However, this process isn't flexible enough to implement the format codes for locale-specific date, time, and datetime representations because `DateTimeFormatter` has no equivalent pattern codes.

Instead, this PR switches the conversion logic to use `DateTimeFormatterBuilder`. This sets us up to use `DateTimeFormatterBuilder.appendLocalized` which gives us similar locale-specific representations. I think it also results in cleaner code.


src/main/java/com/hubspot/jinjava/objects/date/StrftimeFormatter.java
```diff
@@ -16,101 +22,90 @@ public class StrftimeFormatter {
...
-        result.append(c);
+      if (c == 'O') {
+        c = strftime.charAt(++i);
+        components = NOMINATIVE_COMPONENTS;
       }
+
+      Optional
+        .ofNullable(components.get(c))
+        .orElseThrow(() -> new InvalidDateFormatException(strftime))
+        .append(builder, stripLeadingZero);
```
AlpriElse, line 105: Just to make sure I'm understanding this right:

The previous behavior when there was an invalid date character was to no-op but now we are a throwing an error?

tkindy: > The previous behavior when there was an invalid date character was to no-op but now we are a throwing an error?

Good eye not sure why I didn't think about this before. You'd be right if this class was being used directly, but most usages of it will be through `Jinjava.render` which catches the error. [I added a test](https://github.com/HubSpot/jinjava/pull/930/files/cd7c4169afd8f5be29c2370fe95c2c189aa8071b..dec4d209792f1719ad6e8f8a1908d43c27d5d319) in the PR before this one to confirm the current behavior.

> ... could we specify which character is invalid in the exception message, so a developer can more easily debug the issue?

Also a good point. I added that context to the error message which you can see in [the updated version of the aforementioned test](https://github.com/HubSpot/jinjava/blob/c6065e7e18a7fae54613ff26bfa53eeab856124a/src/test/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilterTest.java#L169).

> We could also potentially link a dev to docs for all the valid characters they could use in a format string

Since this is in Jinjava and not the wrapping HubL implementation, I don't think it makes sense to link to HubSpot developer docs here.

src/main/java/com/hubspot/jinjava/objects/date/StrftimeFormatter.java
```diff
@@ -16,101 +22,90 @@ public class StrftimeFormatter {
...
-          // maintain the behaviour similar to invalid ascii char <= 255 i.e. append null
-          result.append(conversions[0]);
-        } else {
-          if (stripLeadingZero) {
-            result.append(conversions[c].substring(1));
-          } else {
-            result.append(conversions[c]);
-          }
-        } // < 255
-      } else if (Character.isLetter(c)) {
```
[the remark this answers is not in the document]
tkindy: Sorry, I should've called this out. I included the rationale in [the commit message](https://github.com/HubSpot/jinjava/pull/931/commits/1ec5d217fc174bbfa3dfd969507bf02ff8006c80), but to expand on it:

This loop is for handling literal characters in the format string. Now that we're using `DateTimeFormatterBuilder.appendLiteral` in the `else` branch, we don't need to manually wrap them in single quotes like the previous implementation did (hence the loop).

src/main/java/com/hubspot/jinjava/objects/date/StrftimeFormatter.java
```diff
@@ -16,101 +22,90 @@ public class StrftimeFormatter {
   /*
    * Mapped from http://strftime.org/, http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html
    */
-  private static final String[] CONVERSIONS = new String[255];
-  private static final String[] NOMINATIVE_CONVERSIONS = new String[255];
+  private static final Map<Character, StrftimeConversionComponent> COMPONENTS;
+  private static final Map<Character, StrftimeConversionComponent> NOMINATIVE_COMPONENTS;
```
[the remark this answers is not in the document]
tkindy: I debated that, but at the time I decided it didn't improve readability all that much. If you feel strongly, I can make that change.

src/main/java/com/hubspot/jinjava/objects/date/StrftimeFormatter.java
AlpriElse: > Since this is in Jinjava and not the wrapping HubL implementation, I don't think it makes sense to link to HubSpot developer docs here.

ah :good-point:

PR #934, opened 2022-10-25
add support for numeric keys in map literal

chrisirhc:
Hit this use case in our usage of jinja. 
Putting up a PR in case there's interest in merging this functionality.

This is a hack to support numeric keys declared in map literals.
Most of such usages typically use the keys as if they were strings, so this
hack is likely safe.

A hack is used as opposed to overhauling all maps used since full support would
require overhauling all maps types to `Map<Object, Object>` which seems
overkill.

Behavior is also undefined for when defining same key of same number value and number value `{'1': '1', 1: '2'}`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
@@ -7,6 +7,7 @@ import de.odysseus.el.tree.Bindings;
 import de.odysseus.el.tree.impl.ast.AstIdentifier;
 import de.odysseus.el.tree.impl.ast.AstLiteral;
 import de.odysseus.el.tree.impl.ast.AstNode;
+import de.odysseus.el.tree.impl.ast.AstNumber;
 import de.odysseus.el.tree.impl.ast.AstString;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


chrisirhc:
This is a draft, and I intend to make some changes.

chrisirhc:
Ok, changes are done and ready for review.

src/main/java/com/hubspot/jinjava/el/ext/AstDict.java
```diff
@@ -44,6 +45,11 @@ public Object eval(Bindings bindings, ELContext context) {
...
           key = ((AstIdentifier) entryKey).getName();
         }
+      } else if (entryKey instanceof AstNumber) {
+        // This is a hack to treat numeric keys as string keys in the dictionary.
+        // In most cases this is adequate since the keys are typically treated as
+        // strings.
+        key = entryKey.eval(bindings, context).toString();
       } else {
         throw new TemplateStateException(
           "Dict key must be a string or identifier, was: " + entryKey
```
samukce, line 55: It's good to update the error message here to include the number as a key possibility, what do you think?
```suggestion
          "Dict key must be a string, or identifier, or a  number, was: " + entryKey
```

boulter:
superseded by https://github.com/HubSpot/jinjava/pull/1152

chrisirhc:
Ah apologies I forgot to got back to this.
Glad someone took the fix and addressed the feedback. 👍

PR #942, opened 2022-10-28
Reconstruct all undefined variables when reconstructing call tag

jasmith-hs:
There is a situation where variables defined in a macro function that are implicitly used in a `caller()` call will be treated as `null`, and interpreted when the `caller()` function is reconstructed. To prevent this, we can treat all undefined variables as deferred values as they may be set from the macro function that is being called, as demonstrated in the added test case.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerCallTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerCallTag.java
@@ -1,6 +1,7 @@
 package com.hubspot.jinjava.lib.tag.eager;
 
 import com.hubspot.jinjava.interpret.DeferredMacroValueImpl;
+import com.hubspot.jinjava.interpret.DeferredValue;
 import com.hubspot.jinjava.interpret.InterpretException;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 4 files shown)


nathanielpautzke (approved):

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerCallTag.java
```diff
@@ -127,6 +128,7 @@ public String eagerInterpret(
       StringBuilder result = new StringBuilder(
         prefixToPreserveState.toString() + joiner.toString()
       );
+      interpreter.getContext().setDynamicVariableResolver(s -> DeferredValue.instance());
```
nathanielpautzke, line 131: This may be a dumb question but what does the Call tag do? Looking for it in the docs and struggling to find it

jasmith-hs: It's calls a macro function and then the body of the call tag will be run when within the called macro function `{{ caller() }}` is run.
```
{% macro foo() -%}
1
{{ caller() }}
2
{% endmacro %}
{% call foo() -%}
3
{% endcall %}
```
Would output `132`

PR #943, opened 2022-10-29
Update title filter to ignore special characters

mattcoley:
Technically a deviation from the Jinja2 implementation but a better one IMO. The issue comes when you have special characters like parenthesis: `(Test Company Name)`. Because the old implementation only took the first characters, uppercased it, and then lowercase the rest of the characters `(Test Company Name)|title` would output `(test Company Name)`. Instead we iterate over the characters until we find the first one that is uppercaseable, uppercase that, and then lowercase the rest of the chars.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/TitleFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/TitleFilter.java
@@ -1,10 +1,10 @@
 package com.hubspot.jinjava.lib.filter;
 
+import com.google.common.base.Splitter;
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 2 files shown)


AlpriElse (approved):

src/main/java/com/hubspot/jinjava/lib/filter/TitleFilter.java
```diff
@@ -30,10 +29,36 @@ public String getName() {
...
     }
-    return var;
+
+    String value = var.toString();
+
+    char[] chars = value.toCharArray();
+    boolean titleCased = false;
+
+    for (int i = 0; i < chars.length; i++) {
+      if (chars[i] == ' ') {
```
jasmith-hs, line 42: I think you want to use `Character.isWhitespace(chars[i])` to handle `\n`, `\t`, etc.

jasmith-hs: Unfortunate timing @mattcoley . Got merged while I was typing this comment

mattcoley: Good catch, I reverted the PR and will add some test cases for tabs and new lines.

PR #948, opened 2022-11-04
add NPE checks on limited collections

boulter:
We've seen some NPEs here even though the method is annotated with `@NotNull`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
@@ -9,6 +9,7 @@ import com.hubspot.jinjava.objects.PyWrapper;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import javax.annotation.Nonnull;
 
 public class SizeLimitingPyList extends PyList implements PyWrapper {
   private int maxSize;
@@ -20,6 +21,9 @@ public class SizeLimitingPyList extends PyList implements PyWrapper {
 
   public SizeLimitingPyList(List<Object> list, int maxSize) {
     super(list);
+    if (list == null) {
+      throw new IllegalArgumentException("list is null");
+    }
     if (maxSize <= 0) {
       throw new IllegalArgumentException("maxSize must be >= 1");
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/objects/collections/SizeLimitingPyList.java
```diff
@@ -31,25 +35,31 @@ public SizeLimitingPyList(List<Object> list, int maxSize) {
...
   @Override
-  public boolean add(Object element) {
+  public boolean add(@Nonnull Object element) {
     checkSize(size() + 1);
     return super.add(element);
   }
 
   @Override
-  public void add(int index, Object element) {
+  public void add(int index, @Nonnull Object element) {
```
jasmith-hs, line 44: Why would you not be able to add a `null` object to a list?
I think this would break some usages because we currently allow null objects in lists.

boulter: Good point, though I don't think the annotation will break anything, especially when called from jinjava templates.

jasmith-hs: Actually, I don't think this annotation actually does anything during runtime so it wouldn't break anything, but my point remains about allowing `null` elements

boulter: Actually, I added this only because of the intellij warning `Not annotated parameter overrides @ParametersAreNonnullByDefault parameter`

jasmith-hs: I don't see that annotation, but I see ForwardingList annotation [@ParametricNullness](https://cs.android.com/android/platform/superproject/+/master:external/guava/android/guava/src/com/google/common/hash/ParametricNullness.java):
> Marks a "top-level" type-variable usage as the closest we can get to "non-nullable when non-nullable; nullable when nullable"

I don't understand what that means

jasmith-hs: Seems okay either way, really

PR #950, opened 2022-11-08
Reconstructing namespace from set tags with dotted variables

jasmith-hs:
There is a bug where if there's a namespace created and then one of the keys on the namespace is set to a deferred value within an inline set tag or a block set tag, the namespace will not be reconstructed. And the output would just have `{% set ns1.b = 'b' ~ deferred %}` so `ns1` would be `null`.
This fixes that by making sure that the variables that have dots in them are reconstructed before the rest of the set tag is included in the output.
This makes it so that they function effectively the same way as if you were to update a namespace/map using a DoTag.


nathanielpautzke (approved):

src/test/resources/eager/reconstructs-namespace-for-set-tags-using-period.jinja
```diff
@@ -0,0 +1,10 @@
+{% set ns1 = namespace({'a': 'aa'}) %}
+{% set ns1.b = 'b ' ~ deferred %}
+
+{% set ns2 = namespace({'c': 'cc'}) %}
+{% set ns2.d -%}
```
hs-lsong, line 5: I am not familiar with this? Is this setting `ns2.d = null` ?

PR #953, opened 2022-11-14
Add Babel-like datetime formatting filters

tkindy:
This PR adds three new filters for formatting datetime objects using either built-in or custom formats using [the Unicode Locale Data Markup Language](https://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns).

These filters are inspired by [functions of the same names](https://babel.pocoo.org/en/latest/dates.html) from Python's popular Babel i18n library.

The new `format_datetime` filter technically supersedes the existing `datetimeformat` filter (and its alias, `date`). The latter uses the C/Unix `strftime` pattern syntax as opposed to LDML, but accomplishes essentially the same task. As such, I've marked `datetimeformat` as deprecated. For backwards compatibility, it'll stick around for the foreseeable future, probably forever.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/DateTimeFormatFilter.java
@@ -7,6 +7,10 @@ import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.lib.fn.Functions;
 import com.hubspot.jinjava.objects.date.StrftimeFormatter;
 
+/**
+ * @deprecated Superseded by {@link com.hubspot.jinjava.lib.filter.time.FormatDatetimeFilter}
+ */
+@Deprecated
 @JinjavaDoc(
   value = "Formats a date object",
   input = @JinjavaParam(
@@ -38,7 +42,8 @@ import com.hubspot.jinjava.objects.date.StrftimeFormatter;
     @JinjavaSnippet(
       code = "{% content.updated|datetimeformat('%a %A %w %d %e %b %B %m %y %Y %H %I %k %l %p %M %S %f %z %Z %j %U %W %c %x %X %%') %}"
     )
-  }
+  },
+  deprecated = true
 )
 public class DateTimeFormatFilter implements Filter {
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/time/DateTimeFormatHelper.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.IllformedLocaleException;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.function.Function;
+
+final class DateTimeFormatHelper {
+  private final String name;
+  private final Function<FormatStyle, DateTimeFormatter> cannedFormatterFunction;
+
+  DateTimeFormatHelper(
```
mattcoley, line 21: Personally not a fan of implied access control but curious to hear your perspective

mattcoley (approved):
Great work here just had one comment on the access modifier but LGTM with the solid tests as well

src/main/java/com/hubspot/jinjava/lib/filter/time/DateTimeFormatHelper.java
tkindy: I wanted this class to be an implementation detail and exclude it from the public API of Jinjava to avoid third parties relying on it. I think that makes package-private (not using an access modifier) the right tool for the job here.

Technically, as long as the class is package-private and `final` (which it is), I could make the constructor and methods `public` and they still can't be accessed by anything outside the package. But I think sticking with package-private helps communicate that this class is only meant to be used in this package.

Does that make sense to you?

src/main/java/com/hubspot/jinjava/lib/filter/time/DateTimeFormatHelper.java
```diff
@@ -0,0 +1,87 @@
...
+  String format(Object var, JinjavaInterpreter interpreter, String... args) {
+    String format = arg(args, 0).orElse("medium");
+    ZoneId zoneId = arg(args, 1).map(this::parseZone).orElse(ZoneOffset.UTC);
+    Locale locale = arg(args, 2)
+      .map(this::parseLocale)
+      .orElseGet(() -> interpreter.getConfig().getLocale());
+
+    return buildFormatter(format)
+      .withLocale(locale)
+      .format(Functions.getDateTimeArg(var, zoneId));
```
boulter, line 38: @tkindy How catastrophic do you think it would be to stop defaulting to the current datetime if `var` is null for these new datetime filters? This would eliminate an introduction of a deferred value and allow us to make a clean transition off of the `datetimeformat` filter.

tkindy: Sorry, just seeing this.

I think that'd be fine, I can't imagine anyone's using the default time here. Want me to make that change?

PR #960, opened 2022-11-21
Run in deferred execution mode if output is too big

jasmith-hs:
For all EagerTagDecorators, we can ensure that the output is checked to not be larger than `maxOutputSize`, and if so we can throw an `OutputTooBigException`. We can catch this as well as any that may have been thrown from building string builders and try interpreting the tag with deferred execution mode (that is what is done in `eagerInterpret`).

For example, if there's a for loop that generates a very large output when using eager execution, we could defer the evaluation of all the for loop's iterations. There are other tags that this is applicable to as well, especially some of HubSpot's HubL-specific tags.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
@@ -9,7 +9,6 @@ import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
-import com.hubspot.jinjava.interpret.OutputTooBigException;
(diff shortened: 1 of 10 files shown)


src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
```diff
@@ -51,12 +50,7 @@ public static String getAsPyishStringOrThrow(Object val)
     throws JsonProcessingException {
     String string = PYISH_OBJECT_WRITER.writeValueAsString(val);
     Optional<JinjavaInterpreter> interpreterMaybe = JinjavaInterpreter.getCurrentMaybe();
-    Optional<Long> maxStringLength = interpreterMaybe
-      .map(interpreter -> interpreter.getConfig().getMaxStringLength())
```
jasmith-hs, line 55: Switched from `getMaxStringLength` to `getMaxOutputSize` because the max string length is used for truncating in the join filter, rather than throwing an output too big exception

hs-lsong (approved):

PR #961, opened 2022-11-22
Acquire deferred words from nested expressions

jasmith-hs:
Nested expressions inside of a deferred tag or expression have been problematic because the deferred words are not marked for the contents of those expressions.
We can mark the words in nested expressions by parsing them and looking at the expression nodes' expressions to recursively gather the deferred words.
This improves eager execution's support of nested interpretation.

For example, if I have code like:
```
{% set foo = 'a' %}
{% do deferred.append('{{ foo }}') %}
{{ deferred[0] }}
```
The end result should look like:
```
{% set foo = 'a' %}
{% do deferred.append('{{ foo }}') %}
{{ deferred[0] }}
```
We need to make sure to reconstruct `foo`, otherwise `{{ foo }}` would render as `""`


src/test/resources/eager/reconstructs-words-from-inside-nested-expressions.expected.jinja
```diff
@@ -0,0 +1,2 @@
+{% set foo = 'a' %}{% do deferred.append('{{ foo }}') %}
```
hs-lsong, line 1: After rendering should this become
```
{% do deferred.append('a') %}
{{ deferred[0] }}
```

jasmith-hs: It would not, because if you were to do:
```
{% set my_list = [] %}
{% do my_list.append('{{ 1 + 2 }}') %}
Printed: {% print my_list[0] %}
Expressioned: {{ my_list[0] }}
```
The output is:
```
Printed: {{ 1 +  2 }}
Expressioned: 3
```

It only does nested interpretation if it's used inside of an expression node. So we shouldn't evaluate `{{ foo }}` yet.

I can make this more clear by updating the unit test/adding another one that demonstrates the second pass

hs-lsong: This is a bit confusing now. What do we expect this?
```
{% set foo = 'hello' %}
Printed: {% print foo %}
Expressioned: {{ foo }}
```

jasmith-hs: That will be:
```
Printed: hello
Expressioned: hello
```

jasmith-hs: It is confusing that jinjava supports nested interpretation, but only inside expression nodes... But that is how it was designed

hs-lsong: OK. So only expression nested in `{{ }}` will be interpreted. Thanks.

hs-lsong (approved):

PR #968, opened 2022-12-05
Add null check before collecting to a map in eager

manheychiu:
This is causing `NullPointerException` in `java.util.stream.Collectors::uniqKeysMapAccumulator`. Therefore, a null check is added.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
@@ -29,6 +29,7 @@ import com.hubspot.jinjava.objects.serialization.PyishObjectMapper;
 import com.hubspot.jinjava.tree.TagNode;
 import com.hubspot.jinjava.tree.parse.TagToken;
 import com.hubspot.jinjava.util.EagerExpressionResolver.EagerExpressionResult;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -104,7 +104,9 @@ public static EagerExecutionResult executeInChildContext(
           .filter(e -> !metaContextVariables.contains(e.getKey()))
           .filter(
             entry ->
-              !(entry.getValue() instanceof DeferredValue) && entry.getValue() != null
+              !(entry.getValue() instanceof DeferredValue) &&
+              entry.getValue() != null &&
+              getObjectOrHashCode(entry.getValue()) != null
```
jasmith-hs, line 109: `getObjectOrHasCode()` is sometimes an expensive call, so it would be good to only have to call it once:
```
          .filter(e -> !(e.getValue() instanceof DeferredValue))
          .map(
            e ->
              new AbstractMap.SimpleImmutableEntry<>(
                e.getKey(),
                getObjectOrHashCode(e.getValue())
              )
          )
          .filter(e -> e.getValue() != null)
          .collect(
            Collectors.toMap(Entry::getKey, Entry::getValue)
          )

jasmith-hs (changes requested):
Good find, but I just have one concern

src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
manheychiu: Thanks for pointing that out. I have changed that. Could you please have another look?

Joeoh (approved):

PR #969, opened 2022-12-05
Register UnescapeHtmlFilter

jasmith-hs:
Follow up on https://github.com/HubSpot/jinjava/pull/967
@boulter


boulter:
Thanks. I guess it's been a while since I added a new filter. For some reason I thought they were automatically registered with guice, but I guess that's just tags.

PR #972, opened 2022-12-09
Fix stack overflow for self-nested pymap.

hs-lsong:
We will run to stack overflow which computing the hash code with the default method if the PyMap has nested entries. 
It is still a mystery that the nested entry can be created since we have the checks to prevent it.
https://github.com/HubSpot/jinjava/blob/33c3e328aeb70bde9dd2fb04a2d0640b26ea25b6/src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java#L22
and https://github.com/HubSpot/jinjava/blob/33c3e328aeb70bde9dd2fb04a2d0640b26ea25b6/src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java#L46

Thus, I am not able to make a test to create such map.

Here is the example code that will cause this issue:
```
{% set context = i.icon.image %}
{% do context.update({"image" : i.icon.image, "type" : 'raw', "border_radius" : 'border-radius-none', "class" : ''}) %}
{{ contentImage(context) }}
```
Where `contentImage` is a macro:
```
{% macro contentImage(data) %}
	{% set args = {
		"image" 		: {},
		"type" 			: "plain",
	} %}
	{% do args.update(data) %}
```

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java
@@ -58,4 +58,15 @@ public class PyMap extends ForwardingMap<String, Object> implements PyWrapper {
     }
     super.putAll(m);
   }
+
+  @Override
+  public int hashCode() {
+    int h = 0;
+    for (Entry<String, Object> entry : map.entrySet()) {
+      if (entry.getValue() != map && entry.getValue() != this) {
+        h += entry.hashCode();
+      }
+    }
+    return h;
+  }
 }


jasmith-hs:
I don't think this will prevent the StackOverflow. Here's a reproduction with this, or similar with a list `{% set list = [map] %}`
```
@Test
  public void itDoesNotHaveInfiniteRecursionInHashCode() {
    RenderResult renderResult = jinjava.renderForResult(
      "{% set map = {} %}" +
      "{% set map2 = {'bar': map} %}" +
      "{% do map.put('foo', map2) %}",
      Collections.emptyMap()
    );
    assertThat(renderResult.getContext().get("map").hashCode()).isEqualTo(1);
  }
```

hs-lsong:
It does not prevent all stack overflow, but it will stop the direct nested ones. The test you have above is an indirect nested.

src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java
```diff
@@ -58,4 +58,15 @@ public void putAll(Map<? extends String, ? extends Object> m) {
     }
     super.putAll(m);
   }
+
+  @Override
+  public int hashCode() {
```
boulter, line 63: what's the default implementation of this look like? You could create an implementation that kept a set of all the hashcodes.

hs-lsong: default implementation ```
public int hashCode() {
        int h = 0;
        for (Entry<K, V> entry : entrySet())
            h += entry.hashCode();
        return h;
    }
```

**Comment by jasmith-hs** [comment]
> It does not prevent all stack overflow, but it will stop the direct nested ones. The test you have above is an indirect nested.

Directly nested ones are already prevented by the insertion prevention logic you mentioned. However, the `putAll` method is not protected enough. We could do something like this. Or it could throw an `IllegalArgumentException`
```
@Override
  public void putAll(Map<? extends String, ? extends Object> m) {
    if (m == this) {
      throw new IllegalArgumentException(
        "Map putAll() operation can't be used to add map to itself"
      );
    }
    super.putAll(
      map
        .entrySet()
        .stream()
        .filter(entry -> entry.getValue() != this)
        .collect(Collectors.toMap(Entry::getKey, Entry::getValue))
    );
  }
```

PR #973, opened 2022-12-13
Handle alternate exptest syntax for is false and is true exp tests

jasmith-hs:
Expression tests like `exptest:false.evaluate(false, ____int3rpr3t3r____)` were failing to parse because `false` is not parsed as an `IDENTIFIER` symbol. 
This PR fixes that problem and adds some tests around it.

cc @hs-lsong


hs-lsong (approved):
Thanks.

PR #974, opened 2022-12-13
Refactor EagerReconstructionUtils.executeInChildContext

hs-lsong:
This function is too long (we will get lint error if >200 lines). Moved a few code blocks to separate functions for simpler reading of the code.

The only logic change is re-ordering the loop checking to changed values.
We used to loop in the `context`'s entrySet, and filter by initiallyResolvedHashes.
```
interpreter
          .getContext()
          .entrySet()
          .stream()
          .filter(e -> initiallyResolvedHashes.containsKey(e.getKey()))
```
However, `interpreter.getContext().entrySet()` can be very slow because of nested `ScopeMap`.
The PR changes the above to loop through `initiallyResolvedHashes`, which is only a simple map. 
```
Set<Entry<String, Object>> initialEntrySet = initiallyResolvedHashes.entrySet();
    Context context = interpreter.getContext();
    Map<String, Object> changedValues = new HashMap<>();
    for (Entry<String, Object> entry : initialEntrySet) {
      Object newValue = context.get(entry.getKey());
      if (newValue == null) {
        continue;
      }
```

This bring a performance boost for some pages. One particular page running locally (M1 cpu) reduces from about 13 sec rendering time to about 10 sec. Still very slow though.  https://tools.hubteam.com/api/request/4661918824ab40a49b4e437db854832b

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
@@ -29,6 +29,7 @@ import com.hubspot.jinjava.objects.serialization.PyishObjectMapper;
 import com.hubspot.jinjava.tree.TagNode;
 import com.hubspot.jinjava.tree.parse.TagToken;
 import com.hubspot.jinjava.util.EagerExpressionResolver.EagerExpressionResult;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -190,89 +121,214 @@ public static EagerExecutionResult executeInChildContext(
...
+    } else {
+      entryStream =
+        interpreter
+          .getContext()
+          .entrySet()
+          .stream()
+          .filter(entry -> initiallyResolvedHashes.containsKey(entry.getKey()))
+          .filter(
+            entry -> EagerExpressionResolver.isResolvableObject(entry.getValue(), 4, 400) // TODO make this configurable
+          );
```
jasmith-hs, line 215: Using something like:
```
initiallyResolvedHashes.keySet()
          .stream()
          .map(key -> new AbstractMap.SimpleImmutableEntry(key, interpreter.getContext().get(key))
          .filter(
            entry -> EagerExpressionResolver.isResolvableObject(entry.getValue(), 4, 400) // TODO make this configurable
          );
```
May be faster too.
Or, simply storing `interpreter.getContext().entrySet()` as a variable in the beginning of this method so that we only need to go through the scope map once in the whole method.

jasmith-hs (commented):
Nice find! I have one suggestion too that may help even more

src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
hs-lsong: > we only need to go through the scope map once in the whole method

Do you mean the `executeInChildContext` method? I think the `before` and `after` the `function.apply`, the entrySet of the context may have changed, we need to go through the scopes again.

jasmith-hs: https://github.com/HubSpot/jinjava/pull/974/commits/9a912af026e807a3019b298b1700a9f78f97697d is exactly what I meant 👍

jasmith-hs (approved):

PR #981, opened 2023-01-03
Add null check to handle null lazy expressions

jasmith-hs:
This was an unintended side-effect of https://github.com/HubSpot/jinjava/pull/974
cc @hs-lsong @tkindy


hs-lsong (approved):

tkindy (approved):
Thanks Jack!

Issue #987, opened 2023-01-12
Serialize maps/lists when reconstructing lazily in case element(s) are updated


jasmith-hs:
```
{% set list = [] %}
{% set map = {} %}
{% do list.append(map) %}
{% if deferred %}
{% do list.append(1) %}
{% endif %}
{% do map.update({'a': 'A'}) %}
{{ list }}
```
This currently becomes:
```
{% set list = [{} ] %}{% if deferred %}
{% do list.append(1) %}
{% endif %}

{{ list }}
```
When it should be more like:
```
{% set list = [{'a': 'A'} ] %}{% if deferred %}
{% do list.append(1) %}
{% endif %}

{{ list }}
```
Or at least create a deferred node so that we know that something went wrong. The best way I can currently think of to solve this is to make the serialization which is part of the set-tag reconstruction lazy.
This would require a great deal of refactoring so I don't know if it is worth it to solve.

jasmith-hs:
Here's another example:
```
{% set references = [] %}
{% macro appender(refArray) %}
{% for i in range(1) %}
{% if deferred %}
{% do refArray.append(1) %}
{% endif %}
{% endfor %}
{{ caller() }}
{% endmacro %}
{% call appender(references) %}
{% set ref = references[0] %}
{{ ref }}
{% endcall %}
{{ references }}
```

PR #988, opened 2023-01-12
Preserve identifiers when reconstructing a tag which sets a value

jasmith-hs:
When reconstructing a tag which has the function of setting a variable to the result of an expression (both `set` and `for` do this), we should preserve the AstIdentifiers in the expression rather than resolving them because they may be directly referencing a value which can be modified later on. If instead, we set our variable to a serialized version of the object, then if that object is modified, the changes will not be reflected in the set value.

The simplest example of this is shown in one of the new tests:
```
{% set dict = {} %}

{% set foo = (dict, deferred) %}

{% do dict.update({'a': 'b'}) %}

{{ foo }}
```
If we don't preserve the identifier for `dict`, then we end up with:
```
{% set dict = {} %}

{% set foo = ({}, deferred) %}

{% do dict.update({'a': 'b'}) %}

{{ foo }}
```
Which would result in `foo` not having the correct value within it!

I realised that this can also happen if a value is reconstructed and then later one of the elements inside of that value is updated. Such a case is much more difficult to handle. I created a separate issue for that: https://github.com/HubSpot/jinjava/issues/987

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/eager/EagerAstIdentifier.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/eager/EagerAstIdentifier.java
@@ -3,7 +3,6 @@ package com.hubspot.jinjava.el.ext.eager;
 import com.hubspot.jinjava.el.ext.DeferredParsingException;
 import com.hubspot.jinjava.el.ext.ExtendedParser;
 import com.hubspot.jinjava.interpret.DeferredValueException;
-import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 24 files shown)


jasmith-hs:
Here's another example:
```
{% set references = [] %}
{% macro appender(refArray) %}
{% for i in range(1) %}
{% if deferred %}
{% do refArray.append(1) %}
{% endif %}
{% endfor %}
{{ caller() }}
{% endmacro %}
{% call appender(references) %}
{% set ref = references[0] %}
{{ ref }}
{% endcall %}
{{ references }}
```

PR #996, opened 2023-01-23
Get best invoke method based on parameters

jasmith-hs:
I copied the code of `BeanELResolver` to make the licensing easier than integrating the code with JinjavaBeanELResolver.

Using java 17 (Oracle JDK 17.0.2, Eclipse Adoptium 17.0.3, and likely more) changes the order in which `String.class.getMethods()` returns methods, which causes the `public String replace(char oldChar, char newChar)` method to be invoked when calling `BeanELResolver.invoke(context, "some string", "replace", null, new Object[]{"some", ""})`. This makes the resulting string `␀ome ␀tring` instead of `string`.

This PR makes it so that we check that the parameter types of the method we select are assignable from the input parameters. And if there are multiple methods that could work, pick the one that has the more specific parameter types.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/BeanELResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/BeanELResolver.java
@@ -1,0 +1,719 @@
+/*
+ * Copyright 2006-2009 Odysseus Software GmbH
+ * Modifications Copyright (c) 2023 HubSpot Inc.
+ *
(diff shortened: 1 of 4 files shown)


jboulter (approved):
Nice. Have we found this to be a problem with any other methods that we know of?

jasmith-hs:
I don't know of any other cases where it changed

PR #997, opened 2023-01-25
Limit pyish serialization length

jasmith-hs:
Some objects may have exceptionally large serialized representations. If they're larger than the `maxOutputSize` set on the `JinjavaConfig`, then they won't be allowed into the final output and an `OutputTooBigException` will be thrown. However, the process of serializing objects that would do that can be costly in terms of time and memory. So these changes make it so we can check the size of the output as we are adding to it during the pyish serialization process.
~~This does require some cooperation from classes that implement `PyishSerializable` if they are composed of several elements in order to optimize the limit checking. For example, if there's some `PyishSerializable` container, that serializes itself in a special way while serializing 1000 objects inside of it, we'd want to check the limit before the entire operation has finished, otherwise a string that is far too large may already have been allocated, copied, and added to several times. Instead, such a class could override `writeSelf(jsonGenerator, serializerProvider)` to split each child object into it's own write to the jsonGenerator.~~ That's not needed any more since I made `LengthLimitingStringBuilder` extend appendable, and have that be used by default. So any appending to the `appendable` argument of `appendPyishString(appendable)` will also check against the limit.

The problem of large string copying is also addressed with a new method `<T extends Appendable & CharSequence> T appendPyishString(T appendable)` which can be used to recursively append to a string builder rather than copying large strings when modifying them.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
@@ -2,28 +2,49 @@ package com.hubspot.jinjava.objects.serialization;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.SerializerProvider;
+import java.io.CharArrayWriter;
 import java.io.IOException;
-import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicInteger;
 
-public class MapEntrySerializer extends JsonSerializer<Map.Entry> {
+public class MapEntrySerializer extends JsonSerializer<Entry<?, ?>> {
   public static final MapEntrySerializer INSTANCE = new MapEntrySerializer();
 
   private MapEntrySerializer() {}
 
   @Override
   public void serialize(
-    Map.Entry object,
+    Entry<?, ?> entry,
     JsonGenerator jsonGenerator,
     SerializerProvider serializerProvider
   )
     throws IOException {
-    String key = PyishObjectMapper.PYISH_OBJECT_WRITER.writeValueAsString(
-      object.getKey()
+    AtomicInteger remainingLength = (AtomicInteger) serializerProvider.getAttribute(
+      SizeLimitingWriter.REMAINING_LENGTH_ATTRIBUTE
     );
-    String value = PyishObjectMapper.PYISH_OBJECT_WRITER.writeValueAsString(
-      object.getValue()
-    );
+    String key;
+    String value;
+    if (remainingLength != null) {
+      ObjectWriter objectWriter = PyishObjectMapper.PYISH_OBJECT_WRITER.withAttribute(
+        SizeLimitingWriter.REMAINING_LENGTH_ATTRIBUTE,
+        remainingLength
+      );
+      key = objectWriter.writeValueAsString(entry.getKey());
+      SizeLimitingWriter sizeLimitingWriter = new SizeLimitingWriter(
+        new CharArrayWriter(),
+        remainingLength
+      );
+      objectWriter.writeValue(
+        new SizeLimitingWriter(new CharArrayWriter(), remainingLength),
+        entry.getValue()
+      );
+      value = sizeLimitingWriter.toString();
+    } else {
+      key = PyishObjectMapper.PYISH_OBJECT_WRITER.writeValueAsString(entry.getKey());
+      value = PyishObjectMapper.PYISH_OBJECT_WRITER.writeValueAsString(entry.getValue());
+    }
     jsonGenerator.writeRawValue(String.format("fn:map_entry(%s, %s)", key, value));
   }
 }
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
@@ -36,10 +36,7 @@ public class MapEntrySerializer extends JsonSerializer<Entry<?, ?>> {
         new CharArrayWriter(),
         remainingLength
       );
-      objectWriter.writeValue(
-        new SizeLimitingWriter(new CharArrayWriter(), remainingLength),
-        entry.getValue()
-      );
+      objectWriter.writeValue(sizeLimitingWriter, entry.getValue());
       value = sizeLimitingWriter.toString();
     } else {
       key = PyishObjectMapper.PYISH_OBJECT_WRITER.writeValueAsString(entry.getKey());
--- a/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
+++ b/src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
@@ -22,22 +22,22 @@ public class MapEntrySerializer extends JsonSerializer<Entry<?, ?>> {
   )
     throws IOException {
     AtomicInteger remainingLength = (AtomicInteger) serializerProvider.getAttribute(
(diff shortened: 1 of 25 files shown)


src/main/java/com/hubspot/jinjava/util/WhitespaceUtils.java
```diff
@@ -142,5 +143,16 @@ public static String unwrap(String s, String prefix, String suffix) {
     return s.substring(start + prefix.length(), end - suffix.length() + 1);
   }
 
+  @Nullable
+  public static StringBuilder quoteIfNotNull(CharSequence charSequence) {
```
jboulter, line 147: where is this used?

src/main/java/com/hubspot/jinjava/objects/serialization/PyishSerializable.java
```diff
@@ -1,27 +1,64 @@
...
+    JsonGenerator jsonGenerator,
+    SerializerProvider serializerProvider
+  )
+    throws IOException {
+    AtomicInteger remainingLength = (AtomicInteger) serializerProvider.getAttribute(
+      LengthLimitingWriter.REMAINING_LENGTH_ATTRIBUTE
+    );
+    jsonGenerator.writeRawValue(
+      appendPyishString(
+          remainingLength != null
```
jboulter, line 56: nit: reverse the expression here to make it easier to understand.

src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
```diff
@@ -40,16 +45,50 @@ public static String getAsUnquotedPyishString(Object val) {
...
-    return string;
+  public static String getAsPyishStringOrThrow(Object val) throws IOException {
+    ObjectWriter objectWriter = PYISH_OBJECT_WRITER;
+    Writer writer;
+    Optional<Long> maxOutputSize = JinjavaInterpreter
+      .getCurrentMaybe()
+      .map(interpreter -> interpreter.getConfig().getMaxOutputSize())
+      .filter(max -> max > 0);
+    if (maxOutputSize.isPresent()) {
+      AtomicInteger remainingLength = new AtomicInteger(
```
jboulter, line 78: I thought jinjava interpreters were single threaded?

jboulter (approved):

src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
jasmith-hs: They are, I am using this and modifying the value rather than replacing the value every time we want to update it.
It simplifies what we have to do in the `LengthLimitingWriter` since we can just store the AtomicInteger as an instance variable, rather than a `serializerProvider`
When I was trying another approach, limiting both depth and width of serialization, I created a separate object that held two AtomicIntegers, but since there's just a since AtomicInteger I didn't make a wrapper class.

src/main/java/com/hubspot/jinjava/util/WhitespaceUtils.java
jasmith-hs: It's not actually used in this repository, but it's a helper method for classes that extend PyishSerializable. Since all the other logic for checking if strings are quoted and removing quotes from strings lives here, I figured it would be nice to have the reverse

src/main/java/com/hubspot/jinjava/util/LengthLimitingStringBuilder.java
```diff
@@ -50,14 +51,41 @@ public void append(Object obj) {
...
+    checkLength();
+    builder.append(csq);
+    return this;
+  }
+
+  @Override
+  public Appendable append(CharSequence csq, int start, int end) {
+    int csqLength = 4; // null
+    if (csq != null) {
+      csqLength = end - start;
```
hs-lsong, line 70: Do you need to check if `end >= start`?

src/test/java/com/hubspot/jinjava/lib/filter/EscapeFilterTest.java
```diff
@@ -54,7 +54,7 @@ public void testNewStringReplaceIsFaster() {
 
     assertThat(newResult).isEqualTo(oldResult);
     System.out.printf("New: %d Old:%d\n", newTime.toMillis(), oldTime.toMillis());
-    int speedUpFactor = 2; // On M1, it is between 50 and 100 times faster.
+    int speedUpFactor = getVersion() < 17 ? 2 : 1; // On M1, it is between 50 and 100 times faster. Difference is much smaller on java 17
```
hs-lsong, line 57: 👍

src/main/java/com/hubspot/jinjava/util/LengthLimitingStringBuilder.java
jasmith-hs: `StringBuilder` will throw an IndexOutOfBoundsException if they're invalid so we don't need to

src/main/java/com/hubspot/jinjava/objects/serialization/MapEntrySerializer.java
```diff
@@ -2,28 +2,46 @@
...
+      LengthLimitingWriter.REMAINING_LENGTH_ATTRIBUTE
     );
+    String key;
+    String value;
+    if (remainingLength != null) {
+      ObjectWriter objectWriter = PyishObjectMapper.PYISH_OBJECT_WRITER.withAttribute(
+        LengthLimitingWriter.REMAINING_LENGTH_ATTRIBUTE,
+        remainingLength
+      );
+      key = objectWriter.writeValueAsString(entry.getKey());
```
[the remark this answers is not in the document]
jasmith-hs: It will count towards the limit. There is a LengthLimitingWriter already on `jsonGenerator` so `writeRawValue` will count towards the value.
Maps that are created in jinjava are `Map<String, Object>` so writing the keys of that is a single step process that won't be recursive. Writing the Objects can be recursive as the object can contain many other objects. So by creating another LengthLimitingWriter here for the `entry.value`, if the object is very expensive to serialize, we are able to break before the `entry.value` is entirely serialized. If we didn't create a new LengthLimitingWriter here, then we may have to wait for the entire `entry.value` to be serialized to check if it's too long.

I could do the same thing for the `key`, but I am trying to be as optimized as possible, and since the keys' serialization are almost never recursive, I'm not putting an additional layer of limit checking over them

hs-lsong (approved):

src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java
```diff
@@ -40,16 +45,50 @@ public static String getAsUnquotedPyishString(Object val) {
...
+      if (e instanceof LengthLimitingJsonProcessingException) {
+        if (
+          JinjavaInterpreter
+            .getCurrentMaybe()
+            .map(
+              interpreter -> interpreter.getConfig().getExecutionMode().useEagerParser()
+            )
+            .orElse(false)
+        ) {
+          throw new DeferredValueException(String.format("%s: %s", e.getMessage(), val));
```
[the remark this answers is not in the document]
jasmith-hs: This is a very good question. It makes me realise I should just add a different method instead. When doing eager execution, if we can't serialize a value properly, then we aren't reconstructing properly for the second render pass. Throwing a DeferredValueException will cause the node in which this is being rendered to become a deferred node.
This will let us know that the pre-render needs to be `REDO`.

But that logic isn't necessary when not using eager execution so I am instead going to make a separate method called `getAsPyishStringOrDefer(Object)` which will wrap the exception in a DeferredValueException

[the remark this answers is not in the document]
jasmith-hs: Hmm, I think what would be even better is just wrapping the exception in `OutputTooBigException` and then scrapping the pre-render if there are any output too big errors when rendering

hs-lsong (approved):

PR #999, opened 2023-01-27
Create format_number filter

julia-uy:
Since locales render numbers differently, there needs to be a way to format numbers based on the locale. Currently, the only way is to use the `format_currency` filter and get rid of the currency symbol. 

This PR creates a `format_number` filter that will do this formatting natively. 

`format_number(locale, decimalPrecisionNumber)` where `locale` is String of an ISO language code and `maxDecimalPrecision` is a number that determines the number of decimal digits of the formatted input value. 

Examples:
- `{{ 1000|format_number('en-US') }}` ==> 1,000
- `{{ 1000.333|format_number('en-US') }}` ==> 1,000.333
- `{{ 1000.333|format_number('en-US', 2) }}` ==> 1,000.33

- `{{ 1000|format_number('fr') }}` ==> 1 000
- `{{ 1000.333|format_number('fr') }}` ==> 1 000,333
- `{{ 1000.333|format_number('fr', 2) }}` ==> 1 000,33

- `{{ 1000|format_number('es') }}` ==> 1.000
- `{{ 1000.333|format_number('es') }}` ==> 1.000,333
- `{{ 1000.333|format_number('es', 2) }}` ==> 1.000,33


src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
```diff
@@ -0,0 +1,110 @@
...
+  params = {
+    @JinjavaParam(
+      value = "locale",
+      desc = "Locale in which to format the number. The default is the page's locale."
+    ),
+    @JinjavaParam(
+      value = "decimal precision number",
+      type = "number",
+      desc = "A number input that determines the decimal precision of the formatted value. If the number of decimal digits from the input value is less than the decimal precision number, use the number of decimal digits from the input value. Otherwise, use the decimal precision number. The default is the number of decimal digits from the input value."
+    )
```
tkindy, line 35: This is really a maximum, right? I think naming it as such would be clearer, something like `max decimal precision`.

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
```diff
@@ -0,0 +1,110 @@
...
+          interpreter.getLineNumber(),
+          e,
+          null,
+          ImmutableMap.of("value", Objects.toString(var))
+        )
+      );
+      return var;
+    }
+
+    int noOfDecimalPlacesInInput = Math.max(0, number.scale());
```
tkindy, line 80: I think `numDecimalPlacesInInput` would be the usual way to name something like this in Java.

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
```diff
@@ -0,0 +1,110 @@
...
+    DecimalFormat df = (DecimalFormat) NumberFormat.getInstance();
+    df.setParseBigDecimal(true);
+
+    return (BigDecimal) df.parseObject(Objects.toString(input));
+  }
+
+  private String formatNumber(
+    Locale locale,
+    BigDecimal number,
+    int noOfDecimalPlacesInInput,
```
tkindy, line 98: Since this parameter's value comes from `number`, I think we can get it inside this method instead of passing it as a separate parameter.

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
```diff
@@ -0,0 +1,110 @@
...
+    int decimalPrecisionNumber
+  ) {
+    NumberFormat numberFormat = NumberFormat.getNumberInstance(locale);
+
+    numberFormat.setMinimumFractionDigits(noOfDecimalPlacesInInput);
```
tkindy, line 103: Do you need to set this, actually? I'd think the default `NumberFormat` would try to show the entire passed-in value by default, which would mean it would use all decimal digits (unless it was over the maximum).

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
```diff
@@ -0,0 +1,110 @@
...
+      desc = "A number input that determines the decimal precision of the formatted value. If the number of decimal digits from the input value is less than the decimal precision number, use the number of decimal digits from the input value. Otherwise, use the decimal precision number. The default is the number of decimal digits from the input value."
+    )
+  },
+  snippets = {
+    @JinjavaSnippet(code = "{{ number|format_number }}"),
+    @JinjavaSnippet(code = "{{ number|format_number(\"en-US\") }}"),
+    @JinjavaSnippet(code = "{{ number|format_number(\"en-US\", 3) }}")
+  }
+)
+public class NumberFormatFilter implements Filter {
```
tkindy, line 43: I think `FormatNumberFilter` would align with how filters are named relative to their representation in Jinjava, no?

tkindy (commented):
Mostly just a couple naming things, but generally LGTM

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
julia-uy: Makes sense! In the main method, I was using this variable as a fallback to the max decimal precision variable if it wasn't passed in, but I ended up switching the max decimal precision variable to an `Optional` so it no longer needs to be a parameter for this method

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
julia-uy: I double checked with the unit tests I wrote and I think you are correct. This doesn't need to be set since by default - it'll just output all the decimal digits unless a maximum less than the number of decimal digits was set

src/main/java/com/hubspot/jinjava/lib/filter/NumberFormatFilter.java
julia-uy: Agreed 👍🏼

src/main/java/com/hubspot/jinjava/lib/filter/FilterLibrary.java
```diff
@@ -82,6 +82,7 @@ protected void registerDefaults() {
       Md5Filter.class,
       MinusTimeFilter.class,
       MultiplyFilter.class,
+      FormatNumberFilter.class,
```
tkindy, line 85: This should go higher up to maintain the alphabetic sort now that the name has changed.

tkindy (approved):

PR #1005, opened 2023-02-09
Add option to EscapeJinjavaFilter to only escape curly braces when part of a token

jasmith-hs:
Allows the `|escape_jinjava` filter to be used on json without breaking it.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
@@ -27,6 +27,14 @@ import org.apache.commons.lang3.StringUtils;
   "Use this filter if you need to display text that might contain such characters in Jinjava. " +
   "Marks return value as markup string.",
   input = @JinjavaParam(value = "s", desc = "String to escape", required = true),
+  params = {
+    @JinjavaParam(
+      value = "all_braces",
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
```diff
@@ -47,8 +55,18 @@ public static String escapeJinjavaEntities(String input) {
     return StringUtils.replaceEach(input, TO_REPLACE, REPLACE_WITH);
   }
 
+  public static String escapeFullJinjavaEntities(String input) {
+    return input
+      .replace("{{", BLBRACE + BLBRACE)
```
jboulter, line 60: should we use https://github.com/HubSpot/jinjava/blob/master/src/main/java/com/hubspot/jinjava/tree/parse/TokenScannerSymbols.java here instead of hard-coding them?

jboulter (approved):

src/main/java/com/hubspot/jinjava/lib/filter/EscapeJinjavaFilter.java
jasmith-hs: It gets a little tricky when using other characters and determining how they should be escaped so I specified in the description that it's only for the default: `default expression, tag, or comment marks`

PR #1006, opened 2023-02-10
Add interface to define pre-process actions for a Node

jasmith-hs:
So that extenders of Jinjava can customise the pre-processing actions to take. These are actions that will be taken before any node gets processed. In HubL, we can extend this to limit resource consumption.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/NodePreProcessor.java
+++ b/src/main/java/com/hubspot/jinjava/el/NodePreProcessor.java
@@ -1,0 +1,8 @@
+package com.hubspot.jinjava.el;
+
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/el/NodePreProcessor.java
```diff
@@ -0,0 +1,8 @@
+package com.hubspot.jinjava.el;
+
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.tree.Node;
+
+public interface NodePreProcessor {
+  void preProcess(Node node, JinjavaInterpreter interpreter);
```
jboulter, line 7: could just be a BiConsumer or a Consumer with a an object with params, no?

jboulter (approved):

PR #1007, opened 2023-02-13
Get current date and time from a provider

jboulter:
It's useful to be able to have a fixed date and time for comparison purposes, so this implements a provider for it and allows setting it in JinjavaConfig. The default of course is to use the current date and time.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/date/PyishDate.java
+++ b/src/main/java/com/hubspot/jinjava/objects/date/PyishDate.java
@@ -60,7 +60,7 @@ public final class PyishDate
                   .map(JinjavaInterpreter::getConfig)
                   .map(JinjavaConfig::getDateTimeProvider)
                   .map(DateTimeProvider::getCurrentTimeMillis)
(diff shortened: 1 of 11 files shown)


src/main/java/com/hubspot/jinjava/objects/date/PyishDate.java
```diff
@@ -50,7 +51,17 @@ public PyishDate(Long epochMillis) {
...
+          Optional
+            .ofNullable(epochMillis)
+            .orElseGet(
+              () ->
+                JinjavaInterpreter
+                  .getCurrentMaybe()
+                  .map(JinjavaInterpreter::getConfig)
+                  .map(JinjavaConfig::getDateTimeProvider)
+                  .map(DateTimeProvider::getCurrentTimeMillis)
+                  .orElse(System.currentTimeMillis())
```
jasmith-hs, line 63:
```suggestion
                  .orElseGet(System::currentTimeMillis)
```

jasmith-hs (approved):

PR #1030, opened 2023-03-16
Implement `{% do %}` blocks and fix `{% set %}` block scoping

jasmith-hs:
These are to inline do tags as set blocks are to inline set tags.
```
{% do %}
  {{ some_macro() }}
  {{ other_macro() }}
{% enddo %}
```
They run the block of code within their body and then discard the output. This is done without going into a child scope so variables that are set here will persist their values outside of the do block. This is unlike a `{% for %}` tag or a macro function execution.
This is also supposed to be unlike a set block, which, in jinja, executes in a child scope. I realised that we were not running set blocks in a child scope and so I wanted to fix that to have parity with jinja. However, we were using set blocks in eager execution to be able to execute code, while ignoring the string output. This is useful for `{% extends %}`, `{% from %}` and `{% import %}` as it would allow us to retain any deferred modifications that happened as a result of using those tags without adding any extra output. But this also hinged on the lack of child scoping that set blocks currently use. So the need for a solution where we could ignore the output, keep the current scope, and fix the scoping of set blocks to match their functionality in jinja arose. Implementing do blocks fit the bill perfectly, and we can continue not using a child scope when executing set blocks with that `__ignored__` variable name so as to not break any previous eager renders that used that.

Additionally, after realising that deferred macro functions should be put into their own scope https://github.com/HubSpot/jinjava/pull/1020, I have been working through another bug the past couple days and found that it's simpler to instead just use the macro function temporary variable functionality I added in https://github.com/HubSpot/jinjava/pull/920, which should scope the modifications done in the macro function properly. That's how I realised that the bug with set blocks existed because that didn't work as it was supposed to.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
@@ -33,8 +33,8 @@ import com.hubspot.jinjava.interpret.TemplateError.ErrorItem;
 import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
 import com.hubspot.jinjava.interpret.TemplateError.ErrorType;
 import com.hubspot.jinjava.interpret.errorcategory.BasicTemplateErrorCategory;
+import com.hubspot.jinjava.lib.tag.DoTag;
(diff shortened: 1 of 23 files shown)


jasmith-hs:
Probably is good to do a minor release after this

PR #1040, opened 2023-03-30
Improve deferred word filtering 

jasmith-hs:
So that deferred words that aren't in the combined scope aren't attempted to be reconstructed in the wrong scope

Follow up on https://github.com/HubSpot/jinjava/pull/1032

This test demonstrates the problem:
```
{% set my_list = [] %}
{% set foo %}
{% do my_list.append(deferred) %}a
{% endset %}
{{ my_list }}
```
We don't want `my_list` to get reconstructed inside the set tag so we should filter the deferred words that we reconstruct before deferring them to only reconstruct those within the current scope.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
+++ b/src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
@@ -130,13 +130,47 @@ public class EagerReconstructionUtils {
     reconstructedValues.putAll(
       reconstructMacroFunctionsBeforeDeferring(deferredWords, interpreter)
     );
+    Set<String> deferredWordBases = filterToRelevantBases(deferredWords, interpreter);
     reconstructedValues.putAll(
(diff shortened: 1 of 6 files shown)


src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -130,15 +130,49 @@ public static Map<String, String> reconstructFromContextBeforeDeferringAsMap(
...
 
+  private static Set<String> filterToRelevantBases(
+    Set<String> deferredWords,
+    JinjavaInterpreter interpreter
+  ) {
+    Map<String, Object> combinedScope = interpreter.getContext().getCombinedScope();
+    Set<String> deferredWordBases = deferredWords
+      .stream()
+      .map(w -> w.split("\\.", 2)[0])
+      .filter(combinedScope::containsKey)
```
jasmith-hs, line 151: This line is the change, the rest is refactoring duplicated code

PR #1044, opened 2023-04-06
NodePostProcessors

boulter:
This is an extension of https://github.com/HubSpot/jinjava/pull/1006 and an alternative to https://github.com/HubSpot/jinjava/pull/1042, allowing the caller to do whatever they want when a node is executed. 

This only covers nodes though, so this could be more generically implemented to add hooks for tags, filters and functions.


boulter:
Closing in favor of https://github.com/HubSpot/jinjava/pull/1045

PR #1048, opened 2023-04-13
Improve for loop explosion success with deferred tokens

jasmith-hs:
Now that we [have better tracking of the PrefixToPreserveState](https://github.com/HubSpot/jinjava/pull/1032), we can separate the collection expression execution from the iterating of the for loop.
This leads to better explosion success where we can now support exploding for loops even when there are modifications inside for loops with deferred tokens
Additionally, for loop evaluation will be faster as we don't need to re-run the for loop.

For this input:
```
{% set my_list = [] %}
{% for i in range(2) %}
{% for j in deferred %}
{% do my_list.append(i) %}
{% endfor %}
{% endfor %}
{{ my_list }}
```
Previously, we'd have this output:
```
{% set my_list = [] %}{% for i in [0, 1] %}
{% for j in deferred %}
{% do my_list.append(i) %}
{% endfor %}
{% endfor %}
{{ my_list }}
```
But now we can get this exploded output:
```
{% set my_list = [] %}{% for __ignored__ in [0] %}
{% for j in deferred %}
{% do my_list.append(0) %}
{% endfor %}

{% for j in deferred %}
{% do my_list.append(1) %}
{% endfor %}
{% endfor %}
{{ my_list }}
```

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
@@ -44,7 +44,6 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.util.ConcurrentModificationException;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.regex.Matcher;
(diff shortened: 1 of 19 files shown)


boulter:
Can you explain why this is better? Also is it a typo that there are two `{% endfor %}` in the last output?

jasmith-hs:
> Can you explain why this is better practically? What does explosion mean?

This is better because it means that in a loop `item in collection`, we don't have to defer `item`, so we can resolve anything that depends on `item`.

Exploding just means that we're taking something like:`{% for i in range(5) %}{{ i }},{% endfor %}` and turning it to `0,1,2,3,4`. So the difference for eager execution is whether we have to preserve the `for` token, or if we can render each of the iterations out, potentially leaving some deferred tokens in the result.

PR #1050, opened 2023-04-14
Fix CVE dependency issue

CVEDetect:
Fix issue #1049  by update dependency jackson-dataformat-yaml to 2.14.0 @jboulter @jasmith-hs

PR #1051, opened 2023-04-14
upgrade jackson

boulter:
Upgrades to avoid CVE mentioned in https://github.com/HubSpot/jinjava/pull/1050

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,17 +61,17 @@
       <dependency>
         <groupId>com.fasterxml.jackson.core</groupId>
         <artifactId>jackson-annotations</artifactId>
-        <version>2.12.6</version>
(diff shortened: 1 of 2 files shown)


jasmith-hs (changes requested):
Can you make sure that this logic still works:
https://github.com/HubSpot/jinjava/blob/89cd76833372605a125dbb35528f2b938c1db5e0/src/main/java/com/hubspot/jinjava/objects/serialization/PyishObjectMapper.java#L69-L74

jowiho:
In the meantime Jackson 2.15.2 is the latest stable version. I've confirmed that this PR builds and tests correctly with that Jackson version as well.

PR #1064, opened 2023-05-01
add warnings when datetime-related filters called with null arguments

boulter:
In the future we may change this behavior so null args do not default to the current time. I suspect many users didn't know this was happening since there was no warning.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/filter/BetweenTimesFilterTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/filter/BetweenTimesFilterTest.java
@@ -9,15 +9,9 @@ import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.Map;
-import org.junit.Before;
 import org.junit.Test;
 
 public class BetweenTimesFilterTest extends BaseJinjavaTest {
-
-  @Before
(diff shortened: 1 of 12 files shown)


src/test/java/com/hubspot/jinjava/lib/filter/BetweenTimesFilterTest.java
```diff
@@ -9,16 +9,10 @@
...
 import java.time.ZonedDateTime;
 import java.util.Map;
-import org.junit.Before;
 import org.junit.Test;
 
 public class BetweenTimesFilterTest extends BaseJinjavaTest {
 
-  @Before
-  public void setup() {
-    jinjava.getGlobalContext().registerClasses(EscapeJsFilter.class);
```
boulter, line 19: seems like this was copied from some other test

jasmith-hs (approved):
Want to add a test that the warning gets added?

boulter:
> Want to add a test that the warning gets added?

Oh weird, I had one. Where did it go?

boulter:
Here it is https://github.com/HubSpot/jinjava/pull/1064/files#diff-2740cb9df07e1f33ad3d52accf29288495951cbbdecc47da93e6aec69f47ec22R36

PR #1066, opened 2023-05-02
add features flags to jinjava config

boulter:
When building https://github.com/HubSpot/jinjava/pull/1065, I didn't like adding yet another property to the JinjavaConfig builder. I figured we could use a simple Feature Flags implementation. 

Partially inspired by https://ff4j.org/ and https://www.togglz.org/ . 

This can replace any boolean configs like enablePreciseDivideFilter, enableRecursiveMacroCalls, failOnUnknownTokens,  nestedInterpretationEnabled and more in the future.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/features/DelegatingFeatureActivationStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/features/DelegatingFeatureActivationStrategy.java
@@ -1,0 +1,20 @@
+package com.hubspot.jinjava.features;
+
+import java.util.function.Supplier;
+
+public class DelegatingFeatureActivationStrategy implements FeatureActivationStrategy {
+  public Supplier<Boolean> delegate;
+
+  public static FeatureActivationStrategy of(Supplier<Boolean> delegate) {
+    return new DelegatingFeatureActivationStrategy(delegate);
+  }
+
+  private DelegatingFeatureActivationStrategy(Supplier<Boolean> delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public boolean isActive() {
+    return delegate.get();
+  }
+}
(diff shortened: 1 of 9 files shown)


src/main/java/com/hubspot/jinjava/features/DelegatingFeatureActivationStrategy.java
```diff
@@ -0,0 +1,20 @@
+package com.hubspot.jinjava.features;
+
+import java.util.function.Supplier;
+
+public class DelegatingFeatureActivationStrategy implements FeatureActivationStrategy {
```
jasmith-hs, line 5: I think `DisabledFeatureActivationStrategy`, `DelegatingFeatureActivationStrategy`, and `StaticFeatureActivationStrategy` are redundant implementations as you can use a lambda:
```java
FeatureConfig
  .newBuilder()
  .add(name1, () -> false)
  .add(name2, () -> true)
  .add(name3, supplier)
```

jasmith-hs (commented):
This is definitely more sustainable than the current approach, and it lets you define features for extensions of Jinjava!

src/main/java/com/hubspot/jinjava/features/FeatureStrategies.java
```diff
@@ -0,0 +1,10 @@
+package com.hubspot.jinjava.features;
+
+public class FeatureStrategies {
+  public static final FeatureActivationStrategy INACTIVE = DelegatingFeatureActivationStrategy.of(
+    () -> false
+  );
+  public static final FeatureActivationStrategy ACTIVE = DelegatingFeatureActivationStrategy.of(
+    () -> true
+  );
```
jasmith-hs, line 9:
```suggestion
  public static final FeatureActivationStrategy INACTIVE = () -> false;
  public static final FeatureActivationStrategy ACTIVE = () -> true;
```

src/main/java/com/hubspot/jinjava/features/DelegatingFeatureActivationStrategy.java
jasmith-hs: The `DelegatingFeatureActivationStrategy` boilerplate can be removed too

boulter: Done and also isActive now gets `Context` objects so it can make decisions based on other data that's in there, perhaps a user or customer userid.

jasmith-hs (approved):

PR #1068, opened 2023-05-04
Isolate interpreter scope when rendering value in striptags filter

mattcoley:
A bit of a complex bug with a simple solution. 

We are seeing an issue where the execution of the striptags value on a String is causing the whole template to fail. I have narrowed it down to the fact that in the striptags filter we execute the Jinjava in the String before escaping the value. 

This execution has existed since the initial commit in the repo https://github.com/HubSpot/jinjava/commit/ee1a3a99442603c4a8560ad0563cae41d4f83076

What was missed is that this execution is not scoped. Meaning that any Jinjava inside the `striptags` argument can pollute the outer context and break the overall rendering. This seems to have been a problem in the early days of Jinjava resulting in some filters implementing the isolated scope pattern (see https://github.com/HubSpot/jinjava/commit/1e5f17b6747745e77811596ff6f2fc39c54b82de) but it appears this filter was missed in that initial fix from 8 years ago.

So here I am 8 years later. We have not come across this issue until recently when a common context variable that commonly has `striptags` applied to it started including Jinjava frequently. This included Jinjava is polluting the outer scope and causing rendering issues.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/StripTagsFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/StripTagsFilter.java
@@ -36,20 +36,22 @@ public class StripTagsFilter implements Filter {
     }
     int numDeferredTokensStart = interpreter.getContext().getDeferredTokens().size();
 
-    String val = interpreter.renderFlat((String) object);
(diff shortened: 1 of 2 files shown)


boulter:
Are there any other filters that could have this problem?

PR #1069, opened 2023-05-04
Creating EscapeHubLUtil to share across codebases

Ullas-Gurukumar:
Introduces static methods which escapes HubL delimiters. Going to use this in renderer and content4j

### DIFF
--- a/src/main/java/com/hubspot/jinjava/util/EscapeHubLUtil.java
+++ b/src/main/java/com/hubspot/jinjava/util/EscapeHubLUtil.java
@@ -1,0 +1,34 @@
+package com.hubspot.jinjava.util;
+
+import java.util.Collection;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


boulter (changes requested):
Thanks for the contribution! 

HubL, content4j and cos-renderer are HubSpot concepts which are not known in the open-source Jinjava project. 

At a minimum, HubL should be replaced with Jinjava in this class.

Is there another relevant class where these methods would make sense to add to? *Util classes tend to accumulate junk.

jasmith-hs (changes requested):
I concur with Jeff's points here.
Making this into `EscapeJinjavaFilter.escapeInCollection` and `EscapeJinjavaFilter.escapeInMap` and `<T> T escapeObject(T input)` would be better

PR #1076, opened 2023-05-23
Make restricted methods and properties to be configurable in `JinjavaConfig`

manheychiu:
In `JinjavaBeanELResolver`, restrict methods and properties based on the values from `JinjavaConfig` (`restrictedMethods`, `restrictedProperties` and `deferredExecutionRestrictedMethods`.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -43,6 +43,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
@@ -63,7 +64,13 @@ public class JinjavaConfig {
   private final boolean enableRecursiveMacroCalls;
   private final int maxMacroRecursionDepth;
 
-  private final Map<Context.Library, Set<String>> disabled;
+  private final Map<Library, Set<String>> disabled;
+
+  private final Set<String> restrictedMethods;
+
+  private final Set<String> restrictedProperties;
+  private final Set<String> deferredExecutionRestrictedMethods;
+
   private final boolean failOnUnknownTokens;
   private final boolean nestedInterpretationEnabled;
   private final RandomNumberGeneratorStrategy randomNumberGenerator;
@@ -120,6 +127,9 @@ public class JinjavaConfig {
     timeZone = builder.timeZone;
     maxRenderDepth = builder.maxRenderDepth;
     disabled = builder.disabled;
+    restrictedMethods = builder.restrictedMethods;
+    restrictedProperties = builder.restrictedProperties;
+    deferredExecutionRestrictedMethods = builder.deferredExecutionRestrictedMethods;
     trimBlocks = builder.trimBlocks;
     lstripBlocks = builder.lstripBlocks;
     enableRecursiveMacroCalls = builder.enableRecursiveMacroCalls;
@@ -217,6 +227,18 @@ public class JinjavaConfig {
     return disabled;
   }
 
+  public Set<String> getRestrictedMethods() {
+    return restrictedMethods;
+  }
+
+  public Set<String> getRestrictedProperties() {
+    return restrictedProperties;
+  }
+
+  public Set<String> getDeferredExecutionRestrictedMethods() {
+    return deferredExecutionRestrictedMethods;
+  }
+
   public boolean isFailOnUnknownTokens() {
     return failOnUnknownTokens;
   }
@@ -305,6 +327,10 @@ public class JinjavaConfig {
     private long maxOutputSize = 0; // in bytes
     private Map<Context.Library, Set<String>> disabled = new HashMap<>();
 
+    private Set<String> restrictedMethods = new HashSet<>();
+    private Set<String> restrictedProperties = new HashSet<>();
+    private Set<String> deferredExecutionRestrictedMethods = new HashSet<>();
+
     private boolean trimBlocks;
     private boolean lstripBlocks;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -37,7 +38,7 @@ public class JinjavaBeanELResolver extends BeanELResolver {
     .add("wait")
     .build();
 
-  private static final Set<String> DEFERRED_EXECUTION_RESTRICTED_METHODS = ImmutableSet
+  private static final Set<String> DEFAULT_DEFERRED_EXECUTION_RESTRICTED_METHODS = ImmutableSet
```
jasmith-hs, line 41: I think we can leave this as a static set for now. There's no use-case for making it configurable

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -208,10 +227,19 @@ private static int pickMoreSpecificMethod(Method methodA, Method methodB) {
...
 
-    if (RESTRICTED_PROPERTIES.contains(propertyName)) {
+    JinjavaInterpreter interpreter = (JinjavaInterpreter) context
+      .getELResolver()
+      .getValue(context, null, ExtendedParser.INTERPRETER);
+
+    Set<String> allRestrictedProperties = Sets.union(
+      interpreter.getConfig().getRestrictedProperties(),
+      DEFAULT_RESTRICTED_PROPERTIES
+    );
```
jasmith-hs, line 240: My hunch is that it's faster to check if either set contains it rather than unionizing the sets.

src/test/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolverTest.java
```diff
@@ -11,10 +15,17 @@ public class JinjavaBeanELResolverTest {
...
+  JinjavaConfig config = mock(JinjavaConfig.class);
+
   @Before
   public void setUp() throws Exception {
     jinjavaBeanELResolver = new JinjavaBeanELResolver();
     elContext = new JinjavaELContext();
+    elContext
+      .getELResolver()
+      .setValue(elContext, null, ExtendedParser.INTERPRETER, interpreter);
+    when(interpreter.getConfig()).thenReturn(config);
```
jasmith-hs, line 28: Could add a couple tests that custom properties/methods are restricted

manheychiu: I have added the unit tests to test that custom properties/methods are restricted

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
manheychiu: I changed it to a check instead of unionizing the sets.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
manheychiu: I remove this from config.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -305,6 +321,9 @@ public static class Builder {
     private long maxOutputSize = 0; // in bytes
     private Map<Context.Library, Set<String>> disabled = new HashMap<>();
 
+    private Set<String> restrictedMethods = new HashSet<>();
+    private Set<String> restrictedProperties = new HashSet<>();
```
jasmith-hs, line 325: Would feel a bit safer if these were `ImmutableSet.of()`

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -355,6 +374,16 @@ public Builder withDisabled(Map<Context.Library, Set<String>> disabled) {
       return this;
     }
 
+    public Builder withRestrictedMethods(Set<String> restrictedMethods) {
+      this.restrictedMethods = restrictedMethods;
```
jasmith-hs, line 378:
```suggestion
      this.restrictedMethods = ImmutableSet.copyOf(restrictedMethods);
```

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -355,6 +374,16 @@ public Builder withDisabled(Map<Context.Library, Set<String>> disabled) {
...
+      return this;
+    }
+
+    public Builder withRestrictedProperties(Set<String> restrictedProperties) {
+      this.restrictedProperties = restrictedProperties;
```
jasmith-hs, line 383:
```suggestion
      this.restrictedProperties = ImmutableSet.copyOf(restrictedProperties);
```

jasmith-hs (approved):
LGTM after the last point about immutability is addressed

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -100,7 +104,15 @@ public Object invoke(
     Class<?>[] paramTypes,
     Object[] params
   ) {
-    if (method == null || RESTRICTED_METHODS.contains(method.toString())) {
+    JinjavaInterpreter interpreter = (JinjavaInterpreter) context
```
hs-lsong, line 107: Do you need to check `null `for the interpreter?

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
```diff
@@ -100,7 +104,15 @@ public Object invoke(
...
+      .getELResolver()
+      .getValue(context, null, ExtendedParser.INTERPRETER);
```
hs-lsong, line 109: Can you use `JinjavaInterpreter.getCurrent()` instead of using the resolver to get the interpreter? I think it might be faster.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
manheychiu: Thanks. I have changed it to `ImmutableSet`.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
manheychiu: I added the null check.

src/main/java/com/hubspot/jinjava/el/ext/JinjavaBeanELResolver.java
manheychiu: I changed to `JinjavaInterpreter.getCurrent()`,

hs-lsong (approved):

PR #1081, opened 2023-06-12
Fix EagerDoTag not committing results when partially evaluated

jasmith-hs:
In EagerDoTag, because of this logic we weren't committing information about variables which were declared inside the `{% do %}`:
```java
if (
  !eagerExecutionResult.getResult().isFullyResolved() ||
  interpreter.getContext().isDeferredExecutionMode()
) {
  prefixToPreserveState.withAll(eagerExecutionResult.getPrefixToPreserveState());
} else {
  interpreter.getContext().putAll(eagerExecutionResult.getSpeculativeBindings());
}
```
The do block doesn't create a child scope so variables declared inside of it should propagate outside of the block. This was happening when it was fully resolved, but when it was partially resolved, we weren't committing the results. Specifically, we weren't either deferring the variable or taking any fully resolved values.

To do this, we need to distinguish between which speculative bindings (those child bindings that come from evaluating the `{% do %}` block) are fully resolved, and which are deferred values.
This distinction is well demonstrated in this test case:
```
{% do %}
  {% set list1 = ['a'] %}
  {% set list2 = ['b'] %}
  {% do list2.append(deferred) %}
{% enddo %}
L1: {{ list1 }}
L2: {{ list2 }}
```
`list1` is fully resolved when exiting the `{% do %}` block, but `list2` is deferred. We want to commit that information and make the context have `list1` = `['a']` and `list2` = `DeferredValue.instance(['b'])`. This way we can resolve `{{ list1 }}`, but not resolve `{{ list2 }}`. This is what we need, and before this PR, both `list1` and `list2` would remain null so they'd resolve to nothing.

So to make this change, instead of converting DeferredValues back to their original value anytime we add one to the `SpeculativeBindings`, we'll only convert back to the original value when calling `EagerReconstructionUtils#resetSpeculativeBindings` or when creating a set tag to reconstruct the value.


npharate (approved):

src/test/resources/eager/commits-variables-from-do-tag-when-partially-resolved.expected.expected.jinja
```diff
@@ -0,0 +1,2 @@
+L1: ['a']
```
boulter, line 1: for short files like these, I generally prefer to inline them in the test so it's easier to see the input and output without flipping between 3 files.

boulter (commented):
The description of the solution makes sense, but I'm having trouble grasping what effect each of the code changes you made has and what they actually do. Perhaps you could comment on each change?

src/main/java/com/hubspot/jinjava/util/EagerContextWatcher.java
```diff
@@ -209,12 +209,7 @@ private static Map<String, Object> getBasicSpeculativeBindings(
             .peek(
               entry -> ((OneTimeReconstructible) entry.getValue()).setReconstructed(true)
             )
-            .collect(
-              Collectors.toMap(
-                Entry::getKey,
-                entry -> ((OneTimeReconstructible) entry.getValue()).getOriginalValue()
-              )
-            )
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue))
```
jasmith-hs, line 212: Purpose: Allow storing DeferredValues in the SpeculativeBindings

src/main/java/com/hubspot/jinjava/util/EagerContextWatcher.java
```diff
@@ -226,15 +221,12 @@ private static Map<String, Object> getBasicSpeculativeBindings(
       .map(
         entry -> {
           if (
-            (
-              eagerExecutionResult.getResult().isFullyResolved() ||
-              eagerChildContextConfig.takeNewValue
-            ) &&
-            !(entry.getValue() instanceof DeferredValue) &&
-            entry.getValue() != null
```
jasmith-hs, line 234: Moved this filtering to line 249

src/main/java/com/hubspot/jinjava/util/EagerContextWatcher.java
```diff
@@ -248,24 +240,15 @@ private static Map<String, Object> getBasicSpeculativeBindings(
             ) {
               throw new CannotReconstructValueException(entry.getKey());
             }
-            return new AbstractMap.SimpleImmutableEntry<>(
-              entry.getKey(),
-              ((DeferredValue) contextValue).getOriginalValue()
-            );
+            return new AbstractMap.SimpleImmutableEntry<>(entry.getKey(), contextValue);
```
jasmith-hs, line 243: Purpose: Allow storing DeferredValues in the SpeculativeBindings

src/main/java/com/hubspot/jinjava/util/EagerContextWatcher.java
```diff
@@ -373,9 +364,6 @@ private static Object getOriginalValue(
     boolean isFullyResolved
   ) {
     if (eagerChildContextConfig.takeNewValue || isFullyResolved) {
-      if (e.getValue() instanceof DeferredValue) {
-        return ((DeferredValue) e.getValue()).getOriginalValue();
-      }
```
jasmith-hs, line 378: Purpose: Allow storing DeferredValues in the SpeculativeBindings

src/main/java/com/hubspot/jinjava/util/EagerContextWatcher.java
```diff
@@ -385,7 +373,7 @@ private static Object getOriginalValue(
         .get(e.getKey())
         .equals(getObjectOrHashCode(((DeferredValue) e.getValue()).getOriginalValue()))
     ) {
-      return ((DeferredValue) e.getValue()).getOriginalValue();
```
jasmith-hs, line 388: Purpose: Allow storing DeferredValues in the SpeculativeBindings

src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -747,7 +750,14 @@ public static Set<String> resetSpeculativeBindings(
...
       .getSpeculativeBindings()
-      .forEach((k, v) -> replace(interpreter.getContext(), k, v));
+      .forEach(
+        (k, v) -> {
+          if (v instanceof DeferredValue) {
+            v = ((DeferredValue) v).getOriginalValue();
+          }
+          replace(interpreter.getContext(), k, v);
+        }
+      );
```
jasmith-hs, line 760: Purpose: this is used to reset the state of the context to what it was before some logic was run, which we aren't committing.
For example if evaluating something within a deferred `{% if bool %}` block, we'll want to reset the bindings to what they used to be, this includes makes DeferredValues no longer be deferred so that the `{% else %}` block will evaluate properly. We'll then re-defer any values that got deferred during any if-elif-else blocks.
That logic already happens, I'm just explaining why we'd want to put the `DeferredValue#getOriginalValue()` onto the context here

jasmith-hs (commented):
Hopefully this is helpful @boulter

src/test/resources/eager/commits-variables-from-do-tag-when-partially-resolved.expected.expected.jinja
boulter: I don't see how defining these strings statically in the class is any different than using external files, but you seem to feel strongly about this.

boulter (approved):
I think I understand this now. Thanks.

PR #1082, opened 2023-06-15
Unwrap JsonMappingException if cause is LengthLimitingJsonProcessingException

jasmith-hs:
As a follow up to https://github.com/HubSpot/jinjava/pull/1051, since these exceptions are now wrapped in `JsonMappingException`, we can unwrap them to see if they are caused by `LengthLimitingJsonProcessingException` so that we'll convert it to an `OutputTooBigException`

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
@@ -38,27 +38,38 @@ public abstract class EagerTagDecorator<T extends Tag> implements Tag {
   @Override
   public final String interpret(TagNode tagNode, JinjavaInterpreter interpreter) {
     try {
-      String output = innerInterpret(tagNode, interpreter);
-      JinjavaInterpreter.checkOutputSize(output);
-      return output;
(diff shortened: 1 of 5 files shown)


boulter (approved):
thank you

src/main/java/com/hubspot/jinjava/lib/tag/eager/EagerTagDecorator.java
```diff
@@ -38,29 +38,40 @@ public T getTag() {
...
-        throw new DeferredValueException(
-          String.format("Output too big for eager execution: %s", e1.getMessage())
-        );
+        output = innerInterpret(tagNode, interpreter);
+      } catch (DeferredValueException | TemplateSyntaxException e) {
+        return wrapEagerInterpret(tagNode, interpreter, e);
       }
+      if (JinjavaInterpreter.isOutputTooLarge(output)) {
+        return wrapEagerInterpret(tagNode, interpreter, null);
+      }
```
jasmith-hs, line 49: We only want to do the deferred execution eager interpret if we fully succeeded, but the output is too large. If we encountered an uncaught OutputTooBigException while running `innerInterpret`, we haven't handled that, and the context won't be in a proper state. (It likely would come from attempting to reconstruct a value, but its string representation is too large.)

PR #1086, opened 2023-06-23
Unwrap OutputTooBigException when wrapped in ELException

jasmith-hs:
We can add a more specific error for OutputTooBigException here when it's thrown as the result of evaluating an ELExpression, such as calling a macro function with a very large output.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
+++ b/src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
@@ -154,79 +154,7 @@ public class ExpressionResolver {
         )
       );
     } catch (ELException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/hubspot/jinjava/el/ExpressionResolver.java
```diff
@@ -269,6 +198,91 @@ private Object resolveExpression(String expression, boolean addToResolvedExpress
...
+          ErrorType.FATAL,
+          ErrorReason.OUTPUT_TOO_BIG,
+          ErrorItem.FUNCTION,
+          e.getCause().getMessage(),
+          null,
+          interpreter.getLineNumber(),
+          interpreter.getPosition(),
+          e
+        )
+      );
```
jasmith-hs, line 260: New case

PR #1088, opened 2023-06-28
Reconstruct nested interpretation variables in string representation

jasmith-hs:
If a variable is reconstructed, but it implicitly depends on another variable due to the use of nested interpretation, we'll need to reconstruct the dependent variable if nested interpretation is enabled.
This is well demonstrated through the new test:
```
{% set i_am = 'I am' %}
{% set foo = '{{ i_am }} foo' %}
{% set bar = '{{ i_am }} bar' %}
{% set map_with_vals = {'foo': 'Foo is {{ foo }}.'} %}
{% if deferred %}
{% do map_with_vals.put('bar', 'Bar is {{ bar }}.') %}
{% endif %}
map.foo: {{ map_with_vals.foo }}
map.bar: {{ map_with_vals.bar }}
```

`foo` and `bar` are not used explicitly anywhere, but due to nested interpretation, they are implicitly referenced. We can determine these references using a recursive approach with the same depth limit as the depth limit for nested interpretation.

Essentially, when reconstructing a set tag for a variable which may have a string representation which uses nested interpretation, we can use `EagerExpressionResolver#findDeferredWords` to find words which are hidden inside of nested interpretation and also reconstruct them. So, with the above example, when trying to reconstruct `foo`, we realise that there's also a dependency on `i_am`, so we'll reconstruct `i_am` before the reconstruction of `foo`. It would look something like this:
```
{% set i_am = 'I am' %}{% set foo = '{{ i_am }}' %}
```


src/main/java/com/hubspot/jinjava/util/EagerReconstructionUtils.java
```diff
@@ -230,42 +297,126 @@ private static Map<String, String> reconstructMacroFunctionsBeforeDeferring(
...
+          interpreter,
+          depth + 1
+        );
+      }
+    }
+    prefixToPreserveState.put(
+      name,
+      buildSetTag(ImmutableMap.of(name, pyishStringRepresentation), interpreter, false)
+    );
+    return prefixToPreserveState;
```
jasmith-hs, line 416: Here's the recursion part

PR #1105, opened 2023-08-11
A builder pattern to DeferredToken

jasmith-hs:
Adds the builder pattern for a much simpler, and more readable construction of a `DeferredToken`.
Creating a DeferredToken for a SetTag previously looked like:
```java
new DeferredToken(
  new TagToken(
    joiner.toString(),
    tagNode.getLineNumber(),
    tagNode.getStartPosition(),
    tagNode.getSymbols()
  ),
  eagerExecutionResult
    .getResult()
    .getDeferredWords()
    .stream()
    .filter(
      word ->
        !(interpreter.getContext().get(word) instanceof DeferredMacroValueImpl)
    )
    .collect(Collectors.toSet()),
  Arrays.stream(variables).map(String::trim).collect(Collectors.toSet())
)
```

Now looks like:
```java
DeferredToken
  .builderFromImage(joiner.toString(), tagNode.getMaster())
  .addUsedDeferredWords(eagerExecutionResult.getResult().getDeferredWords())
  .addSetDeferredWords(Arrays.stream(variables).map(String::trim))
  .build()
```
An inline `{% set foo = bar %}` tag was the most unwieldy because it could have set variables `['foo']` and used variables `['bar']` to add which get handled differently. This builder pattern cleans that up and makes it obvious which types of deferred words are being added.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/expression/EagerExpressionStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/lib/expression/EagerExpressionStrategy.java
@@ -2,7 +2,6 @@ package com.hubspot.jinjava.lib.expression;
 
 import com.google.common.annotations.Beta;
 import com.hubspot.jinjava.JinjavaConfig;
-import com.hubspot.jinjava.interpret.DeferredMacroValueImpl;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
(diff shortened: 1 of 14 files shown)


boulter (approved):
That is much nicer. If only we used immutables here and it did it for us!

PR #1112, opened 2023-08-23
Use a ThreadLocal semaphore to prevent hashCode recursion

jasmith-hs:
Solves the same problem as https://github.com/HubSpot/jinjava/pull/1111. This demonstrates a different approach where a semaphore is used to ensure that a PyList or PyMap cannot have `hashCode()` called for itself during its own `hashCode()` call.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
+++ b/src/main/java/com/hubspot/jinjava/objects/collections/PyList.java
@@ -7,8 +7,12 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.Semaphore;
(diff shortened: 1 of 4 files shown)


jasmith-hs:
Oh yeah, I could make the semaphore just a boolean. ThreadLocal is to make sure that PyList and PyMap can still have correct hashCode when done async

mattcoley:
Given these are non-concurrent collections do we need to necessarily guarantee multi-threaded support for this hashcode operation? None of the other operations have any kind of locking guarantee in this case. I guess someone could throw in a concurrent map into a pymap and try to use it across multiple threads but that seems pretty atypical for Jinjava use code.

mattcoley (approved):

PR #1122, opened 2023-10-20
Add whitespace trimming functionality for notes and expressions

jasmith-hs:
In Jinja, tags, expressions, and notes should all support whitespace trimming:
> You can also strip whitespace in templates by hand. If you add a minus sign (-) to the start or end of a block (e.g. a [For](https://jinja.palletsprojects.com/en/3.1.x/templates/#for-loop) tag), a comment, or a variable expression, the whitespaces before or after that block will be removed:

Currently, only TagNodes have support for whitespace trimming.

This also handles the scenario from https://github.com/HubSpot/jinjava/pull/599 natively without needing to merge text nodes so I'm removing the text node merging logic.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/TreeParser.java
+++ b/src/main/java/com/hubspot/jinjava/tree/TreeParser.java
@@ -62,12 +62,7 @@ public class TreeParser {
       Node node = nextNode();
 
       if (node != null) {
-        if (node instanceof TextNode && getLastSibling() instanceof TextNode) {
-          // merge adjacent text nodes so whitespace control properly applies
-          getLastSibling().getMaster().mergeImageAndContent(node.getMaster());
-        } else {
-          parent.getChildren().add(node);
-        }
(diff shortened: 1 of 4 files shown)


boulter:
Does jinja support trimming whitespace from notes and expressions? I can't think of notes causing any problems, but could leading whitespace trimming lead to misinterpretation as a negative value?

jasmith-hs:
> Does jinja support trimming whitespace from notes and expressions?

Yes

> I can't think of notes causing any problems, but could leading whitespace trimming lead to misinterpretation as a negative value?

No. There is a legacy issue where `{{ -10 }}` was treated as just `10`, that was fixed with https://github.com/HubSpot/jinjava/pull/903.
With the legacy override, the `-` is only looked for directly after the `{{`
Without the legacy override, the `-` is looked for as the first non-whitespace character after `{{`.

If you don't use the legacy override, then the whitespace trimming just won't work, but it wouldn't cause any change in the interpreted value

PR #1123, opened 2023-10-24
Prevent accidental expressions from being output

jasmith-hs:
This adds a feature flag for `PREVENT_ACCIDENTAL_EXPRESSIONS` where, when combining output, it will check if the last character of the previous node and the first character of the current node create a tag/expression/note start string. If so, it will add a `\n{#- #}` spacer, which, if the entire output is passed back to jinjava, will tokensise properly, and maintain the correct whitespace due to the left-trim of the note.

Note/expression trimming is added in https://github.com/HubSpot/jinjava/pull/1122

For example:
```
if (true) {
{%- print '% print 123 %' -%}
}
```
Would normally output as:
```
if (true) {% print 123 %}
```
But with this feature enabled, it will output as:
```
if (true) {
{#- #}% print 123 %}
```
Which will prevent a `PrintTag` from being in it, if it were to be passed back into the interpreter.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
+++ b/src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
@@ -71,16 +71,7 @@ public class OutputList {
     LengthLimitingStringBuilder val,
     TokenScannerSymbols tokenScannerSymbols
   ) {
-    @SuppressWarnings("StringBufferReplaceableByString")
-    String separator = new StringBuilder()
-      .append('\n')
-      .append(tokenScannerSymbols.getPrefixChar())
-      .append(tokenScannerSymbols.getNoteChar())
-      .append(tokenScannerSymbols.getTrimChar())
-      .append(' ')
-      .append(tokenScannerSymbols.getNoteChar())
-      .append(tokenScannerSymbols.getExprEndChar())
-      .toString();
+    String separator = getWhitespaceSeparator(tokenScannerSymbols);
     String prev = null;
     String cur;
     for (OutputNode node : nodes) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
```diff
@@ -48,6 +51,71 @@ public List<BlockPlaceholderOutputNode> getBlocks() {
...
+    @SuppressWarnings("StringBufferReplaceableByString")
+    String separator = new StringBuilder()
+      .append('\n')
+      .append(tokenScannerSymbols.getPrefixChar())
+      .append(tokenScannerSymbols.getNoteChar())
+      .append(tokenScannerSymbols.getTrimChar())
+      .append(' ')
+      .append(tokenScannerSymbols.getNoteChar())
+      .append(tokenScannerSymbols.getExprEndChar())
+      .toString();
```
boulter, line 83: You could create this statically.

src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
```diff
@@ -48,6 +51,71 @@ public List<BlockPlaceholderOutputNode> getBlocks() {
...
+          prev != null &&
+          prev.length() > 0 &&
+          prev.charAt(prev.length() - 1) == tokenScannerSymbols.getExprStartChar()
+        ) {
+          if (
+            cur.length() > 0 &&
+            (
+              cur.charAt(0) == tokenScannerSymbols.getTag() ||
+              cur.charAt(0) == tokenScannerSymbols.getExprStartChar() ||
+              cur.charAt(0) == tokenScannerSymbols.getNoteChar()
```
boulter, line 99: Maybe this could be extracted to a useful function for checking if it's the start of something?

boulter (approved):

src/main/java/com/hubspot/jinjava/tree/output/OutputList.java
jasmith-hs: I'm not sure what you mean. Do you mean memoize it inside of TokenScannerSymbols? The result is different depending on what symbols are used

boulter: I was thinking that TokenScannerSymbols were static. Doesn't seem that this will be called too often. If it does, then I think memoization would be helpful.

PR #1128, opened 2023-11-06
Introduces a limited length render filter and HTML tag close filter

asegal-hs:
Currently, Jinjava provides filters for truncating HTML.  The purpose of these is to shorten or summarize rich text.  They correctly handle HTML so that tags are not left hanging open.  However, they do not accommodate text that contains raw HubL as well as HTML.  The purpose of this PR is to allow summaries of objects that contain raw HubL text.

This also includes a filter that closes HTML tags that may have been left hanging open.

Example usage in pseudocode:
```
user.name = Alice
text = "<p> Hello, {% user.name %} </p>"
rendered = text | render(13) // <p> Hello, Ali
closed = renderer | closehtml //<p> Hello, Ali</p>
```

This serves as an alternate option to the `truncatehtml` filter that will do the same thing to raw rich text but also correctly render HubL included in it.


src/main/java/com/hubspot/jinjava/lib/filter/CloseHtmlFilter.java
```diff
@@ -0,0 +1,32 @@
...
+    @JinjavaSnippet(
+      code = "{{ \"<p> Hello, world\"|closehtml }}"
+    )
+  }
+)
+public class CloseHtmlFilter implements Filter{
+
+  @Override
+  public String getName() {
+    return "render";
```
jasmith-hs, line 21:
```suggestion
    return "closehtml";
```

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -24,6 +24,12 @@ public String getName() {
 
   @Override
   public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
+    if (args.length > 0) {
+      /*
+       This means a render limit length has been provided.
+       Here we begin a left to right render where we add to an HTML string until the length reaches a certain limit.
```
jasmith-hs, line 30: Yes, not necessarily an HTML string, but the functionality makes sense

jasmith-hs (commented):
These seem like good features to include

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -259,18 +259,46 @@ public String renderFlat(String template) {
...
+  /**
+   * Render the given root node to a certain limit, processing extend parents.
+   * @param root
+   *          node to render
+   * @param renderLimit
+   *          the number of characters in response text
+   * @return
+   */
+  public String render(Node root, long renderLimit) {
+    return render(root, true, renderLimit);
```
jasmith-hs, line 301: Don't need to add this method

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -280,11 +308,12 @@ public String render(Node root) {
...
    * @param processExtendRoots
    *          if true, also render all extend parents
+   * @param renderLimit
+   *          the number of characters the result may contain
    * @return rendered result
    */
-  public String render(Node root, boolean processExtendRoots) {
-    OutputList output = new OutputList(config.getMaxOutputSize());
-
+  public String render(Node root, boolean processExtendRoots, long renderLimit) {
```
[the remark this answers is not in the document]
jasmith-hs: It's easier to increase visibility later than to decrease it

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -259,18 +259,46 @@ public String renderFlat(String template) {
...
+    return render(template, config.getMaxOutputSize());
+  }
+
+  public String render(String template, long renderLimit) {
+    return render(parse(template), true, renderLimit);
   }
 
   /**
-   * Render the given root node, processing extend parents. Equivalent to render(root, true)
+   * Render the given root node, processing extend parents. Equivalent to render(root, true, -1)
```
jasmith-hs, line 270: Don't need to change this, also it's not the same as that since it passes the max output size from the config

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -24,6 +25,10 @@ public String getName() {
 
   @Override
   public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
+    if (args.length > 0) {
+      String firstArg = args[0];
+      return interpreter.render(Objects.toString(var), NumberUtils.toLong(firstArg, 0));
```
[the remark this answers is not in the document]
jasmith-hs: And we should add some nicer error handling here if we can't coerce the String to a Long

jasmith-hs (changes requested):
Looks good, just a few points that we should address before merging

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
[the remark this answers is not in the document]
asegal-hs: That's what I thought, except default [max output size is 0](https://github.com/HubSpot/jinjava/blob/bb95c0c4ee260e11407c492b83b9b15129b0bbe1/src/main/java/com/hubspot/jinjava/JinjavaConfig.java#L321) (unlimited I assume?) so clamping it always goes lower than the provided value.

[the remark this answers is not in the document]
jasmith-hs: Only clamp if the max output size is > 0

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -280,11 +296,15 @@ public String render(Node root) {
...
    * @return rendered result
    */
-  public String render(Node root, boolean processExtendRoots) {
-    OutputList output = new OutputList(config.getMaxOutputSize());
-
+  private String render(Node root, boolean processExtendRoots, long renderLimit) {
+    long safeRenderSize = (config.getMaxOutputSize() == 0)
+      ? renderLimit
+      : Math.min(renderLimit, config.getMaxOutputSize());
+    OutputList output = new OutputList(safeRenderSize);
```
jasmith-hs, line 307: Someone could pass a negative or 0-value render limit and bypass the max output size.

jasmith-hs: And it would be good to test these different cases for ensuring render limit doesn't allow rendering more than configured

jasmith-hs (changes requested):

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
asegal-hs: So by default, this using 0 (unlimited) as the render size makes me wonder what the problem is here.  Do we have cases where we modify this config value for limited length rendering?  As per my logic here in the filter, if no value/an invalid value is provided, we just pass in 0 anyway, so that to me seems like an easy workaround.

Right now, `config.getMaxOutputSize()` returns 0.

jasmith-hs: If `config.getMaxOutputSize()` is 100 and `renderLimit` is 0, then it will bypass the configured limit and render unlimited characters.
For HubL, we are always setting a non-zero value for `config.getMaxOutputSize()`

src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -907,6 +924,20 @@ private String getWrappedErrorMessage(
...
+    if (configMaxOutput == 0) {
+      return providedLimit;
+    }
+
+    if (providedLimit <= 0) {
+      return configMaxOutput;
+    }
+
+    return Math.min(providedLimit, configMaxOutput);
+  }
```
jasmith-hs, line 939: Looks good, can we just get some tests for this logic?

asegal-hs: Done, thanks a lot Jack.  To avoid repeating tests or changing signatures I moved this into a small static utils class which I tested on it's own.  Let me know if you'd rather I just make it public in the interpreter or add more tests to the render method as a whole and take it back to being private.

I did like this a bit for simplicity though.

jasmith-hs (approved):

PR #1135, opened 2023-12-11
Fix serialization of ZonedDateTime

maple-buice:
Serializing `ZonedDateTime` is not supported by Jackson by default, so this PR registers `JavaTimeModule` to add that support.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,11 @@
       <groupId>ch.obermuhlner</groupId>
       <artifactId>big-math</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>2.14.0</version>
+    </dependency>
 
     <dependency>
       <groupId>ch.qos.logback</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -83,11 +83,6 @@
         <artifactId>jackson-dataformat-yaml</artifactId>
         <version>2.14.0</version>
       </dependency>
-      <dependency>
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-jsr310</artifactId>
-        <version>2.14.0</version>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 
@@ -151,10 +146,6 @@
     <dependency>
       <groupId>com.fasterxml.jackson.dataformat</groupId>
       <artifactId>jackson-dataformat-yaml</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.datatype</groupId>
-      <artifactId>jackson-datatype-jsr310</artifactId>
     </dependency>
     <dependency>
       <groupId>ch.obermuhlner</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,11 @@
         <artifactId>jackson-dataformat-yaml</artifactId>
         <version>2.14.0</version>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jsr310</artifactId>
+        <version>2.14.0</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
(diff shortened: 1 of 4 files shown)


niketpatel2525 (approved):
LGTM

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -157,7 +158,8 @@ private JinjavaConfig(Builder builder) {
 
   private ObjectMapper setupObjectMapper(@Nullable ObjectMapper objectMapper) {
     if (objectMapper == null) {
-      objectMapper = new ObjectMapper();
+      objectMapper = new ObjectMapper().registerModule(new JavaTimeModule());
```
niketpatel2525, line 161: What if ObjectMapper is not null? Shouldn't you register JavaTimeModule in that case?

src/main/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilter.java
```diff
@@ -48,7 +50,16 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
     ) {
       varStr = Objects.toString(var);
     } else {
-      varStr = new POJONode(var).toPrettyString();
+      try {
+        varStr =
+          interpreter
+            .getConfig()
+            .getObjectMapper()
```
juy-hs, line 57: Thoughts on registering the module here so we scope the change to the object mapper to the pretty print filter? I think my concern here is if changing the mapper on the config level will inadvertently change something else

src/test/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilterTest.java
```diff
@@ -94,6 +96,10 @@ public int getBar() {
       return 123;
     }
 
+    public ZonedDateTime getDate() {
+      return ZonedDateTime.of(2023, 12, 11, 0, 0, 0, 0, ZonedDateTime.now().getZone());
+    }
```
juy-hs, line 101: Is this method used?

maple-buice: Yes, this tests that `ZonedDateTime` (the problem type) is being serialized successfully.

src/main/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilter.java
maple-buice: Good idea. I'll update.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
maple-buice: Yes, that's a good point. This will be addressed by the change Julia suggested.

src/main/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilter.java
tkindy: I don't think registering here is the right move. This will register the module on every call to the `PrettyPrintFilter`.

I need to look through the PR to figure out where the right place to do it would be.

src/main/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilter.java
```diff
@@ -48,7 +51,17 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
+            .registerModule(new JavaTimeModule())
```
hs-lsong, line 59: Does this mean we register the module every time the function is called? Is this an expensive operation?

src/main/java/com/hubspot/jinjava/lib/filter/PrettyPrintFilter.java
maple-buice: Updated to remove module registration, but the remaining changes are still needed to support using the interpreter's config's `ObjectMapper`.

hs-lsong: 👍

pom.xml
```diff
@@ -83,6 +83,11 @@
         <artifactId>jackson-dataformat-yaml</artifactId>
         <version>2.14.0</version>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
```
hs-lsong, line 87: Do you still need those POM update?

maple-buice: Only for the test, but I'm going to move it into my upcoming `cos-renderer` PR

PR #1140, opened 2023-12-22
Make null values loop-able.

hs-lsong:
When the loop collection contains `null`, the loop var will be searching the outer scope of the value instead of just use the `null` value. For example,

```
context.put("number", -1);
context.put("the_list", Lists.newArrayList(1L, 2L, null, null, null));
String template = "{% for number in the_list %} {{ number }} {% endfor %}";
```

When rendering the above for tag, because the 3rd and later values are `null`, the loop var is `number`, and it will find the value of it from outside of the forTag scope. It will result in the following

```
Expected :" 1  2  null  null  null "
Actual   :" 1  2  -1  -1  -1 "
```

With this special `NullValue` instance,  the loopVar value will stay inside the loop.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/NullValue.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/NullValue.java
@@ -1,0 +1,19 @@
+package com.hubspot.jinjava.interpret;
+
+/**
+ * Marker object of a `null` value. A null value in the map is usually considered
+ * the key does not exist. For example map = {"a": null}, if map.get("a") == null,
+ * we treat it as the there is not key "a" in the map.
+ */
+public final class NullValue {
+  public static final NullValue INSTANCE = new NullValue();
+
+  static NullValue instance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public String toString() {
+    return "null";
+  }
+}
--- a/src/main/java/com/hubspot/jinjava/interpret/NullValue.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/NullValue.java
@@ -8,7 +8,9 @@ package com.hubspot.jinjava.interpret;
 public final class NullValue {
   public static final NullValue INSTANCE = new NullValue();
 
-  static NullValue instance() {
+  private NullValue() {}
+
+  public static NullValue instance() {
     return INSTANCE;
   }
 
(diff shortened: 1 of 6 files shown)


boulter (commented):
Perhaps this should be put in a feature because I could see someone relying on this "default" functionality, intentionally or not.

src/main/java/com/hubspot/jinjava/interpret/NullValue.java
```diff
@@ -0,0 +1,21 @@
+package com.hubspot.jinjava.interpret;
+
+/**
+ * Marker object of a `null` value. A null value in the map is usually considered
+ * the key does not exist. For example map = {"a": null}, if map.get("a") == null,
+ * we treat it as the there is not key "a" in the map.
+ */
+public final class NullValue {
+  public static final NullValue INSTANCE = new NullValue();
```
jasmith-hs, line 9: Just having this have the String value of `null` wouldn't be sufficient.
Likely you'd want to unwrap the `NullValue -> null` inside of `com.hubspot.jinjava.util.ScopeMap#get(Object)`

Creating some construct such as:
```java
record ScopeMap.ScopeValue(Object value) {}
```
And always unwrapping those values when accessing them inside of a ScopeMap seems like it would do the trick

jasmith-hs (changes requested):
We use https://github.com/HubSpot/jinjava/blob/0c47c7f14e6d02b378dc1c625319e6ed9adb090f/src/main/java/com/hubspot/jinjava/LegacyOverrides.java to override the legacy Jinjava behaviour so this could be a good place for this to reside
`withNullableLoopValues()`

src/main/java/com/hubspot/jinjava/interpret/NullValue.java
hs-lsong: Can you elaborate why we need to unwrap the NullValue -> null? I don't think we need anywhere to use the `null` value.

jasmith-hs: `null != "null"`
Here's a quick example:
```
{% set foo = [0, 1, None, 2] %}
{% for i in foo %}
{% if i %}
 i is {{ i }}
{% endif %}
{% endfor %}
```
The expected jinja output is:
```
 i is 1
 i is 2
```
If treating `null` as `"null"`, the output would be:
```
i is 1
i is null
i is 2
```

PR #1143, opened 2024-01-08
Upgrade basepom

jaredstehler:
* upgrade basepom (25 -> 56)
* moves checkstyle config out of src, to omit from artifact
* creates filters for new spotbugs violations [reference](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html)
* fixes line endings in source file
* removes pin forcing build on jdk17


boulter (approved):
Welcome home.

PR #1147, opened 2024-01-18
Add custom serializer for NullValue.

hs-lsong:
We need to serialize NullValue.
```
No serializer found for class com.hubspot.jinjava.interpret.NullValue and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS)
```


jasmith-hs:
This shouldn't be necessary if https://github.com/HubSpot/jinjava/pull/1140#discussion_r1439616572 is handled. If you unwrap the `NullValue` -> `null` inside of `Context#get()` you'll never have to work with a `NullValue`. As its purpose is just to bypass the ScopeMap parent checking, we can isolate it's behaviour to just affecting the ScopeMap parent checking

PR #1072, opened 2023-05-20
render back unresolved tokens

marius-diumia:
This change allow to render back tokens which are unresolved. 

Default behaviour is:
- context `key = "resolved"`
- template `this is {{ key }} but this is {{ missing }}`
- with output `this is resolved but this is `

Enabling configuration `.withRenderBackUnknownTokens(true)` the behaviour is:
- context `key = "resolved"`
- template `this is {{ key }} but this is {{ missing }}`
- with output `this is resolved but this is {{ missing }}`

PR #1150, opened 2024-01-19
Debug Undefined Feature to echo back Uknown tokens.

abarrafo:
Inspired by this [PR - Render back unresolved tokens](https://github.com/HubSpot/jinjava/pull/1072). Implemented instead with features as suggested by @jasmith-hs and named `ECHO_UNDEFINED` . I originally wanted to call it DEBUG_UNDEFINED to get close to this [feature in jinja2](https://github.com/pallets/jinja/blob/3fd91e4d11bdd131d8c12805177dbe74d85e7b82/tests/test_api.py#L354), but that offers a bit more.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/expression/DefaultExpressionStrategy.java
+++ b/src/main/java/com/hubspot/jinjava/lib/expression/DefaultExpressionStrategy.java
@@ -1,5 +1,6 @@
 package com.hubspot.jinjava.lib.expression;
 
+import com.hubspot.jinjava.features.FeatureActivationStrategy;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
 import com.hubspot.jinjava.lib.filter.EscapeFilter;
 import com.hubspot.jinjava.objects.SafeString;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/expression/DefaultExpressionStrategy.java
```diff
@@ -41,6 +43,14 @@ public RenderedOutputNode interpretOutput(
...
       result = EscapeFilter.escapeHtmlEntities(result);
     }
+    final FeatureActivationStrategy feat = interpreter
+          .getConfig()
+          .getFeatures()
+          .getActivationStrategy(ECHO_UNDEFINED);
+
+  if (result.isEmpty() && feat.isActive(interpreter.getContext())){
+      result = master.getImage();
+  }
```
jasmith-hs, line 53: If the expression renders an empty string, `{{ '' }}` then we should allow an empty string to be output.

If instead we moved this to line 24:
```java
if (var == null && feat.isActive(interpreter.getContext()) {
  return new RenderedOutputNode(master.getImage());
}
```
We can make sure that we're only echoing back when it's undefined.

Also, make sure that you run `mvn prettier:write` to format the code, as we have tab spacing set to `2` for this project.

jasmith-hs (changes requested):
Thank you for taking a look at this

marius-diumia:
thank you both @jasmith-hs & @abarrafo for helping with this and merging

abarrafo:
Hey @jasmith-hs , just curious on what your release lifecycle is? Any idea on when the next release will make is into maven central? Thanks so much

PR #1151, opened 2024-01-22
Discards disabled module errors in interpreter when using render filter

asegal-hs:
Includes a method for removing disabled modules from the interpreter when running the render filter.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
@@ -5,7 +5,10 @@ import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateError;
+import com.hubspot.jinjava.interpret.TemplateError.ErrorReason;
 import java.util.Objects;
+import java.util.Stack;
 import org.apache.commons.lang3.math.NumberUtils;
 
 @JinjavaDoc(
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -38,4 +46,20 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
     }
     return interpreter.render(Objects.toString(var));
   }
+
+  private void removeModuleErrors(JinjavaInterpreter jinjavaInterpreter) {
+    Stack<TemplateError> validErrors = new Stack<>();
+    while (jinjavaInterpreter.getLastError().isPresent()) {
+      TemplateError te = jinjavaInterpreter.getLastError().get();
+      jinjavaInterpreter.removeLastError();
+      if (te.getReason() == ErrorReason.DISABLED && te.getMessage().contains("module")) {
```
jasmith-hs, line 55: Let's get rid of the `module` terminology since that's just for HubL. I think it would be fine to remove the module terminology.

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -26,6 +31,9 @@ public String getName() {
 
   @Override
   public Object filter(Object var, JinjavaInterpreter interpreter, String... args) {
+    if (IGNORE_MODULE_DISABLED_EXCEPTION) {
+      removeModuleErrors(interpreter);
+    }
```
jasmith-hs, line 36: This needs to happen after the `interpreter.render` is called.

src/main/java/com/hubspot/jinjava/lib/filter/RenderFilter.java
```diff
@@ -38,4 +46,20 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
+      jinjavaInterpreter.removeLastError();
+      if (te.getReason() == ErrorReason.DISABLED && te.getMessage().contains("module")) {
+        continue;
+      }
+      validErrors.add(te);
+    }
+
+    while (!validErrors.empty()) {
+      jinjavaInterpreter.addError(validErrors.pop());
+    }
```
jasmith-hs, line 63: I imagine this could be quite slow, especially since in most cases there won't be any errors. We could count the number of errors present before calling `interpreter.render` and only check those which are new errors

jasmith-hs (changes requested):

PR #1157, opened 2024-02-14
Update to lastest version of basepom

PtrTeixeira:
This changes the version of basepom to `56.3`. This fixes the mismatch between the `targetJdk` and use of javadoc here. Incidentally, it includes an update to the `maven-release-plugin`, which should make your release flow easier (you shouldn't need to manually go to Sonatype anymore 🙂 )


jaredstehler:
closing in favor of https://github.com/HubSpot/jinjava/pull/1166

PR #1164, opened 2024-03-04
Fix NPE for loop tuple vars of null.

hs-lsong:
If the a loop value is null, we should not try to get the `.getClass()` of it. It throws a NPE. We fixed this for single var loops, did not expect this also happens for tuples.

A bad jinjava tag can trigger it:
```
{% for  for row in module.specific_projects  %}
```

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
@@ -215,6 +215,17 @@ public class ForTag implements Tag {
         } else {
           for (int loopVarIndex = 0; loopVarIndex < loopVars.size(); loopVarIndex++) {
             String loopVar = loopVars.get(loopVarIndex);
+            if (val == null) {
+              if (
+                interpreter.getContext().get(loopVar) != null &&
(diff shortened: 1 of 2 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/ForTag.java
```diff
@@ -215,6 +215,17 @@ public String renderForCollection(
...
             String loopVar = loopVars.get(loopVarIndex);
+            if (val == null) {
+              if (
+                interpreter.getContext().get(loopVar) != null &&
+                interpreter.getConfig().getLegacyOverrides().isKeepNullableLoopValues()
+              ) {
+                interpreter.getContext().put(loopVar, NullValue.INSTANCE);
+              } else {
+                interpreter.getContext().put(loopVar, null);
+              }
```
jasmith-hs, line 226: This and L#208-214 can be extracted to a single method.

jasmith-hs (approved):
This change looks good, but my concern about `NullValue` from https://github.com/HubSpot/jinjava/pull/1147#issuecomment-1898655242 and https://github.com/HubSpot/jinjava/pull/1140#discussion_r1439616572 still exists

PR #1169, opened 2024-03-15
hard max limit for slices

lozenko:
In some cases using ForLoop can led to OOM when we deal with many elements because we start to create lists with huge capacity. 

In this PR we introduce MAX_SLICES limit. Additional improvement is to relax the heap usage by not copying elements to new `ArrayLists` rather reusing the one in `ForLoop` which has `Iterator<?>` it inside.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
@@ -1,13 +1,12 @@
 package com.hubspot.jinjava.lib.filter;
 
-import static com.hubspot.jinjava.util.Logging.ENGINE_LOG;
-
 import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
 import com.hubspot.jinjava.doc.annotations.JinjavaParam;
 import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
 import com.hubspot.jinjava.interpret.InvalidArgumentException;
 import com.hubspot.jinjava.interpret.InvalidReason;
 import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateError;
 import com.hubspot.jinjava.interpret.TemplateSyntaxException;
 import com.hubspot.jinjava.util.ForLoop;
 import com.hubspot.jinjava.util.ObjectIterator;
@@ -84,8 +83,17 @@ public class SliceFilter implements Filter {
         args[0]
       );
     } else if (slices > MAX_SLICES) {
-      ENGINE_LOG.warn(
-        "The limit input value is too large, it's been reduced to " + MAX_SLICES
+      interpreter.addError(
+        new TemplateError(
+          TemplateError.ErrorType.WARNING,
+          TemplateError.ErrorReason.OVER_LIMIT,
+          TemplateError.ErrorItem.FILTER,
+          "The limit input value is too large, it's been reduced to " + MAX_SLICES,
+          null,
+          interpreter.getLineNumber(),
+          interpreter.getPosition(),
+          null
+        )
       );
       slices = MAX_SLICES;
     }
--- a/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
+++ b/src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
@@ -90,8 +90,11 @@ public class SliceFilter implements Filter {
           TemplateError.ErrorType.WARNING,
           TemplateError.ErrorReason.OVER_LIMIT,
           TemplateError.ErrorItem.FILTER,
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -80,20 +80,26 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
...
     int i = 0;
     while (loop.hasNext()) {
-      Object next = loop.next();
       if (i % slices == 0) {
-        currentList = new ArrayList<>(slices);
-        result.add(currentList);
+        if (currentList != null) {
+          result.add(currentList);
+        }
+        currentList = new ArrayList<>();
```
boulter, line 93:
```suggestion
        currentList = new ArrayList<>(slices);
```
so we never have to resize the array.

boulter (commented):
Hard to tell if this will solve the problem. May need to run a test case through a profiler and check the memory allocation.

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
lozenko: as we discussed we keep it due to better heap results

lozenko:
added hard max limit

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -79,21 +83,32 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
         0,
         args[0]
       );
+    } else if (slices > MAX_SLICES) {
+      ENGINE_LOG.warn(
```
boulter, line 87: This won't be viewable by developers. We'd need to add a warning to the context like https://github.com/HubSpot/jinjava/blob/0de0e078e14e210719c622a6032621bf5c5e49af/src/main/java/com/hubspot/jinjava/lib/filter/PlusTimeFilter.java#L64

lozenko: replaced by your suggestion

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -79,21 +82,41 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
         0,
         args[0]
       );
+    } else if (slices > MAX_SLICES) {
+      interpreter.addError(
+        new TemplateError(
+          TemplateError.ErrorType.WARNING,
+          TemplateError.ErrorReason.OVER_LIMIT,
+          TemplateError.ErrorItem.FILTER,
+          "The limit input value is too large, it's been reduced to " + MAX_SLICES,
```
boulter, line 91: Can you refer to the `slices` count as described in the docs on line 26?
Also please add the 1000 limit to those docs.

boulter (commented):
One comment and then I think we're good to merge.

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
lozenko: added

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -79,21 +86,42 @@ public Object filter(Object var, JinjavaInterpreter interpreter, String... args)
         0,
         args[0]
       );
+    } else if (slices > MAX_SLICES) {
+      interpreter.addError(
+        new TemplateError(
+          TemplateError.ErrorType.WARNING,
+          TemplateError.ErrorReason.OVER_LIMIT,
+          TemplateError.ErrorItem.FILTER,
+          "The limit input value of 'slices' parameter is too large, it's been reduced to " +
```
boulter, line 95:
```suggestion
          "The value of the 'slices' parameter is greater than " + MAX_SLICES + ". It's been reduced to " +
```

src/main/java/com/hubspot/jinjava/lib/filter/SliceFilter.java
```diff
@@ -25,7 +26,11 @@
     @JinjavaParam(
       value = "slices",
       type = "number",
-      desc = "Specifies how many items will be sliced",
+      desc = "Specifies how many items will be sliced." +
+      "Please note, the max limit is " +
+      SliceFilter.MAX_SLICES +
+      ". " +
+      "All values above this limit will be reduced.",
```
boulter, line 33:
```suggestion
      desc = "Specifies how many items will be sliced. Maximum value is " +
      SliceFilter.MAX_SLICES +
      ". ",
```

boulter (commented):
Simplified the language a little here.

boulter (approved):

PR #1173, opened 2024-04-15
find first valid deferred words

boulter:
We end up with a lot blank deferred words when they are attributes of array values. 

For example, `{{ deferred.results[0].attribute }}` will return the used tokens `deferred` and  `.attribute`. `.attribute` is treated a value of `attribute` with a base of (blank) because it splits on periods. This cleans up the general case, but the evaluated eager expressions could also be improved so we don't end up with words with leading periods.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/tag/eager/DeferredToken.java
+++ b/src/main/java/com/hubspot/jinjava/lib/tag/eager/DeferredToken.java
@@ -1,6 +1,7 @@
 package com.hubspot.jinjava.lib.tag.eager;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Strings;
 import com.hubspot.jinjava.interpret.CallStack;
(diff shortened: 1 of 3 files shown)


src/main/java/com/hubspot/jinjava/lib/tag/eager/DeferredToken.java
```diff
@@ -413,10 +418,19 @@ private static CallStack acquireMacroStack() {
       .orElse(null);
   }
 
+  private static String getFirstNonEmptyToken(List<String> strings) {
+    return Strings.isNullOrEmpty(strings.get(0)) ? strings.get(1) : strings.get(0);
```
jasmith-hs, line 422: I think that you need to split more if you're going to take the part after a `.` because you could have `.bar.foobar` and you want `bar` not `bar.foobar`

PR #1174, opened 2024-04-19
Add a feature to output template error when the variable is undefined

manheychiu:
This PR adds a feature `OUTPUT_UNDEFINED_VARIABLES_ERROR`. When enabled, a template error (severity=WARNING) will be added when trying to resolve an undefined variable.


src/main/java/com/hubspot/jinjava/interpret/JinjavaInterpreter.java
```diff
@@ -585,6 +588,28 @@ public Object retraceVariable(String variable, int lineNumber, int startPosition
         }
       }
       obj = var.resolve(obj);
+    } else {
+      if (
+        getConfig()
+          .getFeatures()
+          .getActivationStrategy(OUTPUT_UNDEFINED_VARIABLES_ERROR)
+          .isActive(getContext())
```
jasmith-hs, line 596:
```suggestion
          .isActive(context)
```

jasmith-hs (approved):

PR #1191, opened 2024-06-06
Do not require arguement to unixtimestamp function

boulter:
In https://github.com/HubSpot/jinjava/pull/1064, the intention was to require arguments for the unixtimestamp filter, but not the function. The unixtimestamp filter just calls the function though, so the requirement was incorrectly applied to both.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
@@ -28,13 +28,12 @@ public class UnixTimestampFunctionTest {
 
   @Test
   public void itGetsUnixTimestamps() {
-    JinjavaInterpreter.pushCurrent(
-      new JinjavaInterpreter(
-        new Jinjava(),
-        new Context(),
-        JinjavaConfig.newBuilder().build()
-      )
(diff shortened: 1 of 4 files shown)


src/test/java/com/hubspot/jinjava/lib/fn/UnixTimestampFunctionTest.java
```diff
@@ -28,20 +28,20 @@ public void tearDown() {
...
     );
+    JinjavaInterpreter.pushCurrent(jinjavaInterpreter);
     assertThat(Functions.unixtimestamp())
       .isGreaterThan(0)
       .isLessThanOrEqualTo(System.currentTimeMillis());
     assertThat(Functions.unixtimestamp(epochMilliseconds)).isEqualTo(epochMilliseconds);
     assertThat(Functions.unixtimestamp(d)).isEqualTo(epochMilliseconds);
     assertThat(Functions.unixtimestamp((Object) null))
       .isCloseTo(System.currentTimeMillis(), Offset.offset(1000L));
+    assertThat(jinjavaInterpreter.getErrors()).isEmpty();
```
boulter, line 44: this would have caught the errors being thrown that should have only applied to the filter.

hs-lsong (approved):

PR #1195, opened 2024-09-03
Condense overrideable options in Context to ContextConfiguration immutable

jasmith-hs:
The overrideable flags within `Context` are getting a bit out of hand so I'm grouping these with an immutable class `ContextConfiguration`. I'm planning on adding a new flag in a coming PR and having `Context` directly have so many responsibilities is somewhat off-putting

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,16 @@
         <artifactId>jackson-datatype-jdk8</artifactId>
         <version>2.14.0</version>
       </dependency>
+      <dependency>
+        <groupId>com.hubspot.immutables</groupId>
+        <artifactId>hubspot-style</artifactId>
+        <version>${dep.hubspot-immutables.version}</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -19,6 +19,8 @@
 
     <dep.plugin.jacoco.version>0.8.3</dep.plugin.jacoco.version>
     <dep.plugin.javadoc.version>3.0.1</dep.plugin.javadoc.version>
+    <dep.hubspot-immutables.version>1.9</dep.hubspot-immutables.version>
```
boulter, line 22: Not sure about adding immutables into this. In the past, non-HS users have disliked any other dependencies as it makes it difficult to manage version conflicts. You could manually create a builder object for this instead.

boulter: good point.

PR #1213, opened 2024-10-18
Remove Guava dependency with no direct use in codebase

magdel:
I found no direct use of Guava library in library codebase. Why should we have it in dependencies? Use the jininja often clashes with project`s Guava library version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,10 +110,6 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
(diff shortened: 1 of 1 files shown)


thadguidry (approved):
LGTM, and compiles successful.  `mvn dependency:tree` shows no issues.

thadguidry (changes requested):
Guava is still needed it seems currently, at least for JinjavaConfig

magdel:
@thadguidry thank you. Now I see it comes from other dependencies like
hubspot-style
immutables-exceptions

Removing its now I see how much it is used. And with language level 8 Guava is hardly replaceble

PR #1219, opened 2024-11-23
Implement jinja2.ext.loopcontrols extensions

ccutrer:
`break` and `continue` in for loops.

See https://jinja.palletsprojects.com/en/stable/extensions/#loop-controls

### DIFF
--- a/src/main/java/com/hubspot/jinjava/interpret/NotInLoopException.java
+++ b/src/main/java/com/hubspot/jinjava/interpret/NotInLoopException.java
@@ -1,0 +1,14 @@
+package com.hubspot.jinjava.interpret;
+
+/**
+ * Exception thrown when `continue` or `break` is called outside of a loop
+ */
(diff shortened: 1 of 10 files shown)


jasmith-hs (approved):
I will try. This looks good and I have a quick follow-up PR I'm ready to open to make this compatible with eager execution

jasmith-hs:
> Awesome! Thank you so much
> 
> I saw code for eager execution and deferred values, but I don't fully understand what either are. I'll be watching for your follow up PR to increase my understanding.

It will take a bit of work to be able to handle them in eager execution, I'm planning to work alongside a new engineer on my team to do that. But getting the incompatibility marked as unsupported is an easy change for me to make: https://github.com/HubSpot/jinjava/pull/1222

jasmith-hs:
@ccutrer 2.7.4 has been released

PR #1233, opened 2025-02-05
Implement PyMap.get with the optional default.

ccutrer:


ccutrer:
This will fix https://github.com/openhab/openhab-addons/issues/18219

ccutrer:
Ping @jasmith-hs

jasmith-hs:
@ccutrer , what's the reason for not using `getOrDefault`? Like `{{ value_json['update'].getOrDefault('progress', 'null') }}`

src/main/java/com/hubspot/jinjava/objects/collections/PyMap.java
```diff
@@ -21,6 +21,13 @@ protected Map<String, Object> delegate() {
     return map;
   }
 
+  public Object get(String key, Object defaultValue) {
+    if (!map.containsKey(key)) {
+      return defaultValue;
+    }
+    return map.get(key);
```
jasmith-hs, line 28:
```suggestion
    return getOrDefault(key, defaultValue);
```

Can just proxy to the existing method

jasmith-hs (commented):
> Cause I don't control the Jinja. It's from 3rd party software (zigbee2mqtt), that my software (openHAB) has to work with. And that 3rd party software is primary designed to work with other software (Home Assistant) that is written in Python, and thus they have very little incentive to restrict their usage of Jinja to what is compatible with Jinjava. It's open system too... while zigbee2mqtt is a common piece of software that many users integrate with, there's no limitations - there are hundreds of devices that use the same specification, but I couldn't possibly hope to specifically test with.

Oh mb, I didn't realise that was the jinja functionality.

jasmith-hs (approved):

ccutrer:
@jasmith-hs: Any estimate on when the next release will be made?

jasmith-hs:
I will work on releasing Jinjava 2.8.0 soon; hopefully this week

Issue #1163, opened 2024-02-27
Add support for java records


bdshadow:
```java
public record MyRecord(String recordName) {

	/**
	 * without explicit getter Jinjava doesn't see it
	 */
	public String getRecordName() {
		return recordName;
	}
}
```

The following template won't be correctly filled without the getter, although it's not needed for records
```
{{ myRecord.recordName }}
```

I tried to contribute, but it turned out, that the java version in the base pom.xml is still 8. https://github.com/HubSpot/basepom/blob/master/pom.xml#L17. I think such change must be done by the core maintainers, who know all of the possible implications.

It's 2024, time to go for java 21+

jasmith-hs:
We are thinking of doing a major version release and bumping to Java 17 very soon, and then not too long after that we can bump to Java 21.

jasmith-hs:
> Hello With the new version 2.8.1, based on Java 17, records still don't seem to be working. Is there a new syntax to use them ?

Support for records has not been added in Jinjava yet

sjanel:
Hello, any news on this?

PR #1238, opened 2025-03-24
Target Java 17 with 2.8.0-SNAPSHOT

jasmith-hs:
Bump target version from `8` to `17`.
Removes the Java 11 workflow since targeting java 17 does not work in java 11.

We are preparing for this to be a major release as Jinjava will no longer work on JVMs running pre-17 versions of java.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ 11, 17 ]
+        java: [ 17 ]
     name: jdk-${{ matrix.java }}
     steps:
       - uses: actions/checkout@v3
--- a/pom.xml
+++ b/pom.xml
@@ -10,12 +10,12 @@
 
   <groupId>com.hubspot.jinjava</groupId>
   <artifactId>jinjava</artifactId>
-  <version>2.7.5-SNAPSHOT</version>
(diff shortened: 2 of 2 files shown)


jboulter (approved):
What changes require 17?

jasmith-hs:
In particular, [records](https://github.com/HubSpot/jinjava/issues/1163), also:

1. **Improved Code Readability and Maintainability**:
   - **Text Blocks**: Introduced in Java 13 and finalized in Java 15, text blocks allow for easier multiline string literals, which can simplify JSON or HTML template strings.
   - **Pattern Matching**: Pattern matching improvements for `instanceof` and switch expressions can help make the conditional code in template processing more concise and readable.

2. **Enhanced Development Productivity**:
   - **Records**: Java 16 introduced records that can be used to create immutable data classes with less boilerplate code, which can be particularly useful if jinjava manages data configurations or settings that do not change.

3. **Advanced Language Features**:
   - **Sealed Classes**: Available since Java 15, sealed classes provide more control over inheritance by allowing only specific classes to extend or implement them, potentially useful for defining a fixed set of template component behaviors.
   - **Enhanced Switch Expressions**: These allow for more concise and flexible switch statements, especially useful in control flow operations within template logic.

4. **Better Integration and Library Support**:
   - By migrating to Java 17, jinjava can take advantage of the latest libraries and frameworks that may require a minimum of Java 11 or later, supporting better integration within modern Java ecosystems.

PR #1266, opened 2025-10-22
Bump basepom to 65.1

tkindy:
This PR bumps the version of basepom this project uses to the latest, 65.1.

This bump introduces our usage of Maven toolchains which let us use a different version of the JDK to compile, run tests, etc. than the one that the Maven process is running on. This requires a few changes to the POM and GitHub Actions configuration.

I also configured the CI workflow to build and test against all current LTS versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
     <dep.plugin.javadoc.version>3.0.1</dep.plugin.javadoc.version>
     <dep.hubspot-immutables.version>1.9</dep.hubspot-immutables.version>
     <dep.algebra.version>1.5</dep.algebra.version>
+    <dep.mockito.version>5.20.0</dep.mockito.version>
 
 
     <basepom.test.add.opens>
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +16,19 @@ jobs:
...
+            25
           cache: "maven"
       - name: "Build"
-        run: mvn --batch-mode -no-transfer-progress -V verify
+        run: |
+          mvn \
+            --batch-mode \
+            -no-transfer-progress \
+            -V \
+            -Dproject.build.jdk.version=${{ matrix.java }} \
```
tkindy: This property sets the JDK toolchain-aware plugins, like `maven-compiler-plugin` and `maven-surefire-plugin`, use for their forked JVMs.

pom.xml
```diff
@@ -181,6 +181,10 @@
       <groupId>ch.obermuhlner</groupId>
       <artifactId>big-math</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+    </dependency>
```
tkindy: Some of these annotations (like `@CanIgnoreReturnValue`) are now getting added to the generated Immutables, but I'm not sure why. Seems like Immutables will add them if they're on the classpath, but from what I can tell, they've always been on the classpath via Guava. 🤷

jaredstehler (approved):

PR #1274, opened 2026-01-12
Chained filters optimization

hs-lsong:
## Summary

  This PR introduces an optimization for chained filter expressions in Jinjava templates. Instead of creating deeply nested AST nodes for expressions like `{{ foo|filter1|filter2|filter3 }}`, a flattened `AstFilterChain` node is used, which improves performance by reducing recursive call overhead.

  Key changes:
  - Add `AstFilterChain` class that processes filters iteratively instead of recursively
  - Add `FilterSpec` class to represent individual filter specifications in the chain
  - Make optimization configurable via `enableFilterChainOptimization` in JinjavaConfig (defaults to false)
  - **Note: This optimization is disabled for eager execution mode** - `EagerExtendedParser` overrides the config to always use the traditional nested approach

  ## Test plan

  - [x] Added `AstFilterChainPerformanceTest` to verify both approaches produce identical results
  - [x] Automated test verifies optimized version outperforms unoptimized approach
  - [x] Existing tests pass with optimization disabled (default behavior unchanged)

### DIFF
--- a/src/main/java/com/hubspot/jinjava/el/ext/AstFilterChain.java
+++ b/src/main/java/com/hubspot/jinjava/el/ext/AstFilterChain.java
@@ -170,14 +170,7 @@ public class AstFilterChain extends AstRightValue {
       builder.append('|').append(spec.getName());
       AstParameters params = spec.getParams();
       if (params != null && params.getCardinality() > 0) {
-        builder.append('(');
-        for (int i = 0; i < params.getCardinality(); i++) {
-          if (i > 0) {
-            builder.append(", ");
-          }
-          params.getChild(i).appendStructure(builder, bindings);
-        }
-        builder.append(')');
+        params.appendStructure(builder, bindings);
       }
     }
   }


src/test/java/com/hubspot/jinjava/el/ext/AstFilterChainPerformanceTest.java
```diff
@@ -0,0 +1,244 @@
...
+import org.junit.Test;
+
+/**
+ * Performance test to verify that the optimized filter chain performs better than
+ * the nested method approach.
+ *
+ * Run with: mvn test -Dtest=AstFilterChainPerformanceTest
+ * Or run the main() method directly for more detailed output.
+ */
+public class AstFilterChainPerformanceTest {
```
jasmith-hs: Unit tests should be made separate from performance tests

jasmith-hs (approved):

PR #1305, opened 2026-04-01
feat: add StringTokenScannerSymbols for configurable multi-character delimiters (fixes #195)

jmoraleda:
This PR supersedes #1303 (same content from the correct branch).

Here's the revised PR description:

---

**Title:** `feat: add StringTokenScannerSymbols for configurable multi-character delimiters (fixes #195)`

---

**Description:**

Closes #195.

Python's Jinja2 allows full customization of the six delimiter strings via its `Environment` constructor (`block_start_string`, `block_end_string`, `variable_start_string`, `variable_end_string`, `comment_start_string`, `comment_end_string`), plus `line_statement_prefix` and `line_comment_prefix`. Jinjava had no equivalent, making it impossible to use Jinja-style templating in contexts where `{{`, `{%`, or `{#` appear as literal content (e.g. LaTeX documents, some JSON schemas, or Kubernetes YAML with Helm-style markers).

**What this PR adds:**

A new `StringTokenScannerSymbols` class with a builder API that allows all six delimiter strings to be configured independently, with no constraint on length or shared prefix characters:

```java
JinjavaConfig config = JinjavaConfig.newBuilder()
    .withTokenScannerSymbols(StringTokenScannerSymbols.builder()
        .withVariableStartString("\\VAR{")
        .withVariableEndString("}")
        .withBlockStartString("\\BLOCK{")
        .withBlockEndString("}")
        .withCommentStartString("\\#{")
        .withCommentEndString("}")
        .withLineStatementPrefix("%%")
        .withLineCommentPrefix("%#")
        .build())
    .build();
```

**Changes:**

- **`StringTokenScannerSymbols`** *(new)* — builder-configured `TokenScannerSymbols` implementation. Uses Unicode Private Use Area sentinel characters as internal token-kind discriminators so `Token.newToken()` dispatches correctly without changes to `Token`.

- **`TokenScanner`** — adds a string-matching scan path (`getNextTokenStringBased()`) activated when `symbols.isStringBased()` is true. The original char-based path is completely unchanged. Also supports `lineStatementPrefix` and `lineCommentPrefix`, matching Python Jinja2 semantics including indented prefixes.

- **`TokenScannerSymbols`** — adds `isStringBased()` (default `false`), six delimiter-length accessors (`getTagStartLength()` etc.), and two optional line-prefix accessors (`getLineStatementPrefix()`, `getLineCommentPrefix()`). All default implementations preserve existing behaviour.

[...]

PR #1306, opened 2026-04-02
fix: treat backslash as escape character only inside quoted strings, matching Jinja2 behaviour (fixes #1304)

### LINKED ISSUE #1304: Scanner treats backslash as escape character outside quoted strings, diverging from Jinja2 behaviour
In Jinja2, the template scanner is not responsible for backslash interpretation. It locates block/variable/comment delimiters, extracts the raw content between them, and passes it to Python's expression parser. Backslash handling inside string literals is Python's concern, not the scanner's. A bare backslash outside a string literal is simply invalid Python, so Jinja2 never needs to handle it at the scanner level.

Jinjava's `TokenScanner` currently handles backslash at the scanner level unconditionally:

```java
if (c == '\\') {
    ++currPost; // skips the next character regardless of quote context
    continue;
}
```

This causes two problems:

1. **Inside quoted strings**: it works by accident for simple cases, but the scanner pre-consumes the escaped character before JUEL sees it, meaning the expression language never receives the original token intact.
2. **Outside quoted strings**: a `\` that is legitimately part of a Java expression (e.g. in a regex or a custom EL function) is silently swallowed by the scanner before the expression parser can interpret it.

The correct behaviour, matching Jinja2, is to treat `\` as significant only inside quoted string literals — so that `"\""` closes the string at the right place — and leave all other backslashes untouched for the expression parser. This affects both `TokenScanner`'s char-based path and the new string-based path added in #1303. A fix would be a one-line change in both: move the backslash check inside the `inQuote `branch and remove it from the outer block-scanning logic

**Behaviour change concern**

This is a breaking change for any template that currently relies on `\}}` or `\%}` to prevent a closing delimiter from being recognized. We are not aware of Jinja2 supporting this pattern — in Python, embedding a literal `}}` inside a block expression would be a syntax error regardless — but Jinjava users may have come to depend on it as an undocumented escape mechanism.

We would like to ask the maintainers: is this use case known and intentional? If so, would the fix require a compatibility flag, or is a clean break acceptable given that the current behaviour diverges from Jinja2? We are happy to submit a PR once the preferred approach is clear.

jmoraleda:
This PR implements the fix discussed in #1304 , and builds on the scanner infrastructure introduced in #1305 (configurable multi-character delimiters).

**Background**

In Python's Jinja2, the template scanner is not responsible for backslash interpretation. It locates delimiters, extracts the raw content between them, and passes it to the expression parser. Backslash handling inside string literals is the expression parser's concern. A bare backslash outside a string literal is invalid Python, so Jinja2 never needs to handle it at the scanner level.

Jinjava's `TokenScanner` has historically handled backslash unconditionally in both the char-based and string-based scanning paths:

```java
if (c == '\\') {
    ++currPost; // skips the next character regardless of quote context
    continue;
}
```

This has two consequences:

1. **Outside quoted strings:** a `\` followed by a closing delimiter (e.g. `\}}`) prevents the delimiter from being recognized. The block runs to the next valid closer, swallowing surrounding text. The resulting expression content is a JUEL lexical error.
2. **Inside quoted strings:** works by accident for simple cases, but the scanner pre-consumes the escaped character before JUEL sees it.

**What this PR adds**

A new `LegacyOverrides` flag `handleBackslashInQuotesOnly`. When `true`, backslash is treated as an escape character only inside quoted string literals — matching Jinja2 behaviour — and is left untouched outside quotes for the expression parser to handle. Both the char-based and string-based scanning paths respect this flag.

**`LegacyOverrides` preset behaviour**

- `LegacyOverrides.NONE` — flag is `false` (legacy behaviour, default).
- `LegacyOverrides.THREE_POINT_0` — flag is `false`. Although this represents the "correct" behaviour, it is not included here because existing templates in production may rely on `\}}` preventing delimiter recognition, and `THREE_POINT_0` is the current default config in Jinjava. Changing it would be a silent breaking change.
- `LegacyOverrides.ALL` — flag is `true`. `ALL` is explicitly documented as opting into all new correct behaviours, and the existing test suite passes with this setting.

If the maintainers wish to include this in `THREE_POINT_0`, the only required change is to update the existing test `TokenScannerTest.itTreatsEscapedQuotesSameWhenNotInQuotes` to build its config with `handleBackslashInQuotesOnly` explicitly set to `false`, isolating it from the preset value. That is a one-line change and we are happy to include it in this PR if desired.

**Tests**

`BackslashHandlingTest` covers all four combinations of char-based / string-based scanner × legacy / new flag:

- Escaped quotes inside string literals work correctly in both modes.
- The token boundary difference is verified directly via `TokenScanner` (rather than `render()`, which throws a `FatalTemplateErrorsException` in both modes because the expression `x \` is a JUEL lexical error). In legacy mode, `"prefix {{ x \}} suffix }}"` produces two tokens — `TEXT "prefix "` and `EXPR "{{ x \}} suffix }}"`. In new mode it produces three — `TEXT "prefix "`, `EXPR "{{ x \}}"`, `TEXT " suffix }}"`.
- Backslash in variable *values* (not the template itself) is unaffected by either mode.
- The `LegacyOverrides` preset values are explicitly asserted with comments documenting the rationale.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/LegacyOverrides.java
+++ b/src/main/java/com/hubspot/jinjava/LegacyOverrides.java
@@ -19,6 +19,7 @@ public class LegacyOverrides {
     .withUseTrimmingForNotesAndExpressions(true)
     .withKeepNullableLoopValues(true)
     .withIteratorOnlyReverseFilter(true)
+    .withHandleBackslashInQuotesOnly(true)
     .build();
   public static final LegacyOverrides ALL = new LegacyOverrides.Builder()
     .withEvaluateMapKeys(true)
@@ -31,6 +32,7 @@ public class LegacyOverrides {
     .withUseTrimmingForNotesAndExpressions(true)
     .withKeepNullableLoopValues(true)
     .withIteratorOnlyReverseFilter(true)
+    .withHandleBackslashInQuotesOnly(true)
     .build();
   private final boolean evaluateMapKeys;
   private final boolean iterateOverMapKeys;
@@ -42,6 +44,7 @@ public class LegacyOverrides {
   private final boolean useTrimmingForNotesAndExpressions;
   private final boolean keepNullableLoopValues;
   private final boolean iteratorOnlyReverseFilter;
+  private final boolean handleBackslashInQuotesOnly;
 
   private LegacyOverrides(Builder builder) {
     evaluateMapKeys = builder.evaluateMapKeys;
@@ -54,6 +57,7 @@ public class LegacyOverrides {
     useTrimmingForNotesAndExpressions = builder.useTrimmingForNotesAndExpressions;
     keepNullableLoopValues = builder.keepNullableLoopValues;
     iteratorOnlyReverseFilter = builder.iteratorOnlyReverseFilter;
+    handleBackslashInQuotesOnly = builder.handleBackslashInQuotesOnly;
   }
 
   public static Builder newBuilder() {
@@ -100,6 +104,10 @@ public class LegacyOverrides {
     return iteratorOnlyReverseFilter;
   }
 
+  public boolean isHandleBackslashInQuotesOnly() {
+    return handleBackslashInQuotesOnly;
+  }
+
   public static class Builder {
 
     private boolean evaluateMapKeys = false;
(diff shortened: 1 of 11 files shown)


jmoraleda:
Hello @jasmith-hs Made the option default for 3.0 as [you suggested](https://github.com/HubSpot/jinjava/pull/1311#issuecomment-4231462252). Thank you.

Note that this PR is built on top of #1305 as it touches the same files. So it makes sense to review it after that one is merged, at which point the diff with the main branch will be quite small.

jasmith-hs (approved):

PR #1307, opened 2026-04-08
fix(eager): reconstruct deferred set tags before block content

rewaddell:
When a parent template sets a deferred variable after a block placeholder (e.g. `{% set first_name = deferred %}`), child block content using that variable would fail to resolve on the second render pass. This happened because the `{% set %}` tag appeared after the block content in the output, so the second pass evaluated `{{ first_name }}` before it was defined.

The fix adds `reconstructDeferredVariablesBeforeBlock()` to `resolveBlockStubs()`, which identifies deferred variables used (but not set) within a block, finds prior DeferredTokens that define those variables, and prepends their set tag images before the block content.


rewaddell:
https://github.com/HubSpot/jinjava/pull/1312

PR #1310, opened 2026-04-10
Fix NPE when calling macros inside ternary with variable condition

mnobile-hs:
Using a macro in a ternary would lead to the macro name failing to parse correctly and throw an NPE.

For ternary like:

```
{{ greet('world') if myVar else greet('nobody') }}
```

The `else` keyword in a ternary is registered with `Symbol.COLON`, causing `nonliteral()` to greedily interpret `myVar else greet('nobody') ` as the namespace-prefixed function `myVar:greet` instead of `greet`.

The fix adds an image check (`getImage().equals(":")`) so the parser only enters namespace-prefix logic for a literal colon token, not `else`.

### DIFF
--- a/src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
+++ b/src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
@@ -386,6 +386,19 @@ public class MacroTagTest extends BaseInterpretingTest {
     assertThat(jinjava.render(template, context).trim()).isEqualTo("Hello nobody");
   }
 
+  @Test
+  public void itCallsMacroInStandardTernaryWithVariableCondition() {
+    String template =
+      "{% macro greet(name) %}Hello {{ name }}{% endmacro %}" +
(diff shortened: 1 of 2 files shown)


src/test/java/com/hubspot/jinjava/lib/tag/MacroTagTest.java
```diff
@@ -373,6 +373,19 @@ public void itCorrectlyScopesNestedMacroTags() {
     }
   }
 
+  @Test
+  public void itCallsMacroInTernaryWithVariableCondition() {
+    String template =
+      "{% macro greet(name) %}Hello {{ name }}{% endmacro %}" +
+      "{{ greet('world') if myVar else greet('nobody') }}";
```
jasmith-hs, line 380: Could you add an adjacent test for a ternary using `?` and `:` too to ensure that's also parsing properly
```
{{ myVar ? greet('world') : greet('nobody') }}
```

jasmith-hs (approved):

PR #1311, opened 2026-04-11
Add keepTrailingNewline option to LegacyOverrides to match Python Jinja2 default behaviour

jmoraleda:
**Description:**

By default, Python Jinja2 strips a single trailing newline from rendered output (`keep_trailing_newline=False`). Jinjava has historically preserved trailing newlines, causing a subtle but consistent difference when porting templates from Python Jinja2.

This PR adds a `keepTrailingNewline` flag to `LegacyOverrides` that allows users to opt into Python-compatible behaviour.

**Changes:**

- `LegacyOverrides`: adds `isKeepTrailingNewline()`, defaulting to `true` (preserve existing Jinjava behaviour — no impact on current users).
- `JinjavaInterpreter`: applies the strip in `stripTrailingNewlineIfNeeded()`, called at the normal return points of the private `render()` method. Early-exit error paths (`OutputTooBigException`, `CollectionTooBigException`) are intentionally unaffected.
- `TrailingNewlineTest`: covers both flag values, the `NONE`/`THREE_POINT_0`/`ALL` presets, and edge cases (no trailing newline, multiple trailing newlines, rendered expressions).

**Usage:**

To opt into Python-compatible behaviour:

```java
JinjavaConfig config = JinjavaConfig.newBuilder()
    .withLegacyOverrides(
        LegacyOverrides.newBuilder()
            .withKeepTrailingNewline(false)
            .build()
    )
    .build();
```

**Compatibility note:**

`keepTrailingNewline` is set to `true` (legacy behaviour) in `THREE_POINT_0` to avoid breaking existing tests. It is set to `false` in `ALL`. Maintainers may wish to move it to `false` in `THREE_POINT_0` as well if they consider stripping the trailing newline to be part of the canonical Python-compatible behaviour set — the change is intentionally conservative here to leave that decision to the maintainers.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -216,6 +216,17 @@ public class JinjavaConfig {
     return false;
   }
 
+  /**
+   * When {@code false} (default), a single trailing newline is stripped from the rendered
+   * output, matching Python Jinja2's default.
+   * When {@code true}, the trailing newline of
+   * the rendered output is preserved — matching Jinjava's historical behaviour.
+   */
+  @Value.Default
+  public boolean isKeepTrailingNewline() {
+    return false;
+  }
+
   @Value.Default
   public ObjectMapper getObjectMapper() {
     ObjectMapper objectMapper = new ObjectMapper().registerModule(new Jdk8Module());
--- a/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
+++ b/src/main/java/com/hubspot/jinjava/JinjavaConfig.java
@@ -217,14 +217,15 @@ public class JinjavaConfig {
   }
 
   /**
-   * When {@code false} (default), a single trailing newline is stripped from the rendered
-   * output, matching Python Jinja2's default.
-   * When {@code true}, the trailing newline of
-   * the rendered output is preserved — matching Jinjava's historical behaviour.
+   * When {@code false}, a single trailing newline is stripped from the rendered output,
+   * matching Python Jinja2's default.
+   * When {@code true}, the trailing newline of the rendered output is preserved —
+   * matching Jinjava's historical behaviour.
+   * Defaults to {@link LegacyOverrides#getDefaultKeepTrailingNewlineBehavior()}.
    */
   @Value.Default
   public boolean isKeepTrailingNewline() {
-    return false;
+    return getLegacyOverrides().getDefaultKeepTrailingNewlineBehavior();
   }
 
   @Value.Default
--- a/src/main/java/com/hubspot/jinjava/Jinjava.java
+++ b/src/main/java/com/hubspot/jinjava/Jinjava.java
@@ -250,7 +250,7 @@ public class Jinjava {
         .getInterpreterFactory()
         .newInstance(this, context, renderConfig);
       try {
-        String result = interpreter.render(template);
+        String result = stripTrailingNewlineIfNeeded(interpreter.render(template));
         return new RenderResult(
           result,
           interpreter.getContext(),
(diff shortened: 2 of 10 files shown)


jasmith-hs (commented):
Thanks @jmoraleda for the PR. Could you change it so that the default is `false` like the other legacy override options and rename it something like `stripTrailingNewlines`?
Also, as Jinjava 3.0 has not been published yet, so this option can be turned on for 3.0 so that the default behaviour of Jinjava 3.0 is as close to Jinja as possible

jmoraleda:
Thank you @jasmith-hs :

1.  I am ok changing it to `stripTrailingNewlines` and reversing the logic. But since in Python jinja they use

    image 
    
    Can you please confirm you still want to change the name and reverse the logic?


1. I am happy to include this change in `THREE_POINT_0`.  Since `THREE_POINT_0` is the default used for testing, that breaks some of the existing tests. For tests that break, do you prefer me to change the expected output of the tests to match `THREE_POINT_0` default behavior, or do you prefer to modify the test to use a modified configuration that preserves the older behavior?

1. Should I also modify #1306 to add that feature to `THREE_POINT_0` configuration?

jasmith-hs:
Ah, thanks @jmoraleda, since it's _configurable_ in Jinja, I believe that it would be best to have the option added to the top-level JinjavaConfig. We an have this be defaulting to `false` to match Jinja behaviour in the Jinjava 3.0 release since it's a major version bump and we can have breaking changes. (I may do a final 2.8.4 release that includes these some new changes such as these couple that you're PR-ing and have the 2.8.4 version use non-breaking defaults) 

2. Whatever is easier, really. Slight preference for setting the keep trailing slash on

3. Yes, I'd recommend setting it to `true` in `THREE_POINT_0`, which will be the culmination of everything that currently exists turned on at the time of 3.0.0 release, but then after release, it will not have any breaking changes. Since 3.0.0 isn't released yet, the `THREE_POINT_0` option can have the correct (breaking) functionality turned on

jmoraleda:
Hello @jasmith-hs, thank you.

Moved the option from main LegacyOverrides to main config and set default to `false` matching Python as suggested. Updated the tests that broke to set `keepTrailingNewline`.

Note that I ended making this PR on top of #1305 and #1311 as one of the tests that it breaks is the new `StringTokenScannerTest` added there. So it makes sense to review it after both of those have been merged, at which point the diff with the main branch will be quite small.

jmoraleda:
@jasmith-hs The current implementation removes the trailing new line at `Jinjava.renderForResult`. Is this the best place?

I explored applying it inside `JinjavaInterpreter.render` but I found that it did not work in all cases because the render methods inside the interpreter are called recursively in some cases and the option should only be applied to the complete rendering result.

src/main/java/com/hubspot/jinjava/JinjavaConfig.java
```diff
@@ -216,6 +216,17 @@ public boolean isEnableFilterChainOptimization() {
...
 
+  /**
+   * When {@code false} (default), a single trailing newline is stripped from the rendered
+   * output, matching Python Jinja2's default.
+   * When {@code true}, the trailing newline of
+   * the rendered output is preserved — matching Jinjava's historical behaviour.
+   */
+  @Value.Default
+  public boolean isKeepTrailingNewline() {
+    return false;
```
jasmith-hs: @jmoraleda this may be a confusing ask, but could we have this method delegate to LegacyOverrides. Such that with `LegacyOverrides.NONE` the default value is `true`, and with `LegacyOverrides.ALL`, the default value is `false`



```suggestion
    return getLegacyOverrides().getDefaultKeepTrailingNewlineBehavior();
```

Because for optimal parity with jinja, the default behavior should be `false`, but for someone that doesn't want to change the way jinjava is working for them, then `LegacyOverrides.NONE` will keep this config value as it was before this PR (preserving the newlines)

Lmk if that makes sense

jmoraleda: Thank you @jasmith-hs this makes sense to me too.

jmoraleda:
@jasmith-hs Thank you. I implemented your suggestion. It does make sense to me.

jasmith-hs (approved):

jmoraleda:
Thank you @jasmith-hs 

Do you have an approximate timeframe for the 3.0 release?

jasmith-hs:
@jmoraleda I'm incorporating these changes into Jinjava 2.8.4 and going to try to get that released todaay

jasmith-hs:
@jmoraleda Jinjava 2.8.4 is now released https://github.com/HubSpot/jinjava/releases/tag/jinjava-2.8.4

PR #1332, opened 2026-07-29
feat: add regex_match filter and search expression test

craigayre:
## Description

There's no direct way to test whether a string matches a regex. We've been working around it with things like:

```nunjucks
{% if var and var|regex_replace('^pattern$', '') == '' %}
```

This adds a `regex_match` filter and a `search` expression test so you can write:

```nunjucks
{{ 'hello'|regex_match('^[a-z]+$') }}
```

```nunjucks
{% if var is search '[0-9]+' %}
```

```nunjucks
{{ items|selectattr('name', 'search', '^foo') }}
```

Both use RE2 (`com.google.re2j`) consistent with the existing `regex_replace` filter. Both use partial matching (`Matcher.find`) — anchor with `^...$` for full-string matching.

Names follow Ansible parity: `regex_match` for the filter, `search` for the expression test.

## Testing

- `RegexMatchFilterTest` (9 tests): matching/non-matching, anchored patterns, null input, null arg, missing arg, invalid regex, `SafeString`, and `maxStringLength` enforcement.
- `IsStringSearchExpTestTest` (5 tests): matching/non-matching, null arg, `SafeString`, and end-to-end `selectattr` filtering.

## Notes

Implementations follow the patterns of the existing `string_containing` expression test and `regex_replace` filter for consistency.

### DIFF
--- a/src/main/java/com/hubspot/jinjava/lib/exptest/IsStringMatchingRegexExpTest.java
+++ b/src/main/java/com/hubspot/jinjava/lib/exptest/IsStringMatchingRegexExpTest.java
@@ -1,0 +1,79 @@
+package com.hubspot.jinjava.lib.exptest;
+
+import com.google.re2j.Matcher;
+import com.google.re2j.Pattern;
+import com.google.re2j.PatternSyntaxException;
+import com.hubspot.jinjava.doc.annotations.JinjavaDoc;
+import com.hubspot.jinjava.doc.annotations.JinjavaParam;
+import com.hubspot.jinjava.doc.annotations.JinjavaSnippet;
+import com.hubspot.jinjava.interpret.InvalidArgumentException;
+import com.hubspot.jinjava.interpret.InvalidReason;
+import com.hubspot.jinjava.interpret.JinjavaInterpreter;
+import com.hubspot.jinjava.interpret.TemplateSyntaxException;
+
+@JinjavaDoc(
+  value = "Return true if object is a string which matches a specified regular expression " +
+  "(Java RE2 syntax). Uses partial matching (the pattern can match anywhere in the string). " +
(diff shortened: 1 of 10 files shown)


src/main/java/com/hubspot/jinjava/lib/filter/MatchesRegexFilter.java
```diff
@@ -0,0 +1,108 @@
...
+      "    All lowercase letters\n" +
+      "{% endif %}"
+    ),
+  }
+)
+public class MatchesRegexFilter implements Filter {
+
+  @Override
+  public String getName() {
+    return "matches_regex";
```
jasmith-hs: For better parity with Ansible, this should be named `regex_match`

src/main/java/com/hubspot/jinjava/lib/exptest/IsStringMatchingRegexExpTest.java
```diff
@@ -0,0 +1,79 @@
...
+      desc = "Use with selectattr to filter a list by regex",
+      code = "{{ items|selectattr('name', 'string_matching_regex', '^foo') }}"
+    ),
+  }
+)
+public class IsStringMatchingRegexExpTest extends IsStringExpTest {
+
+  @Override
+  public String getName() {
+    return super.getName() + "_matching_regex";
```
jasmith-hs: For better parity with Ansible, this should be named `search`

src/main/java/com/hubspot/jinjava/lib/filter/MatchesRegexFilter.java
craigayre: Thanks, updated as suggested - please lemme know if you had something else in mind

jasmith-hs (approved):

